PR #11, opened 2016-12-26
Fix generation of Jar file.

JamesDrummond:
Now It can be extracted on top of codenvy source in target directory.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,13 +58,57 @@
         </pluginRepository>
     </pluginRepositories>
     <build>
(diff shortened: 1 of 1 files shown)


bmicklea:
Should this be reviewed by Sergii?

JamesDrummond:
No. I will discuss with them tomorrow. I pretty much have made it work all the way through. The only difference is I use local maven repo instead of public codenvy repo. They may need to work with it a little to get better folder naming in target directory and version control but they can create another PR if needed.

JamesDrummond:
Related to https://github.com/eclipse/che/issues/3510

PR #13, opened 2016-12-28
Remove subfolder docs from war.

JamesDrummond:


JamesDrummond:
Related to https://github.com/eclipse/che/issues/3563

PR #15, opened 2016-12-29
Homepage styles

slemeur:
Update content + styles for the index page


JamesDrummond:
Leave /docs/ as permalink for src/main/_docs/index.html

PR #25, opened 2017-01-02
Added image styling to all layouts. Was only on docs layout.

JamesDrummond:
Signed-off-by: James Drummond jdrummond@codenvy.com


JamesDrummond:
Adding this to another PR https://github.com/codenvy/che-docs/pull/26

PR #27, opened 2017-01-03
Workspace admin links and revs

JamesDrummond:


JamesDrummond:
@slemeur Think we can keep codenvy workspace. Doesn't matter if Eclipse Che platform provides it's still a workspace in codenvy product right?

PR #52, opened 2017-01-10
Updates for new CLI in SYNC page

slemeur:
Updates for new CLI in SYNC page


src/main/_docs/use-che-as-an-ide/ide-ssh.md
```diff
@@ -20,7 +20,10 @@ We then provide various client utilities (authored in Docker!) for connecting to
 # List Workspaces  
 You can get a list of workspaces in a Che server that have an SSH agent deployed. These are the workspaces that you can SSH into.
 ```shell  
-$ che action list-workspaces
+$ docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock
```
TylerJewell: Maybe for simplicity in reading these docs, we should simplify the syntax to be `docker run <DOCKER_OPTIONS> eclipse/che OTHER_STUFF`

slemeur: Yeah, was thinking of this - but then I was asking myself, if I'm discovering the product, I would prefer to get whole instructions, otherwise I would forgot what are those and would have to look for them into another page.

It's less readable - but more useful this way.

TylerJewell: I think on any given page we should show the full syntax one time, and then all other examples on the same markdown page can use the shortened syntax.  That would be a compromise.

slemeur: -1:
- users are sometime not reading the whole page.
- users will not understand, why sometime it has full syntax and sometime not

TylerJewell: I think we can safely assume that if you are a user that doesn't read the entire page, that we can take that risk.  :)

PR #56, opened 2017-01-11
Configuration of auto-snapshot and auto-restore behaviors

slemeur:
Configuration of auto-snapshot and auto-restore behaviors

### DIFF
--- a/src/main/_docs/setup/setup-configuration.md
+++ b/src/main/_docs/setup/setup-configuration.md
@@ -178,8 +178,31 @@ FROM <repository>/<image>:<tag>
 FROM my.registry.url:9000/image:latest
 ```
 
+### Workspace Snapshots Behaviors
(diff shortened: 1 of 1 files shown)


src/main/_docs/setup/setup-configuration.md
```diff
@@ -178,9 +178,32 @@ FROM <repository>/<image>:<tag>
 FROM my.registry.url:9000/image:latest
 ```
 
+### Workspace Snapshots Behaviors
+You can configure behaviors for workspace snapshots. By default any workspace stop event will automatically snapshot the workspace runtime (and all of its images). Anything in your workspace `/project` folder will be saved to the data folder you mounted into the Che container. We then use Docker to snapshot the remaining contents of the workspace into a Docker image. There is a matching auto-restore configuration item which will automatically restart the workspace from the latest saved Docker image on disk. When the workspace is started, your project files will be re-mounted or re-synced into `/projects`.
```
JamesDrummond: Can we change `Anything in your workspace `/project` folder will be saved to the data folder you mounted into the Che container.`?

`Any volumes in your workspace such as `/project` folder are not included in snapshot. The `/project` folder will be saved to the data folder you mounted into the Che container. Refer to ["Logs and User Data"](#logs-and-user-data) section for additional information on how/where `/project` folder is saved.`

JamesDrummond: Maybe put `We then use Docker to snapshot...` into another paragraph.

bmicklea:
I also tried to add in the substance of what James requested without making it overly complicated.

PR #59, opened 2017-01-11
Added list of debbugers

bmicklea:
I wanted to keep the list simple since we have a separate page about how to actually use the debuggers but @JamesDrummond can you add links to my top list please?


bmicklea:
Closing without merging as James did the same.

PR #81, opened 2017-01-20
Add environment variables

JamesDrummond:
Created a environment variable database file env.yml. I base env.yml on che.env file. I left all the commented information in as I think it will be helpful when some reads through it and also I will be able to do a diff on new version of the che.env with env.yml to see if anything changes.

My intent is to have environment variables to be looked up in a single location so that changes will happen globally and also allow other products to change the variable name. Environment variable lookup is optional and contributors are not required to provide this.

Combined revisions to workspace machine docs into the PR.


JamesDrummond:
Related to https://github.com/codenvy/codenvy-docs/issues/32

PR #91, opened 2017-01-24
Update setup-configuration.md

JamesDrummond:
Timeout environment variable added.

Related to https://github.com/codenvy/codenvy/issues/1500


bmicklea:
@JamesDrummond - it feels like we should be only documenting these things in one place. So if we want to keep it in configuration then we should have a reference in the Install docs point there. Something like "[Workspace idle timeouts](link to config doc section) can be set when you configure your Codenvy instance."

JamesDrummond:
No idle timeout for che.

PR #100, opened 2017-01-25
Update sync command with new option --unison-verbose

benoitf:
Introduced by https://github.com/eclipse/che/pull/3900


slemeur:
OK for me

PR #103, opened 2017-01-25
Update volume mount docs with correct syntax

bmicklea:
- Clarified that changes should be made to `che.env`
- Fixed environment variable name
- Removed shared volumes from this page


bmicklea:
@JamesDrummond - yes, please add and commit to this branch.

PR #108, opened 2017-01-26
Updated proxy docs for dockerized ARTIK

bmicklea:
all new syntax and approach.


JamesDrummond:
@bmicklea Merge when you are ready.

PR #121, opened 2017-02-01
New agent section in plugins

bmicklea:
- Adds a new plugins-agent.md file with the content that was originally added to ws-agents in workspace admin
- Removal of the agent creation section from ws-agents
- Reference at the end of ws-agents to custom agents in the plugins guide
- Reference in language server page to the plugins-agent page

### DIFF
--- a/src/main/_docs/write-che-ide-plugins/plugins-agents.md
+++ b/src/main/_docs/write-che-ide-plugins/plugins-agents.md
@@ -3,7 +3,7 @@ tags: [ "eclipse" , "che" ]
 title: Creating Custom Agents
 excerpt: ""
 layout: docs
-permalink: /:categories/languageserver/
+permalink: /:categories/plugins-agents/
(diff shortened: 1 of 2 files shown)


src/main/_docs/write-che-ide-plugins/plugins-agents.md
```diff
@@ -0,0 +1,168 @@
+---
+tags: [ "eclipse" , "che" ]
+title: Creating Custom Agents
+excerpt: ""
+layout: docs
+permalink: /:categories/languageserver/
```
JamesDrummond: Why is this called `languageserver`? Also for links to work on right nav `languageserver` needs to be `agents` or filename needs to change to plugins-languageserver.md .

bmicklea:
Good catch @JamesDrummond - can you review to make sure it's correct or commit changes to fix it?

PR #125, opened 2017-02-06
CHE-3959: Create archetype to add new custom agent that is not language server

tolusha:
### Reference issue
https://github.com/eclipse/che/issues/3959


bmicklea:
@tolusha I think @TylerJewell wanted to review these.

TylerJewell:
@tolusha - thanks for prepping these notes, but I recommend that we close this issue.  For now, let's keep all of the examples in the `che-dockerfiles` repository.  I want to then think about a refactoring of the che-docs repository to separate apart working with assemblies (making that a clear section that is easy for anyone to get started with) apart from using our IDE SDK.   And since we will do a big refactoring, we don't need to worry about PRs on docs for now.

But I would like `che-archetypes` repository to have a README that covers most of the scenarios that we are describing, so please keep updating those files for now.

tolusha:
@TylerJewell 
`README.md` already contains info how to use `che-agent-archetype`

Could you clarify pls:
`keep all of the examples in the che-dockerfiles repository`

TylerJewell:
I just want to document how to use each archetype in the README of archetypes for now, but not for here.  BTW, I note in the README.md, there is text in the table "left-aligned" which really should provide a description of what the archetype does.

PR #145, opened 2017-02-17
Update syntax for ssh command

slemeur:
Updates for the ssh command syntax change.

### DIFF
--- a/src/main/_docs/use-che-as-an-ide/ide-ssh.md
+++ b/src/main/_docs/use-che-as-an-ide/ide-ssh.md
@@ -25,7 +25,7 @@ We provide an optional ssh client built into [cli](#connect) for connecting to a
 You can get a list of workspaces in a {{site.product_mini_name}} server that have an SSH agent deployed. These are the workspaces that you can SSH into.
 
 ```shell  {% assign action="list-workspaces"%}
-$ docker run -ti <volume-mounts> {% if site.product_mini_cli=="codenvy" %}codenvy/cli action {{action}} [parameters]{% else %}eclipse/che action {{action}} [parameters]{% endif %} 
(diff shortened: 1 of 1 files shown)


src/main/_docs/use-che-as-an-ide/ide-ssh.md
```diff
@@ -46,15 +46,15 @@ One nice aspect of our SSH capabilities is that they are all done inside of a Do
...
-    workspace             # Workspace name or id.  
-    machine-name          # Connect to a secondary machine in the workspace(docker compose)
+    workspace                   # Workspace name or id.  
+    machine-name                # Connect to a secondary machine in the workspace(docker compose)
 # Parameters
-    --url <url>           # {{ site.product_mini_name }} host where workspaces are running
-{% if site.product_mini_cli=="codenvy" %}    --user <login email>        # Codenvy user name
-    --password <login password> # Codenvy password{% endif %}
+    --url <url>                 # {{ site.product_mini_name }} host where workspaces are running
+    --user <login email>        # {{ site.product_mini_name }} user name
```
JamesDrummond: Che does not use login credentials. Thing you need to change line 55-57 back.

slemeur: AFAIK we keep ability to provide credentials in the case Che is used securely.
@benoitf Do you confirm?

benoitf: We might use also a remote secured che (like codenvy with che cli)

TylerJewell: We should always include the full syntax - even secured credentials for the utilities.

PR #202, opened 2017-04-11
Che minimal and standard assembly documentation

bmicklea:
Documentation outlining how to build minimal (core only) and standard (core plus plugins) assemblies for Che. This is helpful for organizations that are extending Che and want a quick way to build a minimal Che that they can add plugins to.

Replaces https://github.com/eclipse/che-docs/pull/198


bmicklea:
I've implemented an `if` statement in the docs so that the same doc can be used for Codenvy and Che.

PR #221, opened 2017-05-01
Makes it clear that Java JDK is required for agents

bmicklea:
Adds references in two places to clarify that the JDK is required for agents to execute.

Fixes: https://github.com/eclipse/che-docs/issues/220


bmicklea:
Even if the stack has no Java in it?

eivantsov:
If there's no Java, we will install one.

eivantsov:
Make it clear? We DO NOT require JDK in our stack. It's good if you have it since the workspace will boot faster.

bmicklea:
Sorry @eivantsov I mis-copied the link. In this comment: https://github.com/eclipse/che/issues/4963#issuecomment-298013427 was the issue that there was a bug, but now it's fixed and a stack with no Java inside will add Java once agents are installed?

eivantsov:
Yes. This has been fixed. Ws agent rerquires Java. Either user or the agent itself installs it.

bmicklea:
Closing as fixed without merging.

PR #222, opened 2017-05-05
Remove Bitnami cloud start options

bmicklea:
There will be no more Bitnami cloud start options for Eclipse Che so this PR removes mention of them from the docs. Bitnami stacks inside Che will remain.


src/main/_docs/setup/setup-bitnami.md
```diff
@@ -7,81 +7,4 @@ permalink: /:categories/bitnami/
 ---
```
AndrienkoAleksandr: Do we need bitnami category? And file named setup-bitnami.md ...

bmicklea: I'd like to leave the page there for awhile to avoid 404 errors from external sites. Bitnami will remove their Che cloud starters in the coming weeks but they still exist for now.

PR #238, opened 2017-06-02
Update PHP tutorial for PHPUnit 6.x

kaloyan-raev:
[Review](https://codenvy.io/f?id=factoryu868jvl72lv608kl)

PHPUnit 5 is already old. This is an update for the PHP tutorial for the new PHPUnit 6.

### DIFF
--- a/src/main/_docs/tutorials/tutorial-php.md
+++ b/src/main/_docs/tutorials/tutorial-php.md
@@ -58,8 +58,8 @@ In project root, create `composer.json` file with the following contents:
 
 ```json
 {
-    "require": {
-        "phpunit/phpunit": "5.7.*"
+    "require-dev": {
+        "phpunit/phpunit": "^6.1"
     }
 }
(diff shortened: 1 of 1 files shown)


kaloyan-raev:
@bmicklea @slemeur what's the process for small doc update like this in my area of responsibility (PHP)? Can I just merge them? Or do they require an additional review?

slemeur:
Hi @kaloyan-raev, usually, we prefer to get PRs reviewed by at least another commiter on the project. For docs, reviewers can be @bmicklea, @eivantsov and myself. 

But, for these little fixes on docs (and I'm just speaking for docs), we should be flexible and as you said it is your area of expertise/responsibility as commiter on Che - so if you prefer a quick review, you can request it, but if you judge it is not necessary you should go ahead.

kaloyan-raev:
@slemeur Thanks for clarifying.

PR #255, opened 2017-07-05
Archetype docs updates

jonahkichwacoders:
Various Archetype docs updates for things that I am noticing as I work my way through the docs.

Please let me know if you need these as separate PRs, or if having them as separate commits is sufficient.


slemeur:
Thanks for your PR @jonahkichwacoders . It is perfect this way.

PR #285, opened 2017-09-11
Fixing command so prewiew url always work and setting back factory to codenvy.io

sunix:
[Review](https://codenvy.io/f?id=factoryf8qqofp57qnqpbhf)

### DIFF
--- a/.factory.json
+++ b/.factory.json
@@ -47,7 +47,7 @@
     "name": "che-docs",
     "commands": [
       {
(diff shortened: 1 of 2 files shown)


slemeur:
Thanks @sunix !

slemeur:
Should we also put back the button?

sunix:
@slemeur yes it's part of the PR

PR #292, opened 2017-09-20
Add mention of code completeion snippets

tsmaeder:
Update lsp version to 3.0

Signed-off-by: Thomas Mäder <tmader@redhat.com>


slemeur:
@tsmaeder : Should we expand the documentation and explain how it is possible to add snippets in a LS?

tsmaeder:
@slemeur no, I don't think we should, we would just be duplicating the documentation from Microsoft.

slemeur:
Ok, fine for me.

PR #296, opened 2017-10-04
Multi user docs

eivantsov:
Docs added:

1. Instructions on running multi user Che
2. Instructions on configuring KeyCloak and PostgreSQL instances
3. User management in KC (links to official docs with a short overview of what's possible)
4. Organizations and permission docs ported from Codenvy docs


slemeur:
Checked live.
Fix some links, formatting and missing image.

PR #297, opened 2017-10-04
Configuration of Logstash json file logback encoder and appenders

sunix:
documentation for https://github.com/eclipse/che/issues/6537


bmicklea:
@sunix is this waiting on merge for a different code PR?

PR #307, opened 2017-10-24
PR to che6 branch

ghost:
Instructions on how to deploy multi user Che on OpenSHift for che6 branch


bmicklea (approved):

slemeur:
Thanks @eivantsov .

That's great.
I think we should publish that on the current docs as well.
We could add a page to help user testing Che 6. 
WDYT?

PR #314, opened 2017-11-08
Remove extra slash

apupier:


benoitf:
@apupier could you please add Signed-off-by footer in your commit ? for ECA (when doing edit from github it is not working)

apupier:
signed-off added :-)

benoitf:
thanks Aurélien

PR #337, opened 2018-01-24
New Che 6 docs

ghost:
Initial commit for Che6 docs. A few sections are missing - need to bring back Tutorials (having verified they work for Che6) and Selenium docs, and a few critical dev docs pieces like Language Servers.


slemeur (approved):
LGTM

PR #355, opened 2018-02-08
Token vs service account

ghost:


src/main/pages/setup-openshift/openshift-multi-user.md
```diff
@@ -60,6 +60,21 @@ export ENABLE_SSL=false # true by default. Set to false if you have self signed
 export OPENSHIFT_FLAVOR=ocp
 ./deploy_che.sh
 ```
+
+**IMPORTANT!**
+
+If you provide a **token** rather than username/password, Che will use this token in the deployment script to login in and create all [Che infrastructure objects](#deployment-diagram), as well as when creating workspace pods and associated objects. A token may expire due to expiration timeout policy or it can be invalidated when a new token is requested. In this case, Fabric8 client library that Che uses to communicate with OpenShift will fail to create an object with this bad token and fall back to using a service account instead. There are two options here:
```
slemeur, line 66: I think this is not aligned with the installation instructions. The instructions are requiring the user to provide a token, but then we would be saying "If you provide a token rather than username/password". 

Does that mean that using "username/password" would be a better way to install Che? 

Probably links to the docs on openshift would be nice: https://docs.openshift.com/enterprise/3.0/admin_guide/configuring_authentication.html#token-options
https://docs.openshift.com/container-platform/3.6/dev_guide/service_accounts.html

src/main/pages/setup-openshift/openshift-single-user.md
```diff
@@ -127,6 +127,8 @@ export ENABLE_SSL=false # true by default. Set to false if you have self signed
 export OPENSHIFT_FLAVOR=ocp
 ```
 
+If you use an authentication token, make sure you read how to deal with [token invalidation issues](openshift-multi-user.html#openshift-container-platform).
```
slemeur, line 130: I don't think we should be redirecting the user in the multi-user installation page. This would be confusing.

We have a "configuration" page for openshift where those information could be centralized. Then, into single and multi-user setup page, we link to the configuration page with an "Important" mention as you did.

bmicklea (approved):

src/main/pages/setup-openshift/openshift-multi-user.md
ghost: Username and password cannot be used in 100% of cases - only when oc login -u username -p password is possible. If oAuth is configured that is not possible. Anyone familiar with OpenShift will understand it.

What I propose is to use service account by default and create workspaces in the same namespace with Che server. That is a foolproof installation method

slemeur: Ok should it be the default flow for both single and multi-user installation?

PR #362, opened 2018-02-27
Fix the script installation with openshift

ctron:
This should fix the issue with eclipse/che#8933


bmicklea:
@eivantsov - can you check this please?

slemeur (approved):
Thanks @ctron !

ghost (approved):

PR #385, opened 2018-04-01
fix link to docs

ShalokShalom:
:)


ghost:
@ShalokShalom thanks but we have finally moved Che 6 docs :)

PR #394, opened 2018-04-17
Fix openshift deployment scripts have moved

nickboldt:
deployment scripts have moved in master

Change-Id: I68afcce2e50ac731735cf5cfc4ae3834600c0832
Signed-off-by: nickboldt <nboldt@redhat.com>

### DIFF
--- a/src/main/pages/setup-openshift/openshift-single-user.md
+++ b/src/main/pages/setup-openshift/openshift-single-user.md
@@ -107,7 +107,7 @@ To uninstall the addon from the addon list:
 Alternatively, you may download and run deployment scripts instead of using MiniShift addons:
 
 ```shell
-DEPLOY_ROOT_URL=https://raw.githubusercontent.com/eclipse/che/master/deploy/openshift/scripts/
+DEPLOY_ROOT_URL=https://raw.githubusercontent.com/eclipse/che/master/deploy/openshift/
 curl -fsSL ${DEPLOY_ROOT_URL}/deploy_che.sh -o get-che.sh
 curl -fsSL ${DEPLOY_ROOT_URL}/che-openshift.yml -o che-openshift.yml
 curl -fsSL ${DEPLOY_ROOT_URL}/che-config -o che-config


benoitf (approved):

slemeur (approved):

ghost:
@nickboldt I have just merged a PR that changes instructions on how to deploy Che to OpenShift/K8S.

Closing your PR. Sorry about that.

nickboldt:
Hey, no problem, as long as the docs are accurate. :)

I see on https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/openshift-single-user.md#minishift-templates that you still tell users they have to use `--addon-env OPENSHIFT_TOKEN=$(oc whoami -t)` but it seems you don't need that to launch Che on Minishift (I got it working on Win10 w/ VBox 5.1.30 and minishift 1.15.1. ... But then the Che workspaces never loaded. :(

Note too that `$(oc whoami -t)` doesn't work on Windows ootb (requires that the user manually install oc somewhere in the PATH, and use a bash (not cmd) shell. I imagine you already know this, but I didn't see it in the docs (maybe I skipped over it?)

ghost:
@nickboldt I can't find any instructions in the link you have provided where users are told to use addons with templates.

There are two ways to deploy single user Che on MiniShift: 

1. Addon
2. Manual deploying using oc client and yamls that reside in Che repo.

As to Che workspaces not being able to start, please open a new issue and I will figure it out.

IMHO, we should retire MiniShift addon I think - it is not much easier to deploy Che using addon vs running a few oc commands. Also, when smth changes in deployment yamls, addon yamls have to be updated too.

nickboldt:
I suppose the issue is UX -- some of of these docs read like a series of steps, not a set of OPTIONS of series of steps. Maybe it'd be more clear if we broke the documents into separate pages, eg.,

1. First approach - using addon: https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/openshift-single-user.md#minishift-addon

2. Second approach - using templates INSTEAD OF addon (rather than in addition to) https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/openshift-single-user.md#minishift-templates

Maybe it's me, but when I'm 5 steps down a page of copy-this-code-and-run-it instructions, I often miss key labels like "**Alternatively**, you may download templates and run deployment manually **instead of** using MiniShift addons" and just keep blindly following the instructions of what to copy-paste into a console.

So... take that UX feedback as you will. :D

nickboldt:
BTW there's a couple broken image includes on that page too:

https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/openshift-single-user.md

    {% include image.html file="diagrams/ocp_single_user.png" %}


    {% include links.html %}

PR #406, opened 2018-05-11
Documentation changes associated to PR https://github.com/eclipse/che/pull/9577 

davidfestal:
Documentation changes associated to PR https://github.com/eclipse/che/pull/9577 (Allow creating OpenShift objects under the current user account on OCP).


bmicklea (approved):
excellent stuff, thanks!

src/main/pages/setup-openshift/admin-guide.md
```diff
@@ -50,6 +50,10 @@ Advantages and disadvantages of using token, username/pass, and service account:
 |**Username/Password** | Objects can be created outside Che namespace | If username/pass are changed, they need to be updated in Che deployment as well. Many authentication types like oAuth make it impossible to login with username/pass
 | **Service account**   | No expiration risks                          | Cannot create objects outside its namespace unless given admin privileges
 
+**OpenShift-specific feature: create workspace objects in a personal namespace**
+
+When Che is deployed on OpenShift in mullti-user mode, there is an additional option that allows creating workspace objects **on behalf of the currenlty logged-in user**, instead of using the configured user described above. Refer to the [OpenShift Admin Guide](openshift-admin-guide#create-workspace-objects-in-personal-namespaces) for more details about how it works.
```
slemeur: typo "multi-user"

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
 
 {% include links.html %}
 {% include_relative admin-guide.md %}
+
+## Create workspace objects in personal namespaces
+
+When Che is installed on Openshift in multi-user mode, it is possible to register the Openshift server into the Keycloak server as an identity provider,
```
slemeur: typo OpenShift

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+in order to allow creating workspace objects in the personal OpenShift namespace of the user that is currenlty logged in Che through Keycloak.
+
+This feature is available only when Che is confirgured to create a new OpenShift namespace for every Che workspace.
```
slemeur: typo: is configured

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+
+But for non-interactive workspace actions, such as workspace stop on idling or Che server shutdown, the account used for operations on OpenShift resources will 
+fall back to the dedicated OpenShift account configured for the Kubernetes infrastructure, as described in the [AdminGuide](admin-guide#who-creates-workspace-objects).
+
+To install Che on OpenShift with this feature enabled, see [this section for Minishift](openshift-multi-user#creating-workspace-resources-in-personal-openshift-accounts-on-minishift)
+and [this one for OCP](openshift-multi-user#creating-workspace-resources-in-personal-openshift-accounts)
+
+#### OpenShift identity provider registration
+
+The Keycloak Openshift identity provider is described in [this documentation](https://www.keycloak.org/docs/3.3/server_admin/topics/identity-broker/social/openshift.html).
```
slemeur: typo: OpenShift

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+
+{% include image.html file="keycloak/openshift_identity_provider.png" %}
+
+`Base URL` is the URL of the Openshift console
+
+2. Next thing is to add a default read-token role:
+
+{% include image.html file="git/kc_roles.png" %}
+
+3. Then this identity provider has to be declared as an OAuth client inside Openshift. This can be done with the corresponding command:
```
slemeur: typo: OpenShift

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+__Note__: Adding a OAuth client requires cluster-wide admin rights. 
+ 
+
+#### Che configuration
+
+On the Che deployment configuration:
+- the `CHE_INFRA_OPENSHIFT_PROJECT` environment variable should be set to `NULL` to ensure a new distinct OpenShift namespace is created for every started workspace.
+- the `CHE_INFRA_OPENSHIFT_OAUTH__IDENTITY__PROVIDER` environment variable should be set to the alias of the OpenShift identity provider
+registered in Keycloak. The default value is `openshift-v3`.
+
```
slemeur: How to retrieve the OpenShift identity provider, if it's different from default?

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+#### Providing the Openshift certificate to Keycloak
+
+If the certificate used by the Openshift console is self-signed or is not trusted, then by default the Keycloak will not be able to
```
slemeur: typo: OpenShift

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+ 
+
+#### Che configuration
+
+On the Che deployment configuration:
+- the `CHE_INFRA_OPENSHIFT_PROJECT` environment variable should be set to `NULL` to ensure a new distinct OpenShift namespace is created for every started workspace.
+- the `CHE_INFRA_OPENSHIFT_OAUTH__IDENTITY__PROVIDER` environment variable should be set to the alias of the OpenShift identity provider
+registered in Keycloak. The default value is `openshift-v3`.
+
+#### Providing the Openshift certificate to Keycloak
```
slemeur: typo: OpenShift

slemeur (changes requested):
Good things. Just minor typos and one question!

src/main/pages/setup-openshift/openshift-admin-guide.md
ghost: You get it when creating it in Keycloak. There must be a link to Keycloak docs or a screenshot

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+
+When Che is installed on Openshift in multi-user mode, it is possible to register the Openshift server into the Keycloak server as an identity provider,
+in order to allow creating workspace objects in the personal OpenShift namespace of the user that is currenlty logged in Che through Keycloak.
+
+This feature is available only when Che is confirgured to create a new OpenShift namespace for every Che workspace.
+
+To enable this feature, the administrator should:
+- register, inside Keycloak, an Openshift identity provider that will point to the Openshift console of the cluster in which the workspace resources should be created,
+- configure Che to use this Keycloak identity provider in order to retrieve the Openshift tokens of Che users.
+
```
bmicklea: Sorry, I know I approved this but in re-reading it I think we should be more clear on how to do these changes. Can we either write up the steps people need to do here explicitly or at least link off to specific pages of the Keycloak docs?  We need to assume the person using this doesn't know anything about keycloak, openshift or che.

ghost: It turns out all the explanations are below, including screenshots

bmicklea: Can we remove this section then and either (a) bring the more actionable section up or (b) provide a link down. We should avoid sections in the docs that talk generally about what you can do without being actionable.

ghost: I see this as a summary of what needs to be done with detailed instructions below

bmicklea: Understood - but what value does the user get from reading what they'll have to do without an indication of how to do it?  Wouldn't it be better to just say 'to enable this feature follow the instructions in section X' or 'enabling this feature will allow you to easily use Che workspaces from a desktop IDE'

davidfestal:
@slemeur @bmicklea does it seem OK to you after the last commits ?

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -9,3 +9,76 @@ folder: setup-openshift
...
+If the certificate used by the OpenShift console is self-signed or is not trusted, then by default the Keycloak will not be able to
+contact the OpenShift console to retrieve linked tokens.
+
+In this case the OpenShift console certificate should be passed to the Keycloak deployment as an additional environment property.
+This will enable the Keycloak server to add it to its list of trusted certificates, and will fix the problem.
+
+The environment variable is named `OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE`.
+
+Since adding a multi-line certificate content in a deployment configuration environment variable is not that easy, the best way is to use a secret that contains the certificate,
+and refer to it in the environment variable.
```
slemeur, line 84: Should we put an example of how to do that?

ghost: It's in the example for MiniShift. There's a yaml with for the secret

src/main/pages/setup-openshift/openshift-admin-guide.md
```diff
@@ -12,40 +12,40 @@ folder: setup-openshift
...
 
-`Base URL` is the URL of the Openshift console
+`Base URL` is the URL of the OpenShift console
 
 2. Next thing is to add a default read-token role:
 
 {% include image.html file="git/kc_roles.png" %}
 
-3. Then this identity provider has to be declared as an OAuth client inside Openshift. This can be done with the corresponding command:
+3. Then this identity provider has to be declared as an OAuth client inside OpenShift. This can be done with the corresponding command:
```
bmicklea: Perhaps a dumb question - but would this work equally well with any other Keycloak-supported authentication mechanism?  Most organizations use a directory-based auth like LDAP/AD

ghost: No, unless OpenShift auth is tied to ldap. We need a token from OpenShift with full user rights

bmicklea: So the requirement is just that the identity provider for this needs to be the same as the identity provider for OpenShift. They could both be LDAP though, right?

davidfestal: This work only provides a way to link the main Keycloak account to a secondary Openshift account, in order to get the Openshift token. It doesn't change anything to the way the users log into their main Keycloak account / identity. That's why the registered Openshift identity provider should have the `Link Only`, `Store Tokens` and `Stored Tokens Readable` options (see these options [here](https://www.keycloak.org/docs/3.4/server_admin/index.html#_general-idp-config) ).

However I don't see why other identity providers used for logging into Keycloak couldn't be used. Regarding LDAP I also don't see any reason why the [LDAP Federated user storage provider](https://www.keycloak.org/docs/3.4/server_admin/index.html#_ldap) wouldn't work, at least when the [`import ` storage mode](https://www.keycloak.org/docs/3.4/server_admin/index.html#storage-mode) is enabled.

bmicklea: okay the main thing I wanted to make sure of is that if an organization is using LDAP to access OpenShift that won't break this integration. And it sounds like that's true...

davidfestal: > okay the main thing I wanted to make sure of is that if an organization is using LDAP to access OpenShift that won't break this integration. And it sounds like that's true...

@bmicklea Then it seems I had not correctly understood your question: I answered about the case when users would like to connect to the Che server (=> to the Keycloak main account) through LDAP or any other identity provider.

> would this work equally well with any other Keycloak-supported authentication mechanism? - No, unless OpenShift auth is tied to ldap. We need a token from OpenShift with full user rights

@eivantsov could you explain why you say that ?

ghost: I mean we don't care HOW user logs in to OpenShift v3. I mean I may set up login with GitHub (for example) for OpenShift itself and it will be fine in our use case.

ghost: So, @bmicklea no worried, this PR adds functionality not removes it :)

bmicklea:
LGTM

ghost:
@davidfestal feel free to merge once PR in Che repo is merged. 

@slemeur @bmicklea I think PMs are very much interested in making sure this functionality is regularly checked in QA tests, so I encourage you to bring it to their attention.

slemeur (approved):

PR #407, opened 2018-05-17
Stack sharing instructions

ghost:
Rest calls to be made to share a custom stack with all users in the system


slemeur:
I think this explanation better fit in the https://github.com/eclipse/che-docs/blob/master/src/main/pages/workspace-admin/stacks.md

And then having links from https://github.com/eclipse/che-docs/blob/master/src/main/pages/user-guide/creating-starting-workspaces.md to the "stacks" page.

src/main/pages/user-guide/creating-starting-workspaces.md
```diff
@@ -29,6 +29,33 @@ In this scenario, you need to author a custom [recipe][recipes], then a stack, a
 
 {% include image.html file="workspaces/custom_recipe_stack.gif" %}
 
+**Stack Sharing**
```
slemeur: Should we expend the section to be "Configure System Stacks". Where we explain that an admin can edit/remove the stacks available to all users ?

ghost:
Addressed things @slemeur mentioned in the comments.

slemeur (approved):

PR #410, opened 2018-05-30
Add a note about Factories from URL

ghost:
Since dockerfile recipe type isn't supported on OpenShift and k8s infras, this mechanism of creating Factories works on Docker only

### DIFF
--- a/src/main/pages/portable-workspaces/chedir-getting-started.md
+++ b/src/main/pages/portable-workspaces/chedir-getting-started.md
@@ -9,6 +9,8 @@ folder: portable-workspaces
 
 {% include links.html %}
 
+**<span style="color:red;">Chedir is a Docker only feature and will not work on OpenShift/K8S!</span>**
+
 The Chedir getting started guide will walk you through your first Chedir project, and show the basic features Chedir has to offer.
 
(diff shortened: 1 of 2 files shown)


bmicklea:
Is there an issue to add the ability to put a kube YAML in the repo and achieve the same end result?

ghost:
No, there's no such issue

benoitf:
if SPI is set to be OpenShift/K8s the on-the-fly factory could generate a compliant recipe (and not always try to use dockerfile)

bmicklea:
Created: https://github.com/eclipse/che/issues/9879

bmicklea:
Can you comment on the new issue? Also are you saying this should just
work? Or describing how we should make it work in the future?
-- 

Brad Micklea // Director, Product Management // Developer Tools //
416.707.0792

slemeur (approved):

slemeur:
We should also mention that chedir is not working on OS/Kubernetes - it's pure Docker infra

ghost:
@slemeur yes, I will push info about Chedir to this branch

ghost:
@bmicklea what @benoitf says is that it is possible but of course not at the moment.

PR #412, opened 2018-05-30
Add instructions on how to configure Che with self signed certs

ghost:
This PR adds instructions on how to configure Che to work in https mode with self signed certs.

Also, a little bit of housekeeping work - openshift and k8s installation and config guides were duplicated in 2 folders and it was very confusing.


src/main/pages/setup-openshift/openshift-config.md
```diff
@@ -34,15 +34,52 @@ You can manually convert properties into envs, just make sure to follow [instruc
...
 2. Manually edit or recreate routes for Che and Keycloak `oc apply -f https`
 3. Once done, go to `https://keycloak-${NAMESPACE}.${ROUTING_SUFFIX}`, log in to admin console.
 Default credentials are `admin:admin`.
 Go to Clients, `che-public` client and edit **Valid Redirect URIs** and **Web Origins** URLs so that they use **https** protocol.
 You do not need to do that if you initially deploy Che with https support.
 
+## HTTPS Mode - Self-Signed Certs
+
+If you enable HTTPS mode for Che on an OpenShift installation that does not have certificates signed by a public authority, it won't be possible to start workspaces or even login, if it's a multi user Che deployment. There is a lot of communication between Che server and workspace agents, Che server and Keycloak. Therefore, self signed certs should be added to Java trust store of Che server and Keycloak (only for a multi user Che deployment) pods, as well as workspace images. While there is automation for Che server and Keycloak, certs should be manually added to workspace images, since adding a root certificate requires sudo privileges which an arbitrary OpenShift user may not have.
```
bmicklea: You can change first sentence to: `If you enable HTTPS mode for multi-user Che on an OpenShift installation that does not have certificates signed by a public authority, it won't be possible to start workspaces or even login.`

Also is this true of k8s as well?

src/main/pages/setup-openshift/openshift-single-user.md
```diff
@@ -142,6 +142,24 @@ oc apply -f che-server-pvc.yaml
 oc set volume dc/che --add -m /data --name=che-data-volume --claim-name=che-data-volume
 ```
 
+## HTTPS Mode
+
+<span style="color:red;">IMPORTANT!</span> Self-signed certificates aren't acceptable.
```
bmicklea: I don't understand what this means - you say it's unacceptable, but then you seem to redirect to instructions on how to do it...

ghost: Yeah, makes sense. Fixed.

src/main/pages/setup-openshift/openshift-config.md
ghost: Changed.

And these instructions work for OpenShift only, as OpenShift cert is used (the one that gets generated during installation)

bmicklea (approved):

PR #429, opened 2018-06-30
Add Z option to docker volume mount to fix SELinux context.

### LINKED ISSUE #428: Building docs on SELinux-enabled system fails with permission error
When building the docs on a system using SELinux, running the `run.sh` script fails with the following error:
```
./run.sh 
mkdir: can't create directory '_site': Permission denied
chown: _site: No such file or directory
```
This can be fixed by adding the `Z` (which ensures correct labelling of the mounted volume) option to the `-v` option as follows:
```
-v $(pwd)/src/main:/che-docs:Z
```

rkratky:
### What does this PR do?

Fixes SELinux context for successful docs building on SELinux-enabled systems.

Fixes #428.


rkratky:
OK, ECA now accepted.

ghost (approved):

PR #547, opened 2018-10-31
Append / to url, otherwise Che returns a 302

pilhuhn:
Signed-off-by: Heiko W. Rupp <hrupp@redhat.com>

In the current docs when you `curl` the url you get a 302 with the new location back.
Directly adding the `/` to the end prevents that one round-trip.


ghost (approved):

slemeur:
Thanks @pilhuhn !

PR #566, opened 2018-11-06
Provide a che.openshift.io factory to edit documentation

benoitf:
### What does this PR do?
Provide a che.openshift.io factory to edit documentation

N/A


slemeur (approved):
Thanks !

PR #594, opened 2018-11-13
Remove pure kubectl Che deployment instructions

mkuznyetsov:
### What does this PR do?
replace old instructions for deploying single-user Che using pure kubectl https://github.com/eclipse/che/pull/11894 , with approach of using Helm

https://github.com/eclipse/che/issues/11135

### DIFF
--- a/src/main/pages/setup-kubernetes/kubernetes-single-user.adoc
+++ b/src/main/pages/setup-kubernetes/kubernetes-single-user.adoc
@@ -56,7 +56,7 @@ kubectl apply -f ./tiller-rbac.yaml
 helm init --service-account tiller
 ----
 +
-* Ensure that you have an NGINX-based Ingress controller. This is the default Ingress controller on Minikube. To start it, use the `minikube addons enable ingress` command.
+* Ensure that you have an Ingress controller enabled. To start it on Minikube, use the `minikube addons enable ingress` command.
 * DNS discovery should be enabled. It is enabled by default in Minikube.
 
 [id="cluster-ip"]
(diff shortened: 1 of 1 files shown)


src/main/pages/setup-kubernetes/kubernetes-single-user.adoc
```diff
@@ -7,28 +7,96 @@ permalink: kubernetes-single-user.html
...
+minikube start --cpus 2 --memory 4096 --extra-config=apiserver.authorization-mode=RBAC
+----
++
+** For Minikube 0.25.2 and lower, use the following command:
++
+----
+minikube start --cpus 2 --memory 4096 --extra-config=apiserver.Authorization.Mode=RBAC
+----
++
+* To deploy role binding, use the following command:
```
sleshchenko, line 27: >  To deploy role binding

I think it makes sense to describe: Which role binding? Why it is needed?

src/main/pages/setup-kubernetes/kubernetes-single-user.adoc
```diff
@@ -7,28 +7,96 @@ permalink: kubernetes-single-user.html
...
+kubectl apply -f ./tiller-rbac.yaml
+----
++
+.. Install `tiller`.
++
+----
+helm init --service-account tiller
+----
++
+* Ensure that you have an NGINX-based Ingress controller. This is the default Ingress controller on Minikube. To start it, use the `minikube addons enable ingress` command.
```
sleshchenko: Actually, Che is able to work on Kubernetes if any Ingress controller is there.

sleshchenko (approved):
LGTM

src/main/pages/setup-kubernetes/kubernetes-single-user.adoc
mkuznyetsov: ok

PR #601, opened 2018-11-19
Update pom.xml to use new Docker image for docs building.

rkratky:
### What does this PR do?

Updates `pom.xml` to use the `eclipse/che-docs` Docker image for docs building (introduced in #562).


benoitf (approved):
thx I forgot to update pom.xml

PR #639, opened 2018-12-17
fix loopback alias command

tkothe:
-  fix loopback alias command (`la0` -> `lo0`)
- add information how to access Che once it has launched


rkratky (approved):
Thanks, @tkothe.

PR #641, opened 2018-12-18
Add release notes about prometheus and grafana in the Helm chart.

metlos:
### What does this PR do?
See `$TITLE`

https://github.com/eclipse/che/issues/12137

### DIFF
--- a/release-notes/release-6.16.0.md
+++ b/release-notes/release-6.16.0.md
@@ -1,0 +1,33 @@
+## Release summary
+
+Eclipse Che 6.16 includes:
+
+* **Prometheus and Grafana deployed in the Helm chart**: Allowing to collect and inspect metrics emitted by the Che server
+
+## Upgrading
+
+Instructions on how to upgrade.
+
+## Release details
+
+### Prometheus and Grafana deployed in the Helm chart
+Eclipse Che server has been emitting a number of metrics since the release 6.15.0.
+To enable their easier visualization the Helm chart of the Che server now also deploys a Prometheus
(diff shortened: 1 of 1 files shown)


release-notes/release-6.16.0.md
```diff
@@ -0,0 +1,33 @@
...
+
+Instructions on how to upgrade.
+
+## Release details
+
+### Prometheus and Grafana deployed in the Helm chart
+Eclipse Che server has been emitting a number of metrics since the release 6.15.0.
+To enable their easier visualization the Helm chart of the Che server now also deploys a Prometheus
+server configured to scrape Che's metrics and a Grafana server pre-configured with a sample
+dashboard visualizing some of those metrics.
```
dneary: Suggested rewording:
"As part of our ongoing work to ensure that Eclipse Che is easier to monitor and trace, we have been adding a number of metrics which are tracked since the 6.15.0 release. To make it easier for Che admins to visualize these metrics, a Helm deployment of Eclipse Che now also deploys a Prometheus server which gathers metrics from Che, and Grafana,  a popular visualization project, pre-configured with a sample dashboard to visualize these metrics."

We might also consider wrapping this item with the two metrics which were added in PR #656 and PR #658.

metlos: Since this PR is for Che 6.16.0 and #656 and #658 are for 6.17.0 wouldn't it make more sense to keep them separate?

dneary (approved):
Looks good to me, thanks.

PR #647, opened 2018-12-21
Update recipes documentation to add deployments and configmaps

amisevsk:
### What does this PR do?
Update documentation for Kubernetes YAML recipes to include deployments and configmap as supported objects.

Main PR: https://github.com/eclipse/che/pull/12203

https://github.com/eclipse/che/issues/11505 and https://github.com/eclipse/che/issues/11507

### DIFF
--- a/src/main/pages/che-6/workspace-admin/recipes.adoc
+++ b/src/main/pages/che-6/workspace-admin/recipes.adoc
@@ -156,50 +156,78 @@ The `env_file` Compose option is not supported. Environment variables can be man
 [id="kubernetes-yaml-limitations-and-restrictions"]
 == Kubernetes YAML limitations and restrictions
 
-When a workspace is starting, Che creates a https://kubernetes.io/docs/concepts/workloads/pods/pod/[Kubernetes pod]. The following are limitatons and restrictions:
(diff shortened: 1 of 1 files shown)


src/main/pages/che-6/workspace-admin/recipes.adoc
```diff
@@ -156,50 +156,78 @@ The `env_file` Compose option is not supported. Environment variables can be man
 [id="kubernetes-yaml-limitations-and-restrictions"]
 == Kubernetes YAML limitations and restrictions
 
-When a workspace is starting, Che creates a https://kubernetes.io/docs/concepts/workloads/pods/pod/[Kubernetes pod]. The following are limitatons and restrictions:
+When a workspace is starting, Che creates various Kubernetes resources to support the IDE and development tools. Workspaces primarily consist of a https://kubernetes.io/docs/concepts/workloads/controllers/deployment/[Deployment] which runs a https://kubernetes.io/docs/concepts/workloads/pods/pod/[Kubernetes pod]. The following are limitatons and restrictions:
```
ibuziuk, line 159: @amisevsk maybe it would be a good idea to add info about replicas limitations that should only be `1` ?

amisevsk: We could; I didn't add it because I think it's kind of common sense, and the validator will block stack creation, and currently KubernetesDeployments will directly set replicas to 1 if you happen to modify it.

I would rather keep the docs as short as possible and only cover the weird cases we require.

ibuziuk (approved):

PR #649, opened 2018-12-21
Add release notes for the /activity endpoint changes.

metlos:
### What does this PR do?
`$TITLE`

https://github.com/eclipse/che/pull/12177
https://github.com/eclipse/che/pull/12091

Note that this PR actually **depends** on https://github.com/eclipse/che/pull/12177 so it should not be merged prior to that one.

### DIFF
--- a/release-notes/release-6.16.0.md
+++ b/release-notes/release-6.16.0.md
@@ -2,7 +2,7 @@
 
 Eclipse Che 6.16 includes:
 
-* **Update of the `/activity` REST endpoint**: It is now capable of returning information about
+* **Improved REST API endpoint `/activity`**: It is now capable of returning information about
 workspace activity.
 
 ## Upgrading
@@ -12,7 +12,7 @@ Instructions on how to upgrade.
 
 ## Release details
 
-### Update of the `/activity` REST endpoint
+### Improved REST API endpoint `/activity`
 
 The `/activity` endpoint now accepts `GET` queries for workspace IDs that have been in certain state
 for a certain amount of time. This can be used to query the server for workspaces that seem to have
--- a/release-notes/release-6.16.0.md
+++ b/release-notes/release-6.16.0.md
@@ -1,0 +1,39 @@
+## Release summary
+
+Eclipse Che 6.16 includes:
+
+* **Update of the `/activity` REST endpoint**: which is now capable of returning information about
+workspace activitity over time.
+
+## Upgrading
+
+Instructions on how to upgrade.
+
+
+## Release details
+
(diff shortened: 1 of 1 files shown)


slemeur:
cc @dneary

slemeur (approved):

release-notes/release-6.16.0.md
```diff
@@ -0,0 +1,39 @@
...
+## Upgrading
+
+Instructions on how to upgrade.
+
+
+## Release details
+
+### Update of the `/activity` REST endpoint (https://github.com/eclipse/che/pull/12177)
+
+The `/acitivity` endpoint can now return a set of workspace IDs that have been in certain state
```
dneary: s/acitivity/activity/

release-notes/release-6.16.0.md
```diff
@@ -0,0 +1,39 @@
...
+for a certain amount of time. This information is available only to users with a new `monitorSystem`
+permission.
+
```
dneary, line 66: Please move the PR link to the end, rather than having it in the heading.

release-notes/release-6.16.0.md
```diff
@@ -0,0 +1,39 @@
...
+workspace activitity over time.
+
+## Upgrading
+
+Instructions on how to upgrade.
+
+
+## Release details
+
+### Update of the `/activity` REST endpoint (https://github.com/eclipse/che/pull/12177)
```
dneary: Change title to "Improved REST API endpoint /activity"

In the description: What does "a certain state for a certain time" mean? How would a user use this?

I *think* this means "You can add optional arguments to the /activity REST endpoint to query workspace IDs which are in a specific state (STARTING, RUNNING, or STOPPING) for longer than a given time." - it would be great to give a sample of how this might be used, and how it impacts the end user or admin.

Thanks!

dneary (commented):
Looks good, but I would like to understand how this impacts the admin or user.

release-notes/release-6.16.0.md
metlos: >  What does "a certain state for a certain time" mean? How would a user use this?

I wanted to be brief in the release notes. The generated Swagger documentation for the endpoint should contain more details. See https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityService.java#L91 for the `@ApiOperation` and `@ApiParam` annotations.

metlos:
@dneary, I hope I applied all the requested changes. could you please review again?

dneary (approved):
Looks good to me. Thank you.

PR #675, opened 2019-02-05
Add release notes about mountSources.

metlos:
### What does this PR do?
This PR adds release notes about the `mountSources` attribute in `che-plugin.yaml`.

https://github.com/eclipse/che/pull/12510

### DIFF
--- a/release-notes/release-6.18.md
+++ b/release-notes/release-6.18.md
@@ -4,6 +4,7 @@ The Eclipse Che 6.18.0 release contains the following notable features:
  
 * Add Vue.js language support
 * Improvements for Persistent Volume Claim (PVC) handling
+* Support for `mountSources` attribute in `che-plugin.yaml`
 
 ---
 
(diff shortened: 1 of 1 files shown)


skabashnyuk:
CC @dneary. @slemeur @rkratky  can you take a look too?

slemeur (approved):
Nice work! 
@dneary make sure that this is callout as a feature/capability that is dedicated to Che 7.

dneary (commented):
Is this for 6.18 or 6.19? Just checking when the change landed.

skabashnyuk:
6.18

dneary:
Thanks - and can I have an example of how plug-ins worked when they did not have access to the source code? Did they have to request a text buffer from somewhere, or something? I'm looking for an illustrative example of how this enables something new for plug-in writer.

metlos:
@dneary - there was the possibility to manually define a volume mount that would make the sources available to the plugin. `mountSources` makes that process simpler, but more importantly it makes sure that all plugins get the sources mounted on the same location, as explained in the release note.

dneary (approved):
I think the text needs some clarification, but I will propose a revision in a follow-on commit. Thank you @metlos!

PR #693, opened 2019-03-18
Add initial set of Che 7 guides + user story templates.

rkratky:
### What does this PR do?

Adds initial outline for a new structure of Che 7 docs. Specifically, it adds three directories:

* `installation-guide`
* `administration-guide`
* `developer-guide`

and a number of user-story templates.

Existing Che 7 content will now start to be migrated into the new structure (using prepared templates or adding new templates as required).


slemeur (approved):

slemeur:
Let's proceed, nice step forward

PR #698, opened 2019-03-25
Che 7 structure, 2nd attempt

rkratky:
### What does this PR do?

Adds the complete (so far) Che 7 docs structure. All sections and topics are represented by (empty) template files (with the exception of the quick-start content that was already present -- this content is reviewed, migrated to the new structure, and added to the sidebar yaml to make it show among published docs).

The structure is based on https://hackmd.io/8YMVcGRHRnC2-ycA9xNkvg

@slemeur - a word of explanation: the structure preserves the main ideas of the previous outline, plus it:
* consolidates elements that would be confusing (e.g. sections Overview, Introduction, and Getting Started)
* topic titles have been converted to 'action' titles -- telling the user what will be accomplished
* some topics have moved under different chapters to streamline the user's path through the docs
* language and terminology have been unified for consistency

The structure outline can be modified if desired. When it is acked, I'll add the final structure to the sidebar config file to allow for (gradual) publishing.


rkratky:
@slemeur, @l0rd, would you agree to rename the *Plug-ins contribution points* topic to something like:
* Plug-in component definition reference
* New component metadata definition reference

The current title is rather cryptic for new users, but I don't want to change the meaning.

l0rd:
@rkratky for me it should be "Che extension points" as in https://wiki.eclipse.org/FAQ_What_are_extensions_and_extension_points%3F

slemeur:
Let's go with "extension points"

slemeur (approved):

slemeur:
LGTM

PR #730, opened 2019-06-12
Add SSH key admonition.

rkratky:
### What does this PR do?

Adds info about types of SSH keys to address issue with Gitlab keys.

Content by @nickboldt (thanks).


nickboldt (approved):
As this looks verbatim like what I had in the CRW 1.2 AG, +1 for my own content, which we confirmed works there.

nickboldt:
Does this need to be cherry picked to 6.19.x branch too?

PR #733, opened 2019-06-18
Devfile docs, import from https://redhat-developer.github.io/devfile/

rkratky:
### What does this PR do?

Adds devfile docs based on https://redhat-developer.github.io/devfile/

### DIFF
--- a/src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
+++ b/src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
@@ -1,27 +1,63 @@
 [id="starting-a-workspace-with-a-devfile_{context}"]
-= Starting a workspace with a Devfile
+= Starting a workspace with a devfile
 
-This paragraph is the procedure module introduction: a short description of the procedure.
+This section describes how to start a Che workspace using your own existing devfile.
 
-[discrete]
+
 .Prerequisites
+ 
+* A running instance of Eclipse Che.
+* An existing devfile.
 
-* A bulleted list of conditions that must be satisfied before the user starts following this assembly.
-* You can also link to other modules or assemblies the user must follow before starting this assembly.
-* Delete the section title and bullets if the assembly has no prerequisites.
 
-[discrete]
 .Procedure
 
-. Start each step with an active verb.
+There are several ways to launch a Che workspace based on a prepared devfile:
 
-. Include one command or action per step.
+.Devfile from Git
 
-. Use an unnumbered bullet (*) if the procedure includes only one step.
+Deploy the devfile to a Git source repository and then create a factory from this repository.
 
-[discrete]
-.Additional resources
+. Put a `devfile.yaml` file in the root of your Git repository.
+. Execute the factory by opening the following URL:
++
+[subs="+quotes"]
+----
+https://__<your-che-host>__/f?url=https://github.com/__<mygroup>__/__<myrepo>__
+----
 
-* A bulleted list of links to other material closely related to the contents of the procedure module.
-* For more details on writing procedure modules, see the link:https://github.com/redhat-documentation/modular-docs#modular-documentation-reference-guide[Modular Documentation Reference Guide].
-* Use a consistent system for file names, IDs, and titles. For tips, see _Anchor Names and File Names_ in link:https://github.com/redhat-documentation/modular-docs#modular-documentation-reference-guide[Modular Documentation Reference Guide].
+.Devfile from URL
+
+Execute a devfile by constructing the factory with the URL pointing to the raw content of the devfile, for example:
+
+[subs="+quotes"]
+----
+https://__<your-che-host>__/f?url=https://pastebin.com/raw/ux6iCGaW
+----
+
+.Devfile over the Che REST API
+
+Send a devfile to the dedicated REST API using `curl` or Swagger, which will create a new workspace and return its configuration:
+
+[subs="+quotes"]
+----
+$ curl -X POST  -H "Authorization: __<TOKEN>__" -H "Content-Type: application/yaml" -d _<devfile_content>_ https://__<your-che-host>__/api/devfile
+----
+
+.Devfile using the `chectl` tool
+
+Execute workspace from a devfile using the `workspace:start` parameter with the `chectl` tool as follows:
+
+[subs="+quotes"]
+----
+$ chectl workspace:start --devfile=devfile.yaml
+----
+
+Note that this way only works with local (same machine) devfiles -- URLs cannot be used with `chectl`.
+
+
+// .Additional resources
+// 
+// * A bulleted list of links to other material closely related to the contents of the procedure module.
+// * For more details on writing procedure modules, see the link:https://github.com/redhat-documentation/modular-docs#modular-documentation-reference-guide[Modular Documentation Reference Guide].
+// * Use a consistent system for file names, IDs, and titles. For tips, see _Anchor Names and File Names_ in link:https://github.com/redhat-documentation/modular-docs#modular-documentation-reference-guide[Modular Documentation Reference Guide].
--- a/src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
+++ b/src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
@@ -16,7 +16,7 @@ There are several ways to launch a Che workspace based on a prepared devfile:
 
 .Devfile from Git
 
-Deploy the devfile to a Git source repository and then create a factory from this repository.
+Add the devfile to a Git source repository and then create a factory from this repository.
 
 . Put a `devfile.yaml` file in the root of your Git repository.
 . Execute the factory by opening the following URL:
@@ -35,15 +35,6 @@ Execute a devfile by constructing the factory with the URL pointing to the raw c
 https://__<your-che-host>__/f?url=https://pastebin.com/raw/ux6iCGaW
 ----
 
-.Devfile over the Che REST API
-
-Send a devfile to the dedicated REST API using `curl` or Swagger, which will create a new workspace and return its configuration:
-
-[subs="+quotes"]
-----
-$ curl -X POST  -H "Authorization: __<TOKEN>__" -H "Content-Type: application/yaml" -d _<devfile_content>_ https://__<your-che-host>__/api/devfile
-----
-
 .Devfile using the `chectl` tool
 
 Execute workspace from a devfile using the `workspace:start` parameter with the `chectl` tool as follows:
@@ -52,9 +43,6 @@ Execute workspace from a devfile using the `workspace:start` parameter with the 
 ----
 $ chectl workspace:start --devfile=devfile.yaml
 ----
-
-Note that this way only works with local (same machine) devfiles -- URLs cannot be used with `chectl`.
-
 
 // .Additional resources
 // 
--- a/src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
+++ b/src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
@@ -1,27 +1,81 @@
 [id="writing-a-devfile-for-your-project_{context}"]
 = Writing a Devfile for your project
 
-This paragraph is the procedure module introduction: a short description of the procedure.
+This section describes how to create a minimal devfile for your project and how to include more than one projects in a devfile.
 
-[discrete]
-.Prerequisites
+// .Prerequisites
+// 
+// * A bulleted list of conditions that must be satisfied before the user starts following this assembly.
+// * You can also link to other modules or assemblies the user must follow before starting this assembly.
+// * Delete the section title and bullets if the assembly has no prerequisites.
 
-* A bulleted list of conditions that must be satisfied before the user starts following this assembly.
-* You can also link to other modules or assemblies the user must follow before starting this assembly.
-* Delete the section title and bullets if the assembly has no prerequisites.
 
-[discrete]
-.Procedure
+[id="preparing-a-minimal-devfile_{context}"]
+== Preparing a minimal devfile
 
-. Start each step with an active verb.
+A minimal devfile sufficient to run a workspace consists of the following parts:
 
-. Include one command or action per step.
+* Specification version
+* Name
 
-. Use an unnumbered bullet (*) if the procedure includes only one step.
+Without any further configuration, a workspace with the default editor will be launched along with its default plug-ins, which are configured on Che Server. Che-Theia is configured as the default editor along with the *Che Machine Exec* plug-in.
 
-[discrete]
+To get a more functional workspace, the following parts can be added:
+
+* List of components: development components and user runtimes
+* List of projects: source code repositories
+* List of commands: actions to manage the workspace components, such as running the development tools, starting the runtime environments, and others
+
+.Example of a minimal devfile with a project and a standard set of plug-ins (Chhe-Theia editor, exec plug-in)
+[source,yaml]
+----
+apiVersion: 1.0.0
+metadata:
+  name: petclinic-dev-environment
+projects:
+  - name: petclinic
+    source:
+      type: git
+      location: 'https://github.com/che-samples/web-java-spring-petclinic.git'
+components:
+  - alias: theia-editor
+    type: cheEditor
+    id: eclipse/che-theia/next
+  - alias: exec-plugin
+    type: chePlugin
+    id: eclipse/che-machine-exec-plugin/0.0.1
+----
+
+
+[id="specifying-multiple-projects-in-a-devfile_{context}"]
+== Specifying multiple projects in a devfile
+
+A single devfile can specify multiple projects. For each project, specify the type of the source repository, its location, and optionally also the directory to which the project should be cloned to.
+
+.Example of a devfile with two projects
+[source,yaml]
+----
+apiVersion: 1.0.0
+metadata:
+  name: example-devfile
+projects:
+- name: frontend
+  source:
+    type: git
+    location: https://github.com/acmecorp/frontend.git
+- name: backend
+  clonePath: src/github.com/acmecorp/backend
+  source:
+    type: git
+    location: https://github.com/acmecorp/backend.git
+----
+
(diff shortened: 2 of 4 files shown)


skabashnyuk:
@slemeur @l0rd can we deprecate https://redhat-developer.github.io/devfile/ and https://github.com/redhat-developer/devfile and https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/README-devfile.md after this pr would be merged?

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
```diff
@@ -1,27 +1,63 @@
...
+https://__<your-che-host>__/f?url=https://pastebin.com/raw/ux6iCGaW
+----
+
+.Devfile over the Che REST API
+
+Send a devfile to the dedicated REST API using `curl` or Swagger, which will create a new workspace and return its configuration:
+
+[subs="+quotes"]
+----
+$ curl -X POST  -H "Authorization: __<TOKEN>__" -H "Content-Type: application/yaml" -d _<devfile_content>_ https://__<your-che-host>__/api/devfile
```
skabashnyuk: This api would be deprecated soo. I'll provide a new example soon.

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
```diff
@@ -1,27 +1,81 @@
...
+[id="preparing-a-minimal-devfile_{context}"]
+== Preparing a minimal devfile
 
-. Start each step with an active verb.
+A minimal devfile sufficient to run a workspace consists of the following parts:
 
-. Include one command or action per step.
+* Specification version
+* Name
+
```
skabashnyuk, line 20: Should we put an example (a reference to example) here?

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
```diff
@@ -1,27 +1,81 @@
...
+Without any further configuration, a workspace with the default editor will be launched along with its default plug-ins, which are configured on Che Server. Che-Theia is configured as the default editor along with the *Che Machine Exec* plug-in.
+
+To get a more functional workspace, the following parts can be added:
+
+* List of components: development components and user runtimes
+* List of projects: source code repositories
+* List of commands: actions to manage the workspace components, such as running the development tools, starting the runtime environments, and others
+
+.Example of a minimal devfile with a project and a standard set of plug-ins (Chhe-Theia editor, exec plug-in)
```
skabashnyuk: Chhe-Theia ->  Che-Theia

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
```diff
@@ -1,27 +1,81 @@
...
+----
+apiVersion: 1.0.0
+metadata:
+  name: petclinic-dev-environment
+projects:
+  - name: petclinic
+    source:
+      type: git
+      location: 'https://github.com/che-samples/web-java-spring-petclinic.git'
+components:
```
skabashnyuk, line 48: components part can be omitted in minimal devfile.

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
```diff
@@ -1,23 +1,251 @@
...
+----
+components:
+  - alias: theia-editor
+    type: cheEditor
+    id: eclipse/che-theia/next
+----
+
+When `cheEditor` is missing, a default editor is provided along with its default plug-ins. The default plug-ins are also provided for an explicitly defined editor with the same `id` as the default one (even if it is a different version). Che-Theia is configured as default editor along with the *Che Machine Exec* plug-in.
+
+To specify that a workspace requires no editor, use the `editorFree:true` attribute in the devfile attributes.
```
skabashnyuk: Should we put an example (a reference to example) here?

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
```diff
@@ -1,23 +1,251 @@
...
+      entrypoints:
+      - parentName: mysqlServer
+        command: ['sleep']
+        args: ['infinity']
+      - parentSelector:
+          app: prometheus
+        args: ['-f', '/opt/app/prometheus-config.yaml']
+----
+
+The `entrypoints` list contains constraints for picking the containers along with the `command` and `args` parameters to apply to them. In the example above, the constraint is `parentName: mysqlServer`. which will cause the command to be applied to all containers defined in any parent object called `mysqlServer`. The parent object is assumed to be a top level object in the list defined in the referenced file, which is `app-deployment.yaml` in the example above.
```
skabashnyuk: Remove extra dot.
`parentName: mysqlServer`. which  - > `parentName: mysqlServer` which

skabashnyuk (commented):
in general LGTM

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
rkratky: Let's keep it until it can be replaced.

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
```diff
@@ -1,27 +1,63 @@
...
+Execute workspace from a devfile using the `workspace:start` parameter with the `chectl` tool as follows:
 
-. Include one command or action per step.
+[subs="+quotes"]
+----
+$ chectl workspace:start --devfile=devfile.yaml
+----
 
-. Use an unnumbered bullet (*) if the procedure includes only one step.
+Note that this way only works with local (same machine) devfiles -- URLs cannot be used with `chectl`.
```
l0rd: We should remove this sentence: chectl works with remote devfiles as well since a few weeks.

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
```diff
@@ -1,27 +1,63 @@
...
 
-[discrete]
 .Procedure
 
-. Start each step with an active verb.
+There are several ways to launch a Che workspace based on a prepared devfile:
+
+.Devfile from Git
+
+Deploy the devfile to a Git source repository and then create a factory from this repository.
```
l0rd: "Add the devfile to a Git source repository" is more appropriate then "Deploy"

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
```diff
@@ -1,27 +1,63 @@
...
+.Devfile from URL
+
+Execute a devfile by constructing the factory with the URL pointing to the raw content of the devfile, for example:
+
+[subs="+quotes"]
+----
+https://__<your-che-host>__/f?url=https://pastebin.com/raw/ux6iCGaW
+----
+
+.Devfile over the Che REST API
```
l0rd: I think we can remove the REST API section. All the use cases should be covered by factory + chectl.

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
l0rd: +1 to @skabashnyuk comment. Something as basic as

```yaml
apiVersion: 1.0.0
metadata:
  name: minimal-workspace
```

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
l0rd: +1 to @skabashnyuk comment (no need to specify theia and machine-exec because they are automatically added) but you may want to add java support:

```yaml
components:
  -
    type: chePlugin
    id: redhat/java/latest
```

l0rd: And it may make more sense to reference a well know example (not the one we own) like the original spring petclinic:

```yaml
projects:
  - name: petclinic
    source:
      type: git
      location: 'https://github.com/spring-projects/spring-petclinic.git'
```

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
rkratky: Well, there the petclinic example shortly after. But yes, a minimal example would be nice, too.

l0rd (approved):

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
metlos: Specification version is now called "API version" (or at least the element is called `apiVersion`) and name is nested inside a `metadata` element.

So I would reformulate this whole section to something like:

A minimal devfile sufficient to run a workspace consists of nothing more but its name and API version as in the following example:

```yaml
apiVersion: 1.0.0
metadata:
  name: MyDevfile
```

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
```diff
@@ -1,27 +1,63 @@
...
-* You can also link to other modules or assemblies the user must follow before starting this assembly.
-* Delete the section title and bullets if the assembly has no prerequisites.
 
-[discrete]
 .Procedure
 
-. Start each step with an active verb.
+There are several ways to launch a Che workspace based on a prepared devfile:
+
+.Devfile from Git
```
metlos, line 17: This whole section is specific to github and really only works with github.

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
metlos: The devfile posted like this cannot contain relative URLs to outside content, while the above 2 methods can. Maybe it is worth mentioning here (or somewhere else, if not already done).

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
metlos: This would mean to me that the same restriction as with the devfiles posted to the REST endpoint applies here, too, but confirmation from the chectl folks would be needed.

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
rkratky: I added a link to the section that describes the attribute.

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
rkratky: ok

slemeur:
> @slemeur @l0rd can we deprecate https://redhat-developer.github.io/devfile/ and https://github.com/redhat-developer/devfile and https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/README-devfile.md after this pr would be merged?

No we shouldn't. 
Devfile format is not intended to be only Che. Other tools should be able to leverage that format. So instead of removing and deprecating you should update the documentation of devfile to maintain an agnostic definition of the format - Che being one possible consumer.

How Che can leverage devfiles and things that are Che specifics should be keept in the Che docs.

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
rkratky: ack

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
rkratky: ok

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
rkratky: ack

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
rkratky: ok

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
```diff
@@ -1,23 +1,251 @@
 [id="devfile-reference_{context}"]
 = Devfile reference
 
-This paragraph is the reference module introduction and is only optional. Include it to provide a short overview of the module.
+This section contains devfile reference and instructions on how to use the various elements that devfiles consist of.
 
-A reference module provide data that users might want to look up, but do not need to remember. It has a very strict structure, often in the form of a list or a table. A well-organized reference module enables users to scan it quickly to find the details they want. AsciiDoc markup to consider for reference data:
+== Adding components to a devfile
 
+Each component in a single devfile must have a unique name.
```
metlos, line 8: "name" is misleading here. Each type of component has a different way of identification. plugins have ids, dockerimages have `image` and openshift/kubernetes lists have `reference` used as an identifier. Any two plugins with the same identifier need to be disambiguated by an `alias` which needs to be unique amongst all components in the devfile.

src/main/pages/che-7/end-user-guide/proc_writing-a-devfile-for-your-project.adoc
rkratky: added Mario's suggestion

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
```diff
@@ -1,23 +1,251 @@
...
+`parentName`:: the name of the parent object that (indirectly) contains the containers to override
+`parentSelector`:: the set of labels the parent object needs to have
+
+A combination of these constraints can be used to precisely locate the containers inside the referenced Kubernetes list.
+
+
+[id="component-type-dockerimage_{context}"]
+=== Component type: dockerimage
+
+A component type that allows to define a container image-based configuration of a container in a workspace. A devfile can only contain one component of the `dockerimage` type.
```
metlos, line 168: Is this still true, actually?

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
```diff
@@ -1,23 +1,251 @@
...
+           secure: 'true'
+           public: 'true'
+           discoverable: 'false'
+     memoryLimit: 1536M
+     command: ['tail']
+     args: ['-f', '/dev/null']
+----
+
+
+== Adding commands to a devfile
```
metlos, line 196: This will need to be expanded. We also support embedded editor command definitions (e.g. tasks.json and launch.json).

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
rkratky: Ah, ok. But isn't that an unwelcome limitation?

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
rkratky: Removed as per Mario's comment; it now works with remote.

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
rkratky: OK, so, identifiers can be `id`, `image`, or `reference`. And if any of those coincide, an `alias` is also needed?

Is one alias enough, or do all comps with same identifiers need an alias?

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
rkratky: No idea. @skabashnyuk, @l0rd?

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
metlos: 1 alias should be enough. Also, an alias is **required** if the component is referenced in some commands.

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
rkratky: PR, please? :)

src/main/pages/che-7/end-user-guide/proc_starting-a-workspace-with-a-devfile.adoc
metlos: Well, we don't have the necessary integration for other git hostings ATM. That said, obtaining the devfile from an URL is universal and works across the board - github (using the URL of the raw content), gitlab, pastebin, statically served sites, etc.

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
metlos: Yes, it still is :)

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
metlos: Just an issue, for now :) https://github.com/eclipse/che-docs/issues/734

src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
rkratky: > an alias is required if the component is referenced in some commands

I'm getting a bit lost... Could you please provide a list of all cases in which an alias is required?

l0rd: The sentence `Each component in a single devfile must have a unique name.` was there because in the past each component had a mandatory field `name`. `alias` is required only if you want to reference the component in a command.

PR #742, opened 2019-06-27
More Explicit Instructions on Changing /local/path in Che Single User Mode Instructions

danielhelfand:
### What does this PR do?

This pull request rearranges the Che single user mode instructions under quick start to more explicitly call out changing the `/local/path` in the `docker run` command to start up Che:

`docker run -ti -v /var/run/docker.sock:/var/run/docker.sock -v` **/local/path**`:/data eclipse/che start`

No issue created for this. This is an issue I ran into when setting up Che.

### DIFF
--- a/src/main/pages/che-6/overview/quick-start.adoc
+++ b/src/main/pages/che-6/overview/quick-start.adoc
@@ -37,13 +37,13 @@ $ docker run --rm --net host eclipse/che-ip:nightly
 
 .Procedure
 
-To run Che in single-user mode, enter:
+To run Che in single-user mode, run the command below. Note that `/local/path` in the command below needs to be changed and can be any path on your local machine where you want to store Che data and projects.
 
 ----
 $ docker run -ti -v /var/run/docker.sock:/var/run/docker.sock -v /local/path:/data eclipse/che start
 ----
 
-Note that `/local/path` can be any path on your local machine where you want to store Che data and projects. Once Che has started, it can be accessed at `localhost:8080`.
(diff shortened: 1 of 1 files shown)


slemeur:
Hi @danielhelfand, thanks a lot for your PR, much appreciate.

Before we can merge your PR, you need to agree with the Eclipse Contributor Agreement, you can get more information on the following wiki page:
https://github.com/eclipse/che/wiki/Eclipse-Contributor-Agreement

What version of Che are you using?

danielhelfand:
@slemeur I am using `7.0.0-RC-1.0`. I realize these edits are under `che-6`, but I couldn't find any of the quick start docs under `che-7`. Assuming the `che-6` docs still apply with regard to this.

slemeur:
@danielhelfand : With the version 7, the support of Docker is going to be deprecated to focus on Kubernetes. For the version 7, you should give a look at the following instructions: https://www.eclipse.org/che/docs/che-7/che-quick-starts.html
There is also an operator available from OperatorHub.

But it is still a great enhancement for anybody willing to use Che 6 ;)

slemeur (approved):

danielhelfand:
@slemeur Ah, I see. Looks like the Download button on the [getting started page](https://www.eclipse.org/che/getting-started/) goes to Che 6 docs. Should this point to the [Che 7 docs](https://www.eclipse.org/che/docs/che-7/che-quick-starts.html) you referenced above?

danielhelfand:
Never mind. I see 7 is still beta. Appreciate the help!

PR #788, opened 2019-08-14
Installing chectl as quickstart chapter

themr0c:
fix eclipse/che/#14237 

Installing chectl documentation becomes a quick start chapter and other guides refer to it


themr0c:
cc @benoitf

benoitf (approved):
very nice 🎉

themr0c:
Thank you for the site.baseurl fix!

PR #822, opened 2019-09-19
Fix https deploy doc

sleshchenko:
### What does this PR do?
1. Fixes typo
2. Aligns AWS YAML section options with Azure and GCP ones, to make * displayable. Maybe there are more preferable ways to make * displayable, but I just did in the same way as other similar docs have.

It fixes a documentation issue a used faced in https://github.com/eclipse/che/issues/14584


themr0c (approved):
LGTM

PR #828, opened 2019-09-24
Do Not Strip / From url for Search

danielhelfand:
Signed-off-by: danielhelfand <helfand.4@gmail.com>

### What does this PR do?

This pull request removes the stripping of the `/` from the url in `search.json` which causes a 404 whenever a search is currently done on the documentation site. 

Follow steps in the issue below to reproduce the issue. 

### What issues does this PR fix or reference?

Closes https://github.com/eclipse/che/issues/14647

### Additional Info

Originally opened as #827. Reopened with signed commit.

PR #829, opened 2019-09-26
Prepend site.baseurl in search.json

### LINKED ISSUE #14647: Search for Documentation Site Responds With 404 File not Found
When searching for something on the [Eclipse Che Documentation site](https://www.eclipse.org/che/docs/), you will receive a `404 File not Found` no matter what you search for. This appears to be related to the fact the following is being added to the url for each search `introduction-to-eclipse-che/che-7` for che-7 docs. 

For instance, if you try searching for OpenShift 4 installation instructions, you will get the following url: https://www.eclipse.org/che/docs/che-7/introduction-to-eclipse-che/che-7installing-che-on-openshift-4-from-operatorhub

The actual url for the page is as follows: https://www.eclipse.org/che/docs/che-7/installing-che-on-openshift-4-from-operatorhub

This is possibly related to the site navigation.

### Steps to reproduce

Search for anything via the search bar.

danielhelfand:
Signed-off-by: danielhelfand <helfand.4@gmail.com>

### What does this PR do?

#828 removed the stripping of the url from `search.json` as part of an effort to fix https://github.com/eclipse/che/issues/14647. While running locally with changes in #828 fixes the issue, it does not solve the problem for the che-docs site.

There appears to be a difference in how search works locally versus the site itself. The site requires `/che/docs` as part of the url while it does not locally.

This pull request prepends `/che/docs` in `search.json` as part of the url.

## Considerations

This change will mean that search will not work locally for the site. 

If there is a staging environment for the site, it would make sense to test there as opposed to verifying locally.

### DIFF
--- a/src/main/search.json
+++ b/src/main/search.json
@@ -11,7 +11,7 @@ search: exclude
 "title": "{{ page.title | escape }}",
 "tags": "{{ page.tags }}",
 "keywords": "{{page.keywords}}",
-"url": "{{ page.url }}",
+"url": "{{ page.url | prepend: '/che/docs' }}",
 "summary": "{{page.summary | strip }}"
 }
 {% unless forloop.last and site.posts.size < 1 %},{% endunless %}
@@ -24,7 +24,7 @@ search: exclude
 "title": "{{ post.title | escape }}",
 "tags": "{{ post.tags }}",
 "keywords": "{{post.keywords}}",
-"url": "{{ post.url }}",
+"url": "{{ post.url | prepend: '/che/docs' }}",
 "summary": "{{post.summary | strip }}"
 }
 {% unless forloop.last %},{% endunless %}
--- a/src/main/search.json
+++ b/src/main/search.json
@@ -11,8 +11,8 @@ search: exclude
 "title": "{{ page.title | escape }}",
 "tags": "{{ page.tags }}",
 "keywords": "{{page.keywords}}",
-"url": "{{ page.url | prepend: '/che/docs' }}",
-"summary": "{{page.summary | strip }}"
+"url": "{{ page.url | remove_first: '/' | prepend: site.baseurl }}",
+"summary": "{{ page.summary | strip }}"
 }
 {% unless forloop.last and site.posts.size < 1 %},{% endunless %}
 {% endunless %}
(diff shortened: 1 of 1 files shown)


danielhelfand:
This does not appear to be the issue actually. I'll do some further digging.

danielhelfand:
In reviewing this, it does look like the issue seems to be related to `/che/docs` not being included with the base url. If you go to the site now, you'll get a request url of `https://www.eclipse.org/che-7/installing-che-on-openshift-4-from-operatorhub/` if searching for the OpenShift 4 installation doc. 

So it refers to `https://www.eclipse.org/` as the base url. If there's a more elegant way to change `site.baseurl`, please let me know, but I do believe this should address the search issue.

src/main/search.json
```diff
@@ -11,7 +11,7 @@ search: exclude
 "title": "{{ page.title | escape }}",
 "tags": "{{ page.tags }}",
 "keywords": "{{page.keywords}}",
-"url": "{{ page.url }}",
+"url": "{{ page.url | prepend: '/che/docs' }}",
```
rkratky:
```suggestion
"url": "{{ page.url | remove_first: '/' | prepend: site.baseurl }}",
```

src/main/search.json
```diff
@@ -24,7 +24,7 @@ search: exclude
 "title": "{{ post.title | escape }}",
 "tags": "{{ post.tags }}",
 "keywords": "{{post.keywords}}",
-"url": "{{ post.url }}",
+"url": "{{ post.url | prepend: '/che/docs' }}",
```
rkratky:
```suggestion
"url": "{{ page.url | remove_first: '/' | prepend: site.baseurl }}",
```

rkratky (changes requested):
@danielhelfand, thanks for digging into this. When merging the #828 PR, I didn't realize it would affect the various environments differently.

We need `site.baseurl` to be `/` for local previews, `/che/docs/` for prod builds (because on eclipse.org, the docs are under `/docs/che/`, not in the webserver root), and finally `/docs/` for WAR builds (which is yet to be fixed; see https://github.com/eclipse/che/issues/14386).

I tested the solution I'm suggesting on both local and in a prod-like environment (you can do that by building with `run.sh -prod` and copying the contents of `_site/` to a `che/docs/` dir structure in your local web server's root). The clunky removal of the first `/` is needed because if we were to just prepend `site.baseurl`, the URL on local previews would start with `//`, which wouldn't work.

danielhelfand:
@rkratky Thanks for the detailed explanation and suggestions. I just pushed up a change, but let me know if there's anything else needed.

rkratky (approved):

rkratky:
Merged. Thanks for investigating this, @danielhelfand.

danielhelfand:
@rkratky Thanks! How soon will the changes be live?

danielhelfand:
Never mind, I see that it's published and looks to be working. Thanks again for the help.

PR #837, opened 2019-10-03
Add some known issues when upgrading from Che6 using the operator.

metlos:
### What does this PR do?

Adds notes about upgrade from Che6 using the operator to the installation guide.

https://github.com/eclipse/che/issues/14738

### DIFF
--- a/src/main/pages/che-7/installation-guide/assembly_updating-che.adoc
+++ b/src/main/pages/che-7/installation-guide/assembly_updating-che.adoc
@@ -28,7 +28,43 @@ This paragraph is the assembly introduction. It explains what the user will acco
 
 Include modules here.
 
+== Known issues
 
+=== Upgrading from Che 6
+
+If you installed Che 6 using the operator, you might have to do a couple of adjustments after the upgrade of
+the operator. Upon its upgrade the operator ventures on updating Che server and other components of
+the installation to the versions defined by the operator. Because of the change in how such upgrades are handled
+between Che 6 and Che 7 operator, this might result in some components being stuck in their old versions.
+
+To resolve this situation, it is necessary to manually update the images mentioned in the `checluster` custom resource
+describing the Che installation. Namely, it might be necessary to review the following components:
+
+|===
+| Component | CR attribute name | How to update
+| Postgresql | `spec.database.postgresImage` | The version of the Postgresql database actually hasn't changed. If you
+manually changed the image of the Postgresql server, the change will be preserved.
+| Keycloak | `spec.auth.identityProviderImage` | Che uses a customized version of Keycloak image that follows
+the versioning of Che itself. Please update the image from `eclipse/che-keycloak:6.19.0` to `eclipse/che-keycloak:7.1.0`
+if you upgraded to Che 7.1.0 or use any other corresponding version.
+|===
+
+=== Updating Che installation using the operator
+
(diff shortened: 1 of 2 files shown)


src/main/pages/che-7/installation-guide/assembly_updating-che.adoc
```diff
@@ -28,7 +28,43 @@ This paragraph is the assembly introduction. It explains what the user will acco
...
+If you installed Che 6 using the operator, you might have to do a couple of adjustments after the upgrade of
+the operator. Upon its upgrade the operator ventures on updating Che server and other components of
+the installation to the versions defined by the operator. Because of the change in how such upgrades are handled
+between Che 6 and Che 7 operator, this might result in some components being stuck in their old versions.
+
+To resolve this situation, it is necessary to manually update the images mentioned in the `checluster` custom resource
+describing the Che installation. Namely, it might be necessary to review the following components:
+
+|===
+| Component | CR attribute name | How to update
```
davidfestal: Some remarks: the main things to do, to pull the new default docker images for Che 7, is simply to **reset the CR docker images fields to an empty string**, which means: *use the current default*.
And this should be done for :
- Keycloak: field `identityProviderImage`
- Postgres: field `postgresImage` (because postgres images regularly change due to CVEs and there's not warranty that the CRW 2.0 postgres image will the be the same as the one of the last CRW 1.2.x release) 
- Pvc Jobs Image: field `pvcJobsImage` (same reason related the CVEs)

davidfestal (changes requested):

src/main/pages/che-7/installation-guide/assembly_updating-che.adoc
skabashnyuk: @davidfestal  @metlos  is there an issue for that ?

davidfestal: Here I'm sugesting changes to the current documentation PR. Why would we need an additional issue ? I already explained as part of issue comment https://github.com/eclipse/che/issues/14200#issuecomment-532384111 :

> the user should have to manually clean the frozen docker image versions in the CheCluster (set them back to empty string for Keycloak, Postgres, etc ...) just after upgrading the operator to 7.1.0

skabashnyuk: Ok. I just thought we need some additional coding.

metlos: @davidfestal good point. I have reworded the fix suggestions and added the missing properties to the table. Do you think it is OK like that?

rkratky (commented):
Submitted my suggestions as https://github.com/metlos/che-docs/pull/1

rkratky (approved):
Thanks, @metlos!

src/main/pages/che-7/installation-guide/assembly_updating-che.adoc
rkratky: @davidfestal, are you OK with the reworded text @metlos provided.

davidfestal (approved):

src/main/pages/che-7/installation-guide/assembly_updating-che.adoc
davidfestal: Yes, I just approved.

PR #886, opened 2019-10-30
Modify run.sh to work with podman

themr0c:
- Detect presence of podman and use it.
- Use docker as fallback.
- If podman is used for the first time, delete previous builds if they are owned by root.
- Remove the  `:Z ` mount option that don't behave well on podman.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -8,13 +8,45 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-if [ "$1" = "-prod" ]; then
-  echo "Building for production (publishing on eclipse.org/che/docs/). To preview"
-  echo -e "the docs locally, run the script ($0) without the '-prod' option.\n"
-  docker run --rm -v $(pwd)/src/main:/che-docs:Z eclipse/che-docs sh -c "cd /che-docs; jekyll clean; jekyll build --config _config.yml,_config-war.yml"
-else
-  echo "Building and serving the docs for local preview. To build for production"
-  echo "(publishing on eclipse.org/che/docs/), run the script with the '-prod' option:"
-  echo "$0 -prod"
-  docker run --rm -ti -p 35729:35729 -p 4000:4000 -v $(pwd)/src/main:/che-docs:Z eclipse/che-docs sh -c "cd /che-docs; jekyll clean; jekyll serve --livereload -H 0.0.0.0"
-fi
+set -e
+
+# Detect presence of podman. Fallback to docker
+RUNNER="$(command -v podman 2>/dev/null || command -v docker)"
+
+case "${RUNNER}" in
+  *podman)
+    echo "Using $RUNNER."
+    [ "$(find src/main/_site -user root -print -prune -o -prune 2>/dev/null)" ] && \
+      echo "Previous build was probably created with docker. We need root privileges to delete it." && \
+      sudo rm -rf src/main/_site/ 
+    ;;
+  *docker)
+    echo "podman not detected. Using fallback runner $RUNNER. This runner is deprecated, please install podman."
+    ;;
+  *)
+    echo "No runner detected. Please install podman."
+    ;;
+esac
+
+SRC_PATH="$(pwd)/src/main"
+
+case "$1" in
+  -prod)
+    echo "Building for production (publishing on eclipse.org/che/docs/). To preview"
+    echo "the docs locally, run the script ($0) without the '-prod' option."
+    $RUNNER run --rm \
+      -v "${SRC_PATH}":/che-docs \
+      eclipse/che-docs \
+      sh -c "cd /che-docs && jekyll clean && jekyll build --config _config.yml,_config-war.yml"
+    ;;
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -8,13 +8,45 @@
...
+
+case "${RUNNER}" in
+  *podman)
+    echo "Using $RUNNER."
+    [ "$(find src/main/_site -user root -print -prune -o -prune 2>/dev/null)" ] && \
+      echo "Previous build was probably created with docker. We need root privileges to delete it." && \
+      sudo rm -rf src/main/_site/ 
+    ;;
+  *docker)
+    echo "podman not detected. Using fallback runner $RUNNER. This runner is deprecated, please install podman."
```
benoitf: I'm not sure one runner should deprecate the other

benoitf:
maybe keep `:Z` option for docker (and empty option in case of podman)
It's not because podman doesn't handle well that we should remove it for all runners

themr0c:
@benoitf I am not sure why the `:Z` option is even present and if it is really necessary.

benoitf:
@themr0c selinux

run.sh
themr0c: Podman does not write files as the root user, I believe it is a good enough reason to use podman when it is present, and docker only as a fallback.

Maybe this wording would be better, posponing the deprecation of docker:
```suggestion
    echo "The preferred runner podman is not installed. Using fallback runner $RUNNER."
```

themr0c:
With `podman` I just had errors with the `:Z` option, better results without. Will investigate further. Thanks for the feedbak.

run.sh
```diff
@@ -8,13 +8,45 @@
...
+
+case "${RUNNER}" in
+  *podman)
+    echo "Using $RUNNER."
+    [ "$(find src/main/_site -user root -print -prune -o -prune 2>/dev/null)" ] && \
+      echo "Previous build was probably created with docker. We need root privileges to delete it." && \
+      sudo rm -rf src/main/_site/ 
+    ;;
+  *docker)
+    echo "The preferred runner podman is not installed. Using fallback runner $RUNNER."
```
benoitf, line 24: why is it preferred ?

themr0c: Because it does not run containers as root by default. I find it a very sufficient reason.

rkratky:
@themr0c, the `:Z` option is necessary. See (with `podman`):

```
$ getenforce
Enforcing

$ ./run.sh
Using /usr/bin/podman.
Building and serving the docs for local preview. To build for production
(publishing on eclipse.org/che/docs/), run the script with the '-prod' option:
./run.sh -prod
sh: cd: line 1: can't cd to /che-docs: Permission denied

$ sudo setenforce 0

$ getenforce
Permissive

$ ./run.sh
Using /usr/bin/podman.
Building and serving the docs for local preview. To build for production
(publishing on eclipse.org/che/docs/), run the script with the '-prod' option:
./run.sh -prod
Configuration file: /che-docs/_config.yml
           Cleaner: Nothing to do for /che-docs/_site.
           Cleaner: Nothing to do for /che-docs/.jekyll-metadata.
           Cleaner: Nothing to do for .sass-cache.
Configuration file: /che-docs/_config.yml
            Source: /che-docs
       Destination: /che-docs/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
   GitHub Metadata: No GitHub API authentication could be found. Some fields may be missing or have incorrect data.
                    done in 6.699 seconds.
 Auto-regeneration: enabled for '/che-docs'
LiveReload address: http://0.0.0.0:35729
    Server address: http://0.0.0.0:4000//
  Server running... press ctrl-c to stop.

$ sed -i 's|:/che-docs \\$|:/che-docs:Z \\|' run.sh

$ sudo setenforce 1

$ getenforce
Enforcing

$ ./run.sh
Using /usr/bin/podman.
Building and serving the docs for local preview. To build for production
(publishing on eclipse.org/che/docs/), run the script with the '-prod' option:
./run.sh -prod
Configuration file: /che-docs/_config.yml
           Cleaner: Removing /che-docs/_site...
           Cleaner: Nothing to do for /che-docs/.jekyll-metadata.
           Cleaner: Nothing to do for .sass-cache.
Configuration file: /che-docs/_config.yml
            Source: /che-docs
       Destination: /che-docs/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
   GitHub Metadata: No GitHub API authentication could be found. Some fields may be missing or have incorrect data.
                    done in 6.745 seconds.
 Auto-regeneration: enabled for '/che-docs'
LiveReload address: http://0.0.0.0:35729
    Server address: http://0.0.0.0:4000//
  Server running... press ctrl-c to stop.
```
The results are the same with `docker`.

rkratky:
@themr0c, what errors did you with `:Z`?

rkratky (approved):

rkratky:
@themr0c, merging now -- see if it works for ppl.

PR #893, opened 2019-10-31
Add a known issue to che 7.3.0 detailing that users should not instal…

tomgeorge:
…l Che in the default namespace

Signed-off-by: Tom George <tg82490@gmail.com>

### What does this PR do?

This issue adds a known issue to 7.3.0 about installing Che in the `default` openshift namespace.  Users should instead install Che in a new namespace specifically for Che.

https://github.com/eclipse/che/issues/14426


tomgeorge:
cc @ibuziuk @davidfestal

rkratky (approved):
Thanks, @tomgeorge

PR #913, opened 2019-11-13
Nov13

zeibura:
### What does this PR do?
A couple more minor changes to the Che4z installing page:
- link to devfile location reverted to code block showing the full path. A link implies that the user should download the file and this is not the case.
- link to Brightside documentation added, showing instructions for accessing the named PDF.

15165

### DIFF
--- a/src/main/pages/che-7/extensions/che4z-installing.adoc
+++ b/src/main/pages/che-7/extensions/che4z-installing.adoc
@@ -17,7 +17,7 @@ This page describes how to install the Eclipse Che4z basic stack.
 
 .Prerequisites
 
-* Eclipse Che 7.3.0
+* Eclipse Che 7.3.0 on Kubernetes
 
 .Procedure 
 
--- a/src/main/pages/che-7/extensions/che4z-installing.adoc
+++ b/src/main/pages/che-7/extensions/che4z-installing.adoc
@@ -25,7 +25,7 @@ This page describes how to install the Eclipse Che4z basic stack.
 
 . In a web browser, load `++https://++__<YOUR_CHE_HOST>__/f?url=__<DEV_FILE_LOCATION>__`.
 +
-* For the Eclipse Che4z basic stack, use the devfile `https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml`
+* For the Eclipse Che4z basic stack, use the `++https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml++` devfile.
 
 * For a purchased premium stack, use the devfile location specified in the PDF installation guide available from Broadcom Support. Refer to the link:http://techdocs.broadcom.com/content/broadcom/techdocs/us/en/ca-mainframe-software/devops/ca-brightside-enterprise/2-0/release-notes.html[Brightside documentation] ("Getting Started" section) for instructions on accessing the PDF.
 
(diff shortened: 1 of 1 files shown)


src/main/pages/che-7/extensions/che4z-installing.adoc
```diff
@@ -25,9 +25,9 @@ This page describes how to install the Eclipse Che4z basic stack.
 
 . In a web browser, load `++https://++__<YOUR_CHE_HOST>__/f?url=__<DEV_FILE_LOCATION>__`.
 +
-* For the Eclipse Che4z basic stack, use the link:https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml[`mainframe-basic-stack.yaml`] devfile.
+* For the Eclipse Che4z basic stack, use the devfile `https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml`
```
rkratky: Reason for this? We try to avoid naked URLs -- why is this needed? Also, using the `link:` notation is good practice for live links. As for word order, "use the foo devfile" is IMO better than "use the devfile foo".

zeibura: We provide the URL so that the user can copy and paste it, not so the user can click the link and download the file. Surely a codeblock is preferable to a link for this purpose?

zimlu02: Zeb is right - we don't want users to click on the link, we want them only to copy the string with URL.

rkratky: Ok, in that case, the link should not be live.
```suggestion
* For the Eclipse Che4z basic stack, use the `++https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml++` devfile.
```

src/main/pages/che-7/extensions/che4z-installing.adoc
```diff
@@ -17,17 +17,17 @@ This page describes how to install the Eclipse Che4z basic stack.
...
 
 . Log in to Eclipse Che.
 
 . In a web browser, load `++https://++__<YOUR_CHE_HOST>__/f?url=__<DEV_FILE_LOCATION>__`.
 +
-* For the Eclipse Che4z basic stack, use the link:https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml[`mainframe-basic-stack.yaml`] devfile.
+* For the Eclipse Che4z basic stack, use the devfile `https://github.com/eclipse/che-che4z/raw/1.0.0/mainframe-basic-stack.yaml`
 
-* For a purchased premium stack, use the devfile location specified in the PDF installation guide available from link:https://casupport.broadcom.com/download-center/download-center.html[Broadcom Support].
+* For a purchased premium stack, use the devfile location specified in the PDF installation guide available from Broadcom Support. Refer to the link:http://techdocs.broadcom.com/content/broadcom/techdocs/us/en/ca-mainframe-software/devops/ca-brightside-enterprise/2-0/release-notes.html[Brightside documentation] ("Getting Started" section) for instructions on accessing the PDF.
```
rkratky:
```suggestion
* For a purchased premium stack, use the devfile location specified in the PDF installation guide available from Broadcom Support. See the link:http://techdocs.broadcom.com/content/broadcom/techdocs/us/en/ca-mainframe-software/devops/ca-brightside-enterprise/2-0/release-notes.html[Brightside documentation] (the "Getting Started" section) for instructions on accessing the PDF.
```

rkratky (changes requested):

zeibura:
That's fine by us.

rkratky (approved):

PR #914, opened 2019-11-14
che #15192 Fixing wrong example for che custom properties

ibuziuk:
### What does this PR do?
Fixing wrong example for che custom properties

https://github.com/eclipse/che/issues/15192


rkratky (approved):
Thanks

PR #953, opened 2019-11-29
Replaced: Stacks/Devfiles

MichalMaler:

Fixing: Stacks/Devfiles

One word fix

### DIFF
--- a/src/main/pages/che-7/end-user-guide/assembly_customizing-developer-environments.adoc
+++ b/src/main/pages/che-7/end-user-guide/assembly_customizing-developer-environments.adoc
@@ -1,11 +1,11 @@
 ---
 title: Customizing developer environments
-keywords: 
+keywords:
 tags: []
 sidebar: che_7_docs
 permalink: che-7/customizing-developer-environments/
 folder: che-7/end-user-guide
-summary: 
+summary:
 ---
 
 :parent-context-of-customizing-developer-environments: {context}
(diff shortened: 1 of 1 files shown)


src/main/pages/che-7/end-user-guide/assembly_customizing-developer-environments.adoc
```diff
@@ -23,7 +23,7 @@ There are three different ways to extend {prod}:
 
 * *Che-Theia plug-ins* add capabilities to the Che-Theia IDE. They rely on plug-in APIs that are compatible with Visual Studio Code. The plug-ins are isolated from the IDE itself. They can be packaged as files or as containers to provide their own dependencies.
 
-* *Stacks* are pre-configured {prod-short} workspaces with a dedicated set of tools, which cover different developer personas. For example, it is possible to pre-configure a workbench for a tester with only the tools needed for their purposes.
+* *Devfiles* are pre-configured {prod-short} workspaces with a dedicated set of tools, which cover different developer personas. For example, it is possible to pre-configure a workbench for a tester with only the tools needed for their purposes.
```
rkratky, line 26: @MichalMaler, I don't understand this change. What motivates it? Devfiles are not equivalent to (nor replacements for) stacks.

rkratky: Seems I may be misguided. @l0rd, this substitution was suggested by QE (_"Stacks are not being used anywhere anymore. Since Che 7 the stacks have been deprecated and replaced by Devfiles instead. It's basically just a naming thing."_), but I'm confused by the fact that we still use stacks in CRW. What would you advice?

l0rd: "The stacks are dead, long live the stacks!" 

In Che v6 the stacks were defined in a big `stacks.json` file that was included with the che-server. That file doesn't exist anymore in Che v7.

In Che v7 the stacks are defined in the devfile registry that is a separate service. And every single devfile in the registry correspond to a stack.

The confusion comes from the fact that in Che v6 stacks and workspaces were defined using 2 different formats. With Che 7 the same format (the devfile) is used to define both. 

But a user opening the user dashboard won't notice changes: in Che v7 there is still a list of stacks to choose from as a starting point to create workspace.

I would leave `Stacks` here because the `Devfiles` are **not** "pre-configured workspaces".

rkratky: @l0rd This is really a great summary -- thank you. I'm copying it this to our glossary :)

PR #960, opened 2019-12-05
Added Stacks vs Devfiles - Concept

MichalMaler:

Adding the context for Stacks and Devfiles in Che


rkratky (approved):
Thanks @MichalMaler.

PR #964, opened 2019-12-10
By default, a user can start only one workspace at a time

themr0c:
* By default, a user can start only one workspace at a time
* Various cleanups (remove templates with no content, remove comments that come from templates)
* Explain sections following overview


src/main/pages/che-7/end-user-guide/con_workspaces-overview.adoc
```diff
@@ -15,13 +15,19 @@ summary:
...
 
-{prod} workspaces are managed by Kubernetes pods. Everything running in a {prod-short} developer workspace is running inside containers and is accessible from the embedded browser-based IDE. Because developer workspaces are containerized, they are highly portable. And because they only need a browser to be accessed, they are easily shareable.
+Kubernetes pods are managing each component of a {prod-short} workspace. 
+In other words everything running in a {prod-short} workspace is running inside containers.
+It makes a {prod-short} workspace highly portable.
 
+The embedded browser-based IDE is the point of access for everything running in a {prod-short} workspace.
+It makes a {prod-short} workspace easily shareable.
+
+By default, a user can start only one workspace at a time.
```
ibuziuk: could we point to the config option / env var with instructions for configuration? I think it should be part of https://www.eclipse.org/che/docs/che-7/advanced-configuration-options/

themr0c: OK, will do.

themr0c: I see: https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties#L62

Is this assumption true: The reference file in the documentation https://github.com/eclipse/che-docs/blob/master/src/main/pages/che-7/installation-guide/ref_che-configmaps-fields-reference.adoc is reflecting options that we can find in following files:  https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties and https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties 

Both set of files overlap but are not exactly the same: Is there another file in che source code containing *all* available options, that we could take as a better source of truth?

ibuziuk: https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties#L62 is the property
Both property files are the source of truth - multiuser just adds additional set of properties

themr0c: Latest commits are adding the option in /advanced-configuration-options/ and the link in the overview page.

rkratky (approved):

PR #998, opened 2019-12-19
rework building a custom registry image

themr0c:
Refresh building a custom registry image


src/main/pages/che-7/administration-guide/examples/che-build-the-devfile-registry.sh
```diff
@@ -0,0 +1 @@
+$ docker build -t my-devfile-registry .
```
benoitf: FYI, missing EOF on lot of files

benoitf (approved):

src/main/pages/che-7/administration-guide/examples/che-build-the-devfile-registry.sh
themr0c: Fixed EOF, thank you @benoitf

PR #1000, opened 2019-12-20
Clarifications in monitoring chapter

themr0c:
Reorganize the chapter.
Rewrite the Prometheus part.

### DIFF
--- a/src/main/pages/che-7/administration-guide/assembly_monitoring-che.adoc
+++ b/src/main/pages/che-7/administration-guide/assembly_monitoring-che.adoc
@@ -15,9 +15,9 @@ summary:
 
 :context: monitoring-che
 
-This chapter describes how to use additional tools to process data exposed as metrics by {prod-short}. 
+This chapter describes how to configure {prod-short} to expose metrics and how to build an example monitoring stack with external tools to process data exposed as metrics by {prod-short}. 
 
-The monitoring stack is not an official production-ready solution, but rather has an introduction purpose.
+include::proc_enabling-and-exposing-che-metrics.adoc[leveloffset=+1]
 
 include::proc_collecting-che-metrics-with-prometheus.adoc[leveloffset=+1]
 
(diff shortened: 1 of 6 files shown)


benoitf (approved):

src/main/pages/che-7/administration-guide/assembly_monitoring-che.adoc
```diff
@@ -15,41 +15,9 @@ summary:
...
-
-ifeval::["{project-context}" == "che"]
-Grafana offers a front-end "facade" with tools to create a various visual representation in the form of dashboards with various panels and graph types.
-endif::[]
-
-Note that this monitoring stack is not an official production-ready solution, but rather has an introduction purpose.
-
-.The structure of {prod-short} monitoring stack
-image::monitoring/monitoring-che-stack-structure.png[link="{imagesdir}/monitoring/monitoring-che-stack-structure.png"]
+This chapter describes how to use  additional tools to process data exposed as metrics by {prod-short}.
```
benoitf:
```suggestion
This chapter describes how to use additional tools to process data exposed as metrics by {prod-short}. 
```

src/main/pages/che-7/administration-guide/assembly_monitoring-che.adoc
```diff
@@ -15,41 +15,9 @@ summary:
 
 :context: monitoring-che
 
-{prod-short} can expose certain data as metrics, that can be processed by Prometheus
```
skabashnyuk, line 18: Can you explain the motivation of removing this information from assembly_monitoring-che.adoc

themr0c: Good point. This chapter should also be about Enabling the production of metrics in Che, indeed. I will make it more clear then.

themr0c: handled feedback from @skabashnyuk - add visibility to enabling and exposing che metrics in last commit

themr0c:
It is not clear to me where to find the `che-monitoring.yaml` necessary to configure Grafana.

I found this file, but as far as I understand, this file is not referencing Grafana: https://github.com/redhat-developer/che-monitoring/blob/master/openshift/che-monitoring.yaml

themr0c:
I will merge this PR as it is. We need more input to get further, but it is already in a better state now.

PR #1001, opened 2019-12-21
Update git SSL docs

vinokurig:
### What does this PR do?
Add description about how to add an SSL certificate to Che deployed via operator:
screenshot-0 0 0 0_4000-2019 12 (1)

https://github.com/eclipse/che/issues/15285


themr0c (approved):
Thank you for the update.

We will probably have rework some bits to handle each variant (Kubernetes, Helm) in a different procedure, bit it can happen later.

PR #1012, opened 2020-01-06
Exclude examples for license check

themr0c:
As discussed in https://github.com/eclipse/che-docs/pull/1005, not having a license in examples is a feature. 

This PR is adding the required exclude statements to the `pom.xml` file, so that `mvn clean install -U` can run nicely, without errors and warnings.


benoitf (approved):
👍

themr0c:
Thank you @benoitf  for your help on this task!

PR #1035, opened 2020-01-11
Fix nightly build script

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?
Fix nightly build script


sleshchenko (approved):
I don't really understand this change but LGTM if it fixes nightly builds =)

PR #1036, opened 2020-01-14
Fixing broken link

MichalMaler:
7.12.4. Configuring workspaces and resources

Fixing the link which leads to che.env file at the end of the first paragraph.

Updating the 1.2 guide [1] to link to the 6.19.x branch instead of master [2].

[1] https://access.redhat.com/documentation/en-us/red_hat_codeready_workspaces/1.2/html-single/administration_guide/index#configuring-workspaces-and-resources
[2] https://github.com/eclipse/che/blob/6.19.x/dockerfiles/init/manifests/che.env#L538

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,13 +21,13 @@
     </parent>
     <groupId>org.eclipse.che.docs</groupId>
     <artifactId>che-docs</artifactId>
-    <version>6.19.0</version>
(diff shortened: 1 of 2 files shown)


themr0c (approved):

themr0c (changes requested):
che-6 doc should not end up in master

PR #1041, opened 2020-01-15
Add script to autogenerate Che configuration properties documentation

mshaposhnik:
### What does this PR do?
Add script to autogenerate Che configuration properties documentation
Script produces `adoc` format, which can be converted to html using asciidoctor.
Example page is at https://mshaposhnik.github.io/

https://github.com/eclipse/che/issues/15356


che_docs.sh
```diff
@@ -0,0 +1,77 @@
...
+  if [ $? -ne 0 ]; then
+    echo "Failure: Cannot read multiuser.properties from URL $MULTIUSER_PROPERTIES_URL" >&2
+    exit 1
+  fi
+}
+
+parse_content() {
+  while read -r LINE
+  do
+    if [[ $LINE == '##'* ]]; then
```
sparkoo: could you please comment individual conditions? Like `starting with ## means we're starting new topic` etc.

che_docs.sh
```diff
@@ -0,0 +1,77 @@
...
+      VALUE=""
+    elif [[ ! -z $TOPIC ]]; then
+      IFS=$'=' read KEY VALUE <<< $LINE             # property split into key and value
+      ENV=${KEY^^}                                  # capitalize property name
+      ENV="+${ENV//_/__}+"                          # replace single underscores with double
+      ENV=${ENV//./_}                               # replace dots with single underscore
+      BUFF="$BUFF $ENV,\"$VALUE\",\"${DESCR_BUFF//\"/\'}\" $NEWLINE"   # apply key value and description buffer
+    fi
+  done <<< $RAW_CONTENT
+  echo "$BUFF" >> $OUTPUT_PATH
```
sparkoo: I'm not sure about just appending the content at the end of current file. What about having full table in extra file, generate it fully and import it from `static` adoc? Maybe warn if file already exist and add some force parameter to replace the file?

sparkoo (commented):
Please, output what the script is doing to stdout/err. Like print each line it's currently parsing (basically print in each if/else block of `parse_content`), what resource it is downloading, what version in pom.xml it found, etc. It will really help with debugging it later. It will fail at some point and it would be nice to have some output.

Good job btw, I've expected much worse for bash script parsing properties.

che_docs.sh
mshaposhnik: Ok

che_docs.sh
skabashnyuk:
```
Table from CSV data in file
|===
include::envvars.csv[]
|===
```
see https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/

mshaposhnik: I saw that. But we have abt. 20 tables, so it's inconvenient to generate 20 csv files and hardcode theirs names in template etc

benoitf:
FYI there are some warnings on https://www.shellcheck.net/ for the .sh script

mshaposhnik:
@themr0c @rkratky Guys pls take a look on this. Is such an approach OK when we have an static adoc with headers\titles and introduction word. And script will dynamically append the tables to it ?

che_docs.sh
skabashnyuk: ok for me.

themr0c:
Please put the script in the tools directory, and run shellcheck on it (it's almost completely shellcheck compliant ;already)).

che_docs.sh
themr0c: I prefer multiple CSV to include, and no editable documentation content in the script.
So we keep the asciidoc file editable "normally" without hiding it in a script.
It makes easier to understand that the data in the CSV is generated.

mshaposhnik: There is no editable content into script now. Bot for the multiple includes there is complexity since table quantity and  names are also dynamic so it's not know how to include them all in template. This is why i'm generating csv directly into template content.

themr0c: The script is altering  `src/main/pages/che-7/administration-guide/ref_configuring-system-variables.adoc`. We will have mixed content inside of it, and high risks to commit it by error. That is precisely what we can avoid with includes.

As an alternative, we could add a static include statement  and a fully generated file  with static name `src/main/pages/che-7/administration-guide/examples/system-variables.adoc`, added in .gitignore.

NB: titles shoud be like that

```
.My title
```

instead of the `== My title` form.

mshaposhnik: Static include makes sense. Refactored.

tools/enviromnent_docs_gen.sh
```diff
@@ -0,0 +1,84 @@
...
+NEWLINE=$'\n'
+NEWLINEx2="$NEWLINE$NEWLINE"
+TABLE_HEADER="$NEWLINEx2,=== $NEWLINE Environment Variable Name,Default value, Description $NEWLINE"
+TABLE_FOOTER=",=== $NEWLINEx2"
+BUFF=""
+OUTPUT_PATH="../src/main/pages/che-7/administration-guide/examples/system-variables.adoc"
+
+fetch_current_version() {
+  echo "Trying to read current product version from pom.xml..." >&2
+  CURRENT_VERSION=$(grep -ri "<version>" ../pom.xml | tail -n 1 |sed -e "s/^[ \t]*<version>\([^<]*\)<.*$/\1/")
```
themr0c: grep: ../pom.xml: No such file or directory

Maybe use something using `dirname "${BASH_SOURCE[0]}"` or something like that instead?

themr0c (changes requested):
You need to close the last table in the generated file with an ending line:```,===```

```./tools/enviromnent_docs_gen.sh
...

Processing done. Output file is /home/remote/ffloreth/src/gh/themr0c/che-docs/src/main/pages/che-7/administration-guide/examples/system-variables.adoc
./run.sh
...
asciidoctor: WARNING: system-variables.adoc: line 281: unterminated table block
...
```

mshaposhnik:
Any more comments?

themr0c:
@mshaposhnik these were the last ones frm myself. It is a very nice and helpful contribution, much appreciated!

sparkoo:
@themr0c if you're ok with patch, could you please approve the PR?

themr0c (approved):

PR #1045, opened 2020-01-17
prod-url-replacement-first-batch

MichalMaler:

Using `  prod-url: http(s)://che-host:che-port` and improoving consistency of used links.


themr0c (approved):
Nice, thanks!

PR #1046, opened 2020-01-17
Substitutions: chectl -> {prod-cli}

themr0c:
- Replace occurrences of `chectl` by the `prod-cli` attribute.
- Execute `tools/substitutions.sh` to perform all attributes substitutions on current state of the documentation.
- Fix errors due to attributes substitutions.
- New script `tools/validate_attributes_processing.sh` to validate that all attributes are processed correctly in the HTML output.


yhontyk:
@themr0c should the file `src/main/pages/che-7/administration-guide/examples/system-variables.adoc` be part of this PR?

themr0c:
(Currently) `src/main/pages/che-7/administration-guide/examples/system-variables.adoc` is generated at runtime, but not committed to the repository. So we should either run the substitutions at runtime after generating `system-variables.adoc`, either challenge the idea to not commit the `system-variables.adoc` file.

src/main/pages/che-7/end-user-guide/assembly_adding-tools-to-che-after-creating-a-workspace.adoc
```diff
@@ -22,14 +22,14 @@ When installed in the workspace, {prod-short} plug-ins bring new capabilities to
 * Communicating with multiple databases directly from the IDE.
 * Enhanced code navigation, auto-completion and error highlighting.
 
-This chapter provides basic information about {prod-short} plug-in installation, enabling, and use in {prod-short) workspaces.
+This chapter provides basic information about Che Plugin installation, enabling, and use in {prod-short} workspaces.
```
Katka92: Would it be better to have `{prod-short} Plugin` here?

Katka92 (approved):

src/main/pages/che-7/end-user-guide/assembly_adding-tools-to-che-after-creating-a-workspace.adoc
themr0c: It is indeed quite difficult to do the automated processing of all the  Plugin/plug-in occurrences.

- On one hand we have the generic family for all plug-ins that we should write `{prod-short} plug-in` and `{prod-short} plug-ins`.
- On the other hand we have a specific `type` of plug-in which is called `Che Plugin`  in the configuration files, and in that case attribute substitution is not an option.

themr0c:
```suggestion
This chapter provides basic information about {prod-short} plug-in installation, enabling, and use in {prod-short} workspaces.
```

PR #1050, opened 2020-01-21
fix che-docs factory URL

themr0c:
fix che-docs factory URL


ibuziuk:
@themr0c have you verified this on Hosted Che? does all the docs use-cases work as expected?

themr0c:
On Hosted Che:

- It loads with Ephemeral mode set to true.
- Task 'build che-docs (plugin)' has exited with code 0.
- The task “serve che-docs” is running OK, but redirection to http://0.0.0.0:4000 makes the service unusable.

ibuziuk:
@themr0c should work also with PVC based workspaces. Could you create a separate issue for `serve che-docs` command?

themr0c:
[dogfooding] wrong redirection in serve che-docs task:  https://github.com/eclipse/che/issues/15865

Maybe we can merge this PR right now?

slemeur (approved):

PR #1058, opened 2020-01-29
Add newdoc, test-adoc.sh and vale to the che-docs container

themr0c:
Add `newdoc`, `test-adoc.sh`and  `vale` to the che-docs container.


benoitf (approved):
nice improvements 👍

PR #1069, opened 2020-02-10
Execute environment variable docs generation during publish docs ci process

skabashnyuk:
### What does this PR do?
Execute environment variable docs generation during publish docs ci process

Docs published without environment variables page content
Знімок екрана 2020-02-10 о 10 02 00

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -67,7 +67,7 @@ spec:
       steps {
         container('che-docs') {
             dir('che-docs') {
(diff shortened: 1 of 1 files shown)


skabashnyuk:
NOTE: change is not tested since this code https://github.com/eclipse/che-docs/blob/master/tools/enviromnent_docs_gen.sh#L24 is not working on macOS.

themr0c (approved):

PR #1072, opened 2020-02-12
New attribute `prod-url-secure`

yhontyk:
### What does this PR do?
Adds conditionals to differentiate between che and CRW - where default project name is mentioned.

We don't have an attribute for `che-che`/`codeready-workspaces` so I decided to use conditionals.

https://issues.redhat.com/browse/RHDEVDOCS-1548

### DIFF
--- a/src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
+++ b/src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
@@ -1,5 +1,4 @@
 [id="finding-{prod-id-short}-cluster-url-using-openshift-4-cli-tools_{context}"]
-
 = Finding {prod-short} cluster URL using the OpenShift 4 CLI
 This section describes how to obtain the {prod-short} cluster URL using the OpenShift 4 CLI (command line interface). The URL can be retrieved from the OpenShift logs or from the `checluster` Custom Resource.
 .Prerequisites
--- a/src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
+++ b/src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
@@ -1,4 +1,5 @@
 [id="finding-{prod-id-short}-cluster-url-using-openshift-4-cli-tools_{context}"]
+
 = Finding {prod-short} cluster URL using the OpenShift 4 CLI
 This section describes how to obtain the {prod-short} cluster URL using the OpenShift 4 CLI (command line interface). The URL can be retrieved from the OpenShift logs or from the `checluster` Custom Resource.
 .Prerequisites
--- a/src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator-and-ssl.adoc
+++ b/src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator-and-ssl.adoc
@@ -34,5 +34,4 @@ To create the {prod-short} instance on demo OpenShift clusters that have not bee
 ----
 Command server:start has completed successfully.
 ----
-
-. Navigate to the {prod-short} cluster instance. The domain is now prefixed with HTTPS and using _Let’s Encrypt_ ACME certificates: `++https://++che-che.__<OPENSHIFT_APPS_URL>__`.
+. Navigate to the {prod-short} cluster instance. The domain is now prefixed with HTTPS and using _Let’s Encrypt_ ACME certificates: `{prod-url-secure}`.
--- a/src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator.adoc
+++ b/src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator.adoc
@@ -40,4 +40,4 @@ $ {prod-cli} server:start --platform=openshift --installer=operator \
 Command server:start has completed successfully.
 ----
 
-. Navigate to the {prod-short} cluster instance: `++http://++che-che.__<OPENSHIFT_APPS_URL>__`.
+. Navigate to the {prod-short} cluster instance: `{prod-url}`.
--- a/README.md
+++ b/README.md
@@ -23,7 +23,24 @@ Then, instead of the `run.sh` script, run Maven directly:
 $ mvn clean install -Pnative
 ```
 
-## Adding a new page
+## Adding a new page with `newdoc`
+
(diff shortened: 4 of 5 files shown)


src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
```diff
@@ -20,3 +20,15 @@ time="2019-07-12T09:50:13Z" level=info msg="Custom resource eclipse-che updated"
 time="2019-07-12T09:50:13Z" level=info msg="{prod} is now available at: http://che-che.apps.cluster-fre-f0a2.fre-f0a2.openshiftworkshop.com"
 time="2019-07-12T09:50:13Z" level=info msg="Updating eclipse-che CR with version: 7.0.0-RC-2.0"
 ----
+endif::[]
+ifeval::["{project-context}" == "crw"]
+* Wait for the logs to show the message `{prod} is now available at:` followed by {prod-short} URL:
++
+[subs="+quotes,+attributes",options="nowrap"]
+----
+time="2019-07-12T09:50:13Z" level=info msg="Updating eclipse-che CR with {prod} server URL: http://codeready-workspaces.apps.cluster-fre-f0a2.fre-f0a2.openshiftworkshop.com"
```
nickboldt: is this block really needed? If it's the same in upstream too, then you could probably just define a new variable called `{deployment-route}` and set it to `che-che` or `codeready-workspaces` based on value of `{project-context}` variable.

src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
```diff
@@ -20,3 +20,15 @@ time="2019-07-12T09:50:13Z" level=info msg="Custom resource eclipse-che updated"
 time="2019-07-12T09:50:13Z" level=info msg="{prod} is now available at: http://che-che.apps.cluster-fre-f0a2.fre-f0a2.openshiftworkshop.com"
 time="2019-07-12T09:50:13Z" level=info msg="Updating eclipse-che CR with version: 7.0.0-RC-2.0"
 ----
+endif::[]
+ifeval::["{project-context}" == "crw"]
+* Wait for the logs to show the message `{prod} is now available at:` followed by {prod-short} URL:
```
nickboldt: this line is the same in the che block above.

src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
```diff
@@ -20,3 +20,15 @@ time="2019-07-12T09:50:13Z" level=info msg="Custom resource eclipse-che updated"
...
++
+[subs="+quotes,+attributes",options="nowrap"]
+----
+time="2019-07-12T09:50:13Z" level=info msg="Updating eclipse-che CR with {prod} server URL: http://codeready-workspaces.apps.cluster-fre-f0a2.fre-f0a2.openshiftworkshop.com"
+time="2019-07-12T09:50:13Z" level=info msg="Custom resource eclipse-che updated"
+time="2019-07-12T09:50:13Z" level=info msg="{prod} is now available at: http://codeready-workspaces.apps.cluster-fre-f0a2.fre-f0a2.openshiftworkshop.com"
+time="2019-07-12T09:50:13Z" level=info msg="Updating eclipse-che CR with version: 7.0.0-RC-2.0"
```
nickboldt: Should we maybe update this from 7.0.0 RC2 (about a year old, unsupported, pre-GA) to a more current actual version, like 7.8.0? Or if this is a CRW version, then 2.0.0?

nickboldt (changes requested):
see suggestions inline.

src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator.adoc
```diff
@@ -39,5 +39,9 @@ $ {prod-cli} server:start --platform=openshift --installer=operator \
 ----
 Command server:start has completed successfully.
 ----
-
+ifeval::["{project-context}" == "che"]
 . Navigate to the {prod-short} cluster instance: `++http://++che-che.__<OPENSHIFT_APPS_URL>__`.
```
themr0c: I believe this is exactly a case where the use of the {prod-url} attribute would  be beneficial. No conditional would be even necessary.

```suggestion
. Navigate to the {prod-short} cluster instance: `{prod-url}`.
```

src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator-and-ssl.adoc
```diff
@@ -34,5 +34,9 @@ To create the {prod-short} instance on demo OpenShift clusters that have not bee
 ----
 Command server:start has completed successfully.
 ----
-
+ifeval::["{project-context}" == "che"]
 . Navigate to the {prod-short} cluster instance. The domain is now prefixed with HTTPS and using _Let’s Encrypt_ ACME certificates: `++https://++che-che.__<OPENSHIFT_APPS_URL>__`.
```
themr0c: Using the {prod-url} attribute is clearly not an option here. What about introducing and using a `{prod-url-secure}` attribute with value `++https://che-host:che-port++` ? NB: attributes are defined in `src/main/_config.yml `.

src/main/pages/che-7/overview/proc_finding-che-cluster-url-using-openshift-4-cli-tools.adoc
```diff
@@ -10,7 +10,7 @@ This section describes how to {prod-short} cluster URL using OpenShift 4 CLI too
 
 [discrete]
 == Procedure
-
+ifeval::["{project-context}" == "che"]
```
themr0c: If here only the logs are different, what about externalizing the logs in `include/examples/{project-context}something.log` instead of the ifs?

themr0c (changes requested):

src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator.adoc
nickboldt: 👍

yhontyk:
@nickboldt @themr0c  Apparently, a lot of the content in these procedures was changed in https://github.com/eclipse/che-docs/pull/1063 (And I even reviewed it... :woman_facepalming: )

I rebased on master and added attributes where applicable though.

PR #1140, opened 2020-03-16
Added new terms for vale to ignore and new styles

yhontyk:
### What does this PR do?
Adds words to be ignored by the vale spellcheck, new styles, removes styles that are not applicable to our documentation.

https://issues.redhat.com/browse/RHDEVDOCS-1595


rkratky (approved):
@boczkowska, LGTM, but why did you remove all the rules? Some of them seem useful, e.g. ComplexWords.

yhontyk:
@rkratky ComplexWords are unsuitable for our documentation IMO, too many false positives. See my comment in Jira. 
For example:
`many` is not an alternative for `multiple` in many of the contexts that we use it, `additional` cannot always be replaced by `extra` or `more`, and so on.

We can compile our own list of suggested simpler alternatives but when testing on actual docs, I found this rule not very helpful. Feel free to re-run tests and we can bring it back if you find it suitable.

PR #1145, opened 2020-03-17
Fixing broken keycloak links, attributes fixes.

yhontyk:
### What does this PR do?
Broken keycloak links. Introduces new attrbute for the URL and the version. Verison confirmed with the QA. We will have to keep an eye on it since it's not the latest.

https://issues.redhat.com/browse/RHDEVDOCS-1665


yhontyk:
Fixed in downstream too.

themr0c (approved):

PR #1155, opened 2020-03-22
Fix lang, formatting in 'Che resources'.

rkratky:
### What does this PR do?

Fixes lang., formatting in "Che resources".


src/main/pages/che-7/administration-guide/con_a-workspace-example.adoc
```diff
@@ -32,50 +32,48 @@ components:
...
-|Default Memory Request
+|Container name
+|Default memory limit
+|Default memory request
 
 |Workspace
 |theia-ide (default `cheEditor`)
-|512Mi
-|512Mi
+|512 Mi
```
yhontyk: This is probably my ignorance but what are these units exactly? I assume they're mebibyes but `mebi`and other binary units are abbreviated as `MiB`, `GiB` and so on, as far as I know.

rkratky: Good point. I didn't realize this. Will fix.

rkratky: Pushed a fix (converted to MiB etc.).

yhontyk (approved):

PR #1158, opened 2020-03-23
Fixing the headings in EuG

MichalMaler:

* Checking EUG IDs uniqueness
* Adding missing `_{context}` attribute to assembly headings
* Adding context information for included files
* Adding :page-liquid: where needed
* Deleting unnecessary content


yhontyk:
:heart: :heart: :heart: 
Maybe we'll get working contexts in the end! I need to build the PR first.

Why was the `src/main/pages/che-7/end-user-guide/proc_starting-a-new-workspace-from-user-dashboard.adoc` procedure deleted though? It also references image file, do we need to keep it?

MichalMaler:
@boczkowska , I am checking if the file is included or linked to somewhere. If not, I am comparing that file against our woking structure https://github.com/eclipse/che/issues/14415 .
If the file aint involved anywhere and it is not even a part of the god's plan, I am nuking that kid from the orbit, son :)

Good catch with these pictures. If they are not involved anywhere else, we can surely delete them.
But this particular one is involved in a working module, so we will leave that one be.
image

yhontyk:
@MichalMaler harsh! :D Do you think it could be a part of https://www.eclipse.org/che/docs/pages/che-7/end-user-guide/proc_creating-a-workspace-from-user-dashboard.html this user story? Aka you create and the configure.

I need time to read through the actual docs.

We had it for something, maybe we just forgot to include it properly.

yhontyk:
@MichalMaler I've looked through the files again. The content is actually duplicated in `creating-a-workspace-from-user-dashboard` so you were right to delete them.

There are some merge conflicts, resolve them and let's merge.

yhontyk (approved):

PR #1159, opened 2020-03-23
Fixing the headings in CG,Extensions,IG

MichalMaler:

* Checking `IG,ContrG,Extensions` IDs uniqueness
* Adding missing _{context} attribute to assembly headings
* Adding context information for included files
* Adding :page-liquid: where needed
* Deleting unnecessary content


src/main/pages/che-7/contributor-guide/ref_che-theia-plug-in-api.adoc
```diff
@@ -8,6 +8,7 @@ folder: che-7/contributor-guide
 summary:
 ---
 
+:page-liquid:
 
 [id="che-theia-plug-in-api_{context}"]
```
yhontyk, line 13: I don't get this one and a few other modules like this. Where is the context defined or inherited from? :( @rkratky @themr0c 

It's headings render with an anchor like this `#theia-namespace_{context}`...

yhontyk: @MichalMaler could you document somewhere where `:page-liquid:` should be defined and what it does? For example, in the contrib draft in our devtools gdrive. Ping me if you can't find it.

yhontyk:
@MichalMaler I left some comments but I honestly don't think I should be reviewing this. =\ I really don't understand our setup with contexts...

src/main/pages/che-7/contributor-guide/assembly_setup-che-in-tls-mode-with-self-signed-certificate.adoc
```diff
@@ -14,13 +14,12 @@ summary:
 [id="setup-{prod-id-short}-in-tls-mode_{context}"]
 = Setup {prod-short} in TLS mode with self-signed certificates
```
yhontyk, line 15: Ids and titles should match (ideally)

src/main/pages/che-7/installation-guide/con_target-platforms.adoc
```diff
@@ -1,17 +1,12 @@
 [id="target-platforms_{context}"]
```
yhontyk, line 1: If it's not included anywhere, should we get rid of it?

src/main/pages/che-7/installation-guide/proc_configuring-che-persistent-volumes-strategy.adoc
```diff
@@ -1,27 +1,17 @@
 [id="configuring-{prod-id-short}-persistent-volumes-strategy_{context}"]
```
yhontyk, line 1: Same as the above comment. Should we get rid of it or is it a placeholder for some planned content?

yhontyk:
@MichalMaler I've looked through the build and those contexts in `ref_che-theia-plug-in-api.adoc` should definitely be fixed. I have filed a separate issue for it: https://issues.redhat.com/browse/RHDEVDOCS-1746, since this PR mostly deals with other stuff.

Pls address some of my other comments, if necessary, resolve conflict and let's merge this baby.

src/main/pages/che-7/installation-guide/con_target-platforms.adoc
MichalMaler: This will be deleted.

src/main/pages/che-7/installation-guide/proc_configuring-che-persistent-volumes-strategy.adoc
MichalMaler: I am reworking this content. Will put it in Che `quick-starts` when ready.

MichalMaler: Oh, that's a different doc with a similar name..but anyway, it has its place in Installation Guide. Here is a plan for it:
https://github.com/eclipse/che/issues/14362

src/main/pages/che-7/installation-guide/con_target-platforms.adoc
rkratky: @MichalMaler, please, before deleting existing content, always track down its origin (who merged and why, associated issue, etc.). It may be that the content is needed, but it wasn't properly incorporated into the docs structure.

yhontyk: @rkratky there's no content, it's an empty file. It was either a placeholder for something or an artifact. Both should not be in the master ideally.

MichalMaler: This was an empty template. I am checking it with doc plans. If it isn't there and @themr0c doesn't know about it, I am deleting it. In a case it's actual content, I do the same, buck backing up the content and asking you or Radim for further action.

src/main/pages/che-7/contributor-guide/ref_che-theia-plug-in-api.adoc
rkratky: @boczkowska 

> could you document somewhere where :page-liquid: should be defined and what it does?

See https://github.com/asciidoctor/jekyll-asciidoc#enabling-liquid-preprocessing for background.

> I don't get this one and a few other modules like this. Where is the context defined or inherited from? :( @rkratky @themr0c

Every assembly should set its context, and modules should be included from assemblies.

> It's headings render with an anchor like this #theia-namespace_{context}...

That's a mistake (because an ID like that is useless and won't compile with ccutil). It is most likely caused by one of the following:
* The parent assembly doesn't set context properly.
* The module is used as stand-alone (i.e. not included from any assembly), but it doesn't set its context. So, it should either define its own context (like assemblies do), or be included from some assembly.

yhontyk: >That's a mistake (because an ID like that is useless and won't compile with ccutil). It is most likely caused by one of the following:

I filed an issue for it, see my comment down the thread. This is obviously a mistake but we should deal with it in a separate PR.

rkratky:
@boczkowska 

> I left some comments but I honestly don't think I should be reviewing this. =\ I really don't understand our setup with contexts...

Will you be reviewing Michal's updates, or do you want to offload this?

yhontyk (approved):

PR #1166, opened 2020-03-25
vale setup enhancements

themr0c:
* Move Vale styles into the tools directory
* Other small vale enhancements


yhontyk:
This should be merged first and then rebased on - https://github.com/eclipse/che-docs/pull/1162
I tested wrote-good style yesterday and removed some of them

PR #1200, opened 2020-04-06
Fix prod id short in ids

themr0c:
* Fix Jekyll config
* Attributes substitutions `./tools/substitutions.sh`
* Use the {prod-id-short} attribute in ids


rkratky (approved):

themr0c:
"Use the {prod-id-short} attribute in ids" commit already cherry-picked into 7.9.x.

PR #1216, opened 2020-04-14
Remove tls from installation intructions.

yhontyk:
### What does this PR do?
Removes `--tls` flag from installation commands and change `tlsSupport: false` to `tlsSupport: true`


dmytro-ndp (approved):
Looks good.

PR #1237, opened 2020-04-20
External PostgreSQL database need an external {identity-provider}.

themr0c:
When using an external PostgreSQL database, it is also necessary to use an external {identity-provider}.


yhontyk (approved):

ibuziuk:
@davidfestal could you please take a look at this PR? is it really true that external DB depends on the external keycloak?

PR #1252, opened 2020-04-27
Bare metal kubespray quick start

adriendelsalle:
### What does this PR do?
This is a proposition of article about how installing Che on bare metal cluster using Kubespray.

This PR is consecutive to this [issue](https://github.com/eclipse/che/issues/16668), after multiple tries to set-up a on-premise Kubernetes cluster (and not only a test/single-node one) to deploy Che.

Eclipse Che 7.*

### Discussion
I did my best to integrate new pieces of articles in the current *install che locally* structure even if this section is described as follow
> Eclipse Che is available in two modes: * Kubernetes (version 1.9 or higher) or OpenShift cluster (version 3.11 or higher) to deploy Che on * Single-user: Non-authenticated Che, lighter and suited for personal desktop environments * Multi-user: Authenticated Che, suited for the cloud, for organizations and developer teams

It does not cover all *on-premise* clusters, and not my use case: deploying a single-node test/personal cluster or multi-node production grade cluster on bare metal with the *same* recipe.

I have some proposition to make about re-ordering `quick-starts`, let me know if you prefer discussing this topic here or in another PR.

Thanks!


yhontyk (approved):
LGTM, thank you!

PR #1256, opened 2020-04-29
Removing an unnecesary procedure step

MichalMaler:

git cherry-pick b53d1d19e21e92bb5cafbc609b30b9b0673e25b9
Auto-merging src/main/pages/che-7/overview/proc_installing-che-on-openshift-3-using-the-operator.adoc
[cherry ebc12013] Removing an unnecesary procedure step


yhontyk:
@MichalMaler sorry, I didn't review the original PR but I have a couple of questions:
The `oc` from the prerequisites can probably be removed also since it's not used in the procedure, WDYT?
Was it verified by the QE (that we don't need to create a project with `oc`)?

PR #1331, opened 2020-06-24
Change docs according self-signed certificate flag automation

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Updates docs according to automation of `--self-signed-certificate` flag.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/16764
https://github.com/eclipse/che/issues/15301

### Specify the version of the product this PR applies to. 
Eclipse Che 7.15.0

PR #1340, opened 2020-06-28
Fixes to user guide on TLS self-signed certificates

rkratky:
### What does this PR do?

* Fixes/reworks "Installing Che in TLS mode with self-signed certificates"
* Updates @mmorhun's PR: #1331.

@mmorhun, apologies for lumping the changes in one commit -- I changed the structure a bit, simplified the content, rewrote large parts, and fixed many mistakes that were in the doc even before your updates. The diffs would have been unusable. Please, review the resulting doc.

Fixes che/16764.


rkratky:
@mmorhun, one question. I'm struggling to understand the meaning of the following:

_In case when default router certificate is configured with non root CA certificate which is signed by non-cluster root CA certificate, it is required to add the root CA into additional trusted CA certificates config map in order to deploy {prod-short} on such OpenShift cluster._

Also, could you please supply the instructions for _add the root CA into additional trusted CA certificates config map_?

Thank you!

rkratky:
@mmorhun, as you're away, I'm going to merge this now. Let's revisit it later if you find problems with my modifications.

PR #1350, opened 2020-07-01
Content + link mentions

MichalMaler:

* Adding upstream-only content 
* Mentions in various docs using a `link`

### DIFF
--- a/src/main/pages/che-7/administration-guide/con_authentication-mods.adoc
+++ b/src/main/pages/che-7/administration-guide/con_authentication-mods.adoc
@@ -1,0 +1,58 @@
+// authenticating-users
+
+[id="authentication-mods_{context}"]
+= Authentication mods
+
+{prod-short} supports `multiuser` and `singleuser` mode.
+ 
+.Single-user mode
+`singleuser` mode is is a mode that requires no authentication and anyone can access all cluster resources. In `singleuser` mode server performs all operations as the predefined user, regardless of who accesses the server. Therefore, this mode is suitable only for use in a private instance for testing product possibilities and configurations.
+ 
+* H2 database is used. 
+ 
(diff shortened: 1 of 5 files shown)


src/main/pages/che-7/administration-guide/con_authentication-mods.adoc
```diff
@@ -0,0 +1,58 @@
+// authenticating-users
+
+[id="authentication-mods_{context}"]
+= Authentication mods
+
+{prod-short} supports `multiuser` and `singleuser` mode.
+ 
+.Single-user mode
+`singleuser` mode is is a mode that requires no authentication and anyone can access all cluster resources. In `singleuser` mode server performs all operations as the predefined user, regardless of who accesses the server. Therefore, this mode is suitable only for use in a private instance for testing product possibilities and configurations.
```
yhontyk:
```suggestion
`singleuser` mode is a mode that requires no authentication and anyone can access all cluster resources. In `singleuser` mode server performs all operations as the predefined user, regardless of who accesses the server. Therefore, this mode is suitable only for use in a private instance for testing product possibilities and configurations.
```

src/main/pages/che-7/administration-guide/con_authentication-mods.adoc
```diff
@@ -0,0 +1,58 @@
...
+ 
+* H2 database is used. 
+ 
+If {prod-short} is deployed in `singleuser` mode then there is no authentication and anyone who can access the URL of the{prod-short} deployment sees all workspaces and owns all resources. Since the deployment in this mode has fewer containers,  RAM and CPU usage requirements are decreased. This mode is useful if the whole {prod-short} deployment is used by a single person or for lowering resources which are used for {prod-short} itself, but since the server does not authenticate, actions of several users logged in the same workspace can easily interfere with each other.
```
yhontyk:
```suggestion
If {prod-short} is deployed in `singleuser` mode then there is no authentication and anyone who can access the URL of the {prod-short} deployment sees all workspaces and owns all resources. Since the deployment in this mode has fewer containers, RAM and CPU usage requirements are decreased. This mode is useful if the whole {prod-short} deployment is used by a single person or for lowering resources which are used for {prod-short} itself, but since the server does not authenticate, actions of several users logged in the same workspace can easily interfere with each other.
```

yhontyk (approved):

PR #1371, opened 2020-07-07
Introducing the orch-namespace attribute

themr0c:
Introducing the orch-namespace attribute.


src/main/_config.yml
```diff
@@ -53,6 +53,7 @@ asciidoc_attributes: &asciidoc_attributes
   prod-ver-major: 7.x
   prod-cli: chectl
   orch-cli: kubectl
+  orch-namespace:
```
rkratky: Why is it not defined here?

rkratky (approved):

src/main/_config.yml
themr0c: Human error

PR #1374, opened 2020-07-07
Revert back attributes in include statements

themr0c:
Revert back attributes in include statements


tools/substitutions.sed
```diff
@@ -20,6 +20,9 @@ s/\(\(\[id=\|xref\|<<\).*\)\bche\b/\1{prod-id-short}/g
 s/\({prod-id-short}-\|{prod-short} \|{prod} \)6/\1{prod-prev-ver}/g
 s/\({prod-id-short}-\|{prod-short} \|{prod} \)7/\1{prod-ver}/g
 
+# Revert back attributes in include statements
+s/\(include::.*\?\){prod-id-short}/\1che/g
```
rkratky:
```suggestion
/^include::/ s/{prod-id-short}/che/g
```

rkratky (approved):

PR #1377, opened 2020-07-07
Remove mention about self-signed certificates in OperatorHub install

themr0c:
Remove mention about self-signed certificates in OperatorHub installation.


nickboldt:
Seems legit, since 

```
Self signed certificates usage is autodetected now.
```

-- https://github.com/redhat-developer/codeready-workspaces-chectl#crwctl-serverstart

but that's the crwctl + olm method, which is *supposed* to be the same as the operatorhub approach. 

I'm not sure if it in fact is, but @tolusha or @mmorhun can probably confirm

tolusha (approved):

themr0c:
Good news :)

PR #1409, opened 2020-07-16
More consistent use of JWT vs machine tokens

yhontyk:

### What does this PR do?
Introduces a more consistent usage of "JWT tokens" and "machine tokens" terms.

### DIFF
--- a/src/main/pages/che-7/administration-guide/proc_authenticating-to-the-che-server-using-openid.adoc
+++ b/src/main/pages/che-7/administration-guide/proc_authenticating-to-the-che-server-using-openid.adoc
@@ -50,7 +50,7 @@ If the validation is successful, the parsed form of the token is passed to the e
 
 . Environment initialization: The filter extracts data from the JWT token claims, creates the user in the local database if it is not yet present, and constructs the `subject` object and sets it into the per-request *EnvironmentContext* object, which is statically accessible everywhere.
 +
-If the request was made using only a machine token, the following single authentication filter is used:
+If the request was made using only a JWT token, the following single authentication filter is used:
 +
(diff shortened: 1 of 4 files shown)


skabashnyuk:
Can you explain the reasoning of replacement "machine token" with "JWT token"?

yhontyk:
@skabashnyuk hi! The comment left by @metlos 
`The section "6.1.2. Authenticating in... - sometimes uses  "workspace JWT token" and "machine token". These are the same thing IMHO so we should call them the same name consistenly. `

I was thinking either to use only one phrase for consistency. What do you think? Should we keep it as it is?

skabashnyuk:
I agree.
- "machine token"  is a concept that mostly comes from che6 that means a token that is used for authentification against machines aka workspaces
- "JWT" token means the format of the token. https://jwt.io/.  Keycloak token is also jwt token. This form doesn't provide information about where this token is going to be used.

themr0c (approved):

src/main/pages/che-7/administration-guide/proc_authenticating-to-the-che-server-using-openid.adoc
```diff
@@ -50,7 +50,7 @@ If the validation is successful, the parsed form of the token is passed to the e
 
 . Environment initialization: The filter extracts data from the JWT token claims, creates the user in the local database if it is not yet present, and constructs the `subject` object and sets it into the per-request *EnvironmentContext* object, which is statically accessible everywhere.
 +
-If the request was made using only a machine token, the following single authentication filter is used:
+If the request was made using only a JWT token, the following single authentication filter is used:
```
skabashnyuk, line 53: this might be incorrect now. not all JWT tokens would be handled by MachineLoginFilter

PR #1423, opened 2020-07-23
Add keys to _config.yml to support different postgres and keycloak

amisevsk:
### What does this PR do?
Add replacements:
  - `{server-postgresql-image-url}` - image used for postgres
  - `{link-server-postgres-dockerfile-location}` - link to dockerfile for above
  - `{identity-provider-image-url}` - image used for identity provider
  - `{link-server-identity-provider-dockerfile-location}` - link to dockerfile for above

With values set so as to not change existing docs. This allows alternative implementations to be specified in docs.

I chose to abstract out keycloak to "identity-provider" since in CRW e.g. we use RH SSO.

Necessary for https://issues.redhat.com/browse/RHDEVDOCS-1902

All versions, should be no change in output HTML


MichalMaler:
Hello Angel! Nice, thanks for adding those. I will need the CRW equivalents for each of them. Please, hit me on the Slack when you will be online.

amisevsk:
I've renamed the postgres key to match what's currently defined in CRW, for simplicity (otherwise we have to define them to match downstream).

yhontyk (approved):

PR #1424, opened 2020-07-24
Fix procedure to import TLS certificates in Che server Java turststore #RHDEVDOCS-1893

l0rd:
The procedure to import TLS certificates in Che server Java truststore was broken. This PR fix it.

IMPORTANT: currently this section is not published in downstream documentation but that was an error and it should be published in CRW documentation too. I don't think this PR address that but I don't know how this section was filtered out from CRW documentation.


yhontyk (approved):

yhontyk:
Thank you, @l0rd. I'll make sure to sync it with the downstream.

PR #1435, opened 2020-07-30
feat(guide): Add guide on how to test a VS Code extension

benoitf:
### What does this PR do?
Add a guide to try a VS Code extension in Eclipse Che without forking/building plug-in registry

https://github.com/eclipse/che/issues/15982
https://issues.redhat.com/browse/RHDEVDOCS-1637

che 7+


rkratky:
@benoitf, we fixed some lang. (thanks, @boczkowska) and markup issues, and it's ready for merging. Only thing I'm unsure of is the placement in guides. I see you pbbly wanted this in the End-user Guide under **Using developer workspaces**.

I originally thought it was meant for the Admin Guide because under **Customizing devfile and plug-in registries**, there's already a **Using a VS Code extension in Che**. So, I thought the new **Testing a VS Code extension in Che** belonged with it.

For the time being, I'm leaving it in the Admin Guide, but if you think it would be better in the End-user Guide, I suggest we move *both* of the VS Code extension topics there.

rkratky (approved):

benoitf:
Hello @rkratky 
This is really something for end users and not admins 
As here, any user can do that without admin access as it's just a way of providing custom devfiles

benoitf:
But for example, customizing plugin registry or devfile registry is more an admin stuff so I'm unsure we can move everything

PR #1443, opened 2020-08-04
Reworking "making a workspace portable with a devfile"

tomgeorge:

### What does this PR do?

Include orch-upstream-name, prod-short-upstream, prod-upstream, kubernetes-or-openshift attributes for referring explicitly to upstream entities, or when referencing both Kubernetes or OpenShift.

Made revisions to address rhdevdocs-2004

Signed-off-by: Tom George <tgeorge@redhat.com>

rhdevdocs-2004


src/main/pages/che-7/end-user-guide/ref_devfile-reference.adoc
```diff
@@ -182,7 +182,7 @@ Preferences may also be specified as an array:
 
 === Component type: kubernetes
 
-A complex component type that allows to apply configuration from a list of Kubernetes or OpenShift components.
+A complex component type that allows to apply configuration from a list of {kubernetes-or-openshift} components.
```
amisevsk, line 185: Hmm, I'm not convinced that this shouldn't be *only* `Kubernetes components`. IIRC we also have an [`openshift`](https://github.com/eclipse/che/blob/5fe4493988abcdf82d19d6f295b5eaa311890869/wsmaster/che-core-api-workspace/src/main/resources/schema/1.0.0/devfile.json#L566-L570) component type -- I believe `kubernetes` components are deserialized using a k8s client, which won't recognize OpenShift-specific elements. However, I last touched the relevant code years ago, so I'm very ready to be wrong.

tomgeorge: The component type or the prose after?

The `{kubernetes-or-openshift}` attribute evals to “Kubernetes or Openshift” so the prose should say the correct thing.

EDIT: it sounds like maybe we need another component type specifically for openshift?

amisevsk: Yeah, that's what I was suggesting, but that complicates the doc fairly significantly (and I'm not sure if we support any OpenShift-specific things in a openshift component anyways).

yhontyk (approved):

PR #1445, opened 2020-08-05
Troubleshooting revamped

### LINKED ISSUE #17468: Create Network Troubleshooting section that would help users find solution if WebSockets are blocked
### Is your enhancement related to a problem? Please describe.
Some VPN blocks WebSockets and then Che becomes unusable since Che Theia really rely on WebSockets for communication between Client and Server sides.

### Describe the solution you'd like

The suggestions:

* Absolutely put some obvious error detection and informational messaging for the end-user to consume when web sockets is a problem. This will help them from banging their head against a brick wall refreshing their browser. The half-drawn IDE with hanging red C symbol we currently have is not great UX.
  - The dashboard is supposed to show banner after https://github.com/eclipse/che-dashboard/pull/72
  - Do we need Theia to check it on its ide?

* Provide useful suggestions to the end user - now this might be tricky, since “Switch off your VPN” or “Disable local firewall" is never going to be reasonable option in a corporate deployment however suggesting this is a network, firewall, VPN or proxy issue might be a useful clue about who to report this problem to.

* Some suggested links on how to configure VPN, firewalls, proxies etc. These maybe third party, but we should be trying to collect some resolutions stories that we can share as a guide to the network team particularly if they are not familiar with wss. If we can also find some security articles that are positive about websockets they would also be useful to share.

* It would be really useful to run or recommend a 3rd party web site that has wss content that can be used as a test site with no login wall (like https://websocketstest.com/). So if this site works, then Che should be fine. That allows any network engineer to have something to resolve against without needing access to Che.

* Document the network requirement in the Che Installation/Admin guide so there is at least a section highlighting this.

MichalMaler:
### What does this PR do?

* Fixing building errors
* Reworking the Troubleshooting section
* Creating left bar menu for the Troubleshooting section
* Applying the content from Oleksii's PR about Networking issues

### Which issues does this PR reference?
This PR fixes https://github.com/eclipse/che/issues/17468


olexii4 (approved):

MichalMaler:
Waiting on @rkratky fixing PR. Will merge then. Thank you for your cooperation :+1:

sleshchenko:
@MichalMaler @rkratky any news about that PR? We would like to merge it as soon as possible because the good feature (notify users if websockets are blocked) on the dashboard is blocked by this.

rkratky:
@sleshchenko, I'll prioritize this. Thanks for the heads up.

sunix:
http://che.openshift.io/factory?url=https://github.com/MichalMaler/che-docs/tree/Troubleshooting-network-problems

sunix (dismissed):
something is wrong ... 
image

PR #1451, opened 2020-08-07
Reorganize documentation for self-signed certificates

### LINKED ISSUE #17150: Docs still mention manual creation of ssl secrets but that's not needed anymore
### Describe the bug

Section `Installing Che in TLS mode with self-signed certificates` of [the doc](https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates) needs to be reviewed. 

In particular:
- [Deploying Che with self-signed TLS certificates on OpenShift 4 using Operator](https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates/#deploying-che-with-self-signed-tls-on-openshift4-using-operator_installing-che-in-tls-mode-with-self-signed-certificates) can be removed since it's now automated and doesn't require any action from the user.
- [Generating self-signed TLS certificates](https://www.eclipse.org/che/docs/che-7/installing-che-in-tls-mode-with-self-signed-certificates/#generating-self-signed-certificates_installing-che-in-tls-mode-with-self-signed-certificates) should be move out from installation section: it's about usage of a Che workspace, not about installation.

The other sections needs to be reviewed by deploy team.

amisevsk:
### What does this PR do?
Reorganizes documentation around self-signed certs to be more in line with the current functionality of Che

- Manual generation of certs is only needed when deploying to minishift
- Steps for importing self-signed certs are for users, not admins, so this step is included in relevant quickstarts (local deployments)
- Steps for exporting a CA cert from Che are no longer necessary (I think); I've separated it out into its own procedure for clarity, but left that procedure in the repo.

Commits should be broken into logical chunks for review.

Fixes https://github.com/eclipse/che/issues/17150

7.17.x


src/main/pages/che-7/overview/proc_importing-self-signed-tls-certificates-to-browsers.adoc
```diff
@@ -9,30 +13,6 @@ This section describes how to import a root certificate authority into a web bro
 
 When a TLS certificate is not trusted, the error message *Authorization token is missing. Click here to reload page* blocks the login process. To prevent this, add the public part of the self-signed CA certificate into the browser after installing {prod-short}.
```
nickboldt, line 14: should we add a link to `Exporting the self-signed CA certificate from a {prod} deployment` section in case user didn't first export the cert they need to import? 

See https://github.com/eclipse/che-docs/pull/1451/files#diff-e34eb0de4c49c67088afcf20d07cb6faR4 above.

amisevsk: I'm not sure -- the doc on exporting the self-signed cert is kind of limited to admins (get secret or run chectl). This part refers to going through the browser process of trusting a cert (click the cert, save it, and import it into your browser). I think the exporting procedure is no longer necessary in general.

The other side of it is importing the cert into your brower, but I don't know how much we want to document external programs.

I think @tomgeorge was working on something related.

rkratky (approved):
Thanks @amisevsk. I committed a few lang. fixes and merging.

PR #1466, opened 2020-08-17
Some fixes.

yhontyk:

### What does this PR do?
Adding missing contexts and fixing a heading.


rkratky (approved):
Thanks, @boczkowska. I pushed some fixes. Namely, we don't need to set new context for modules (so no parent context either). All this is only needed in assemblies -- or better put, in files that include other modules.

PR #1525, opened 2020-09-03
Remove unused CI files. We don't use ant anymore.

### LINKED ISSUE #17778: In che-docs can we remove the config for the maven build?
To publish che-docs on www.eclipse.org, we do not use these files:

- .ci/*
- make-release.sh
- pom.xml

As far as I know, they were used by ant builds to embed the documentation into the application.

This issue is to double-check that we may remove all these files safely.

themr0c:
Remove unused CI files. We don't use ant anymore. fixes https://github.com/eclipse/che/issues/17778


mkuznyetsov (dismissed):
please do not remove "make-release.sh", as it is currently used for generating tags during releases

themr0c:
I restored the file, thanks @mkuznyetsov !

rkratky (approved):

themr0c:
There is a dependency on pom.xml in ./tools/environment_docs_gen.sh. I have to fix it before merge.
```
Trying to read current product version from pom.xml...
grep: /home/jenkins/agent/workspace/che-docs-pipeline_PR-1525/che-docs/pom.xml: No such file or directory
```

PR #1528, opened 2020-09-03
Reordering chapters in Installation Guide

themr0c:
Reordering chapters in Installation Guide in a more logical order.

### DIFF
--- a/modules/end-user-guide/nav.adoc
+++ b/modules/end-user-guide/nav.adoc
@@ -2,7 +2,7 @@
 
 * xref:navigating-che.adoc[]
 ** xref:navigating-che-using-the-dashboard.adoc[]
-** xref:importing-self-signed-certificates-to-browsers.adoc[]
+** xref:importing-certificates-to-browsers.adoc[]
 ** xref:accessing-che-from-openshift-developer-perspective.adoc[]
 
 * xref:che-theia-ide-basics.adoc[]
--- a/modules/installation-guide/nav.adoc
+++ b/modules/installation-guide/nav.adoc
@@ -2,6 +2,8 @@
         
 * xref:supported-platforms.adoc[]
 
+* xref:configuring-the-che-installation.adoc[]
+
 * xref:installing-che.adoc[]
 
 ** xref:using-the-chectl-management-tool.adoc[]
@@ -25,9 +27,8 @@
 
 ** xref:installing-che-in-a-restricted-environment.adoc[]
 
-* xref:configuring-che.adoc[]
+* xref:advanced-configuration.adoc[]
 
-** xref:configuring-the-che-installation.adoc[]
 ** xref:advanced-configuration-options-for-the-che-server-component.adoc[]
 ** xref:configuring-namespace-strategies.adoc[]
 ** xref:configuring-workspaces-nodeselector.adoc[]
--- a/modules/installation-guide/nav.adoc
+++ b/modules/installation-guide/nav.adoc
@@ -1,16 +1,7 @@
 .Installation Guide
         
 * xref:supported-platforms.adoc[]
-* xref:configuring-che.adoc[]
-** xref:configuring-the-che-installation.adoc[]
-** xref:advanced-configuration-options-for-the-che-server-component.adoc[]
-** xref:configuring-namespace-strategies.adoc[]
-** xref:configuring-workspaces-nodeselector.adoc[]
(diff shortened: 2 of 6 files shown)


rkratky (commented):
I don't agree with this change. The current order makes sense. The Config section covers many things that need to be set before or during the install process. The post-install tuning has its place in the AG.

Keep in mind that while good ordering of topics certainly helps users, the docs are not expected (or meant) to be consumed in a linear fashion. No one reads the docs from start to end. So, it's OK to have topics not exactly in the order that someone envisions for a particular path with the product.

sunix:
Hello Robert,
It makes absolutely no sense to have the configuration before installation. In Che, you configure during or after installation. Try to configure the number of workspaces you can run at the same time... before the installation. You will see ... it is impossible.
However, having links in the installation doc referencing detailed configuration doc makes sense.
 
Though you may not read the doc in a linear fashion ... ordering is important so user could easily find the right information just navigating. We have noticed how hard it was to find the right information when the search box was disabled ... good ordering makes navigation better.

rkratky:
@sunix, the fact that some config is done **during** and some **after** installation supports the current ordering. The user is invited to review what and how can be configured before they start with the installation. It's like providing the right context and foundation for what is to follow.

sunix:
@rkratky as you mention, the doc is not to be read like a book. You cannot invite a user to read all the configuration section before starting to make the installation. It doesn't make sense. Foundation and context are not docs or procedure: these are what we have in the overview sections. (high level architecture, etc.)

sunix:
I read the config part again and I insist that should be after installation. Che has default settings which should usually fit what most of the users and customers are expecting. You can install Che without configuring. That may work for 70% of the users maybe. In the process, configuration comes after to adjust the installation if needed. I wouldn't keep the configuration the very first item of the guide if finally most of the users would skip it or have a look to it afterwards.
Maybe it has not to be read as a book, but if you keep the 9 sections of the configuration at the top of the installation guide, this is what the user would read first when looking for installing Che. And this is not ideal at all, user would just loose time figuring out that it is useless at that point and find the right section or give up.

yhontyk:
I'm for the reorg. Installation guides walk users through the default installation and refer to configuration guides as "a way to do the same thing (install the product) but with non-default values". 
It doesn't make sense to me to list an advanced guide before a basic one.

themr0c:
I guess se should have 2 distinct sections:

```Configuring the Che installation``

before the Installation

and 

```Advanced configuration options``

after the installation.

So a diffferent reorg.

sunix:
The current "Configuring the Che installation" is way to advance to me.
It requires that you have already been through the different ways to install Che:

image

And talking about CRD and all these stuff is really not needed for a basic install.

themr0c:
Can we reach a consensus on this:

Screenshot from 2020-09-04 11-12-28

yhontyk:
From what I understand, in current Configurations we have guides of 2 types:
1. How to install Che with custom properties from scratch (e.g. `Configuring the Che installation` - which, btw, actually not a procedure but just a reference table, or `Configuring workspaces nodeSelector`) 
2. How to configure and update the properties of an *already existing* deployment of Che (e.g. `Importing TLS certificates to Che server Java truststore`).

How about `Advanced installation and configuration options`? I don't see a single guide that actually needs to be logically placed before the installation, it's either "Do the install but with this custom property" or "Update the config of an existing instance".

Correct me if I'm misunderstanding something.

P.s. Ideally, imo, these two types should be separated (so it would be clear what can be configured *only* on install and what can be updated at any time and just applied for new workspaces), but that would require quite a rework of many modules, obv.

sunix:
To me we should keep it simple stupid:
1. installing Che
2. configuring Che
etc ...

if during installation, some parts could be done differently, refer to the right sections in Configuring Che or elsewhere.

yhontyk (approved):
I suggest merging it for now, as with current structure the placement makes sense. Let's brainstorm and review the structure and see if we can come up with something in the next release.

benoitf:
Hi, this new ordering looks more logical to me.

sunix:
If it helps to fix CRW, It is ok to me.
But for simplicity I recommend to use the original proposition: https://github.com/eclipse/che-docs/pull/1528/commits/9c28c5ea0645172c907e5643d54f4be22e5eedfc

sunix (approved):

PR #1580, opened 2020-09-11
Anatoly's feedback

MichalMaler:

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -49,6 +49,7 @@ asciidoc:
     ocp: OpenShift&#160;Container&#160;Platform
     ocp3-ver: 3.11
     ocp4-ver: 4.5
+    operator-image-name: che-operator
     orch-cli: kubectl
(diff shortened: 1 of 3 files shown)


MichalMaler:
@tolusha Anatoly :) Please, check and approve. Thank you!

tolusha (approved):

tolusha (changes requested):
Why do we need these changes for upstream? 
It is about restricted env.

MichalMaler:
well, it is part of modules/installation-guide/partials/proc_upgrading-che-using-the-cli-management-tool.adoc
which is for CHE/CRW (writing this for Che then downstreamed to CRW)

And from the restricted environment chapter in CRW, we are pointing to this file. To make changes in that file, we modify the upstream equivalent. That's why we came out with that new attribute.

MichalMaler:
closing this since there is a better solution

PR #1686, opened 2020-11-04
Update instructions to install ingress-nginx

aledbf:

### What does this PR do?

Update inststructions

An old version of ingress-nginx and syncs intructions

### PR Checklist

### DIFF
--- a/modules/installation-guide/partials/proc_installing-che-on-kind-using-chectl.adoc
+++ b/modules/installation-guide/partials/proc_installing-che-on-kind-using-chectl.adoc
@@ -57,9 +57,8 @@ $ kubectl patch storageclass standard -p '{"metadata": {"annotations": {"storage
 . Install the https://kubernetes.github.io/ingress-nginx/deploy/[NGINX Ingress Controller]:
 +
 ----
-$ VERSION=0.30.0
-$ kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-${VERSION}/deploy/static/mandatory.yaml
-$ kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-${VERSION}/deploy/static/provider/cloud-generic.yaml
+$ VERSION=v0.41.0
+$ kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-${VERSION}/deploy/static/provider/aws/deploy.yaml
 ----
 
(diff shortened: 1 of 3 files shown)


yhontyk (approved):

l0rd (approved):

l0rd:
Thank you @aledbf for your contribution. I couldn't help but notice that you are the creator and maintainer of the ngninx ingress controller and I would love to get your opinion on a couple of issues we recently had:

1. We have tried to use the default TLS certificate of the ngninx ingress on minikube but it [is considered invalid by browsers](https://github.com/eclipse/che/issues/17855#issuecomment-705393366) and hence it's mostly useless. Is that right? Is it a problem that you may consider to fix?
2. To update the default TLS certificate used by the nginx controller on minikube (after enabling the ingress addon) I create a secret and patch the nginx-ingress-controller deployment in the kube-system namespace. Isn't there a way to avoid patching the deployment and configure the default TLS cert before the controller is started?

Those are mostly minikube questions but hopefully you can share some thoughts.

aledbf:
> Thank you @aledbf for your contribution. I couldn't help but notice that you are the creator and maintainer of the ngninx ingress controller and I would love to get your opinion on a couple of issues we recently had:
> 
> 1. We have tried to use the default TLS certificate of the ngninx ingress on minikube but it [is considered invalid by browsers](https://github.com/eclipse/che/issues/17855#issuecomment-705393366) and hence it's mostly useless. Is that right? 

You can change that using the flag --default-ssl-certificate to point to a secret containing an SSL certificate to be used by the default HTTPS server **https://kubernetes.github.io/ingress-nginx/user-guide/cli-arguments/**

> Is it a problem that you may consider to fix?

Not sure this is a problem. We need to expose one SSL certificate. If none is specified we need to create one in order to be able to serve HTTPS connections (without one we can only close the connection).

> 2. To update the default TLS certificate used by the nginx controller on minikube (after enabling the ingress addon) I create a secret and patch the nginx-ingress-controller deployment in the kube-system namespace. Isn't there a way to avoid patching the deployment and configure the default TLS cert before the controller is started?

That is one way to do it. You can also use a static manifest with the required changes like we do for [kind](https://github.com/kubernetes/ingress-nginx/blob/master/deploy/static/provider/kind/deploy.yaml) and publish a link  https://kind.sigs.k8s.io/docs/user/ingress/#ingress-nginx

l0rd:
@aledbf got it, thanks

PR #1741, opened 2020-12-15
Configuring storage strategies

themr0c:
* Move Configuring storage strategies to the Installation Guide.
* Fix xref pointing to this section.


themr0c:
As I am just moving some existing content, it is expected that vale checks fail.

themr0c:
Linkchecker fail is also unrelated. Will look at it in another PR.

yhontyk (approved):
LGTM

themr0c:
Skipping the second review, as this PR  is only moving already existing content.

PR #1749, opened 2020-12-17
Rework running custom registries not to use screenshots

sleshchenko:

### What does this PR do?
Changed article https://5fe07a0d74ff480362a5da1f--eclipse-che-docs-pr.netlify.app/che-7/administration-guide/running-custom-registries/

As suggested in https://github.com/eclipse/che/issues/18518#issuecomment-745144648 this PR reworks in way we don't need screenshots.

Actually, it's pretty easy to update screenshots without redoing the whole flow, with Browser Developer Tools we can just edit DOM manually. So, I can update screenshots as well if it's better.

<details>
<summary>See already done Dashboard next screenshots</summary>

Screenshot_20201217_114500

Screenshot_20201217_114407

</details>

In addition, it also removes duplication we had for Verification steps on k8s and OpenShift.
Screenshot_20201217_121434

:notebook: I did not differentiate kubectl and oc, it can be done pretty easy but I'm not sure we really need it since kubectl is fully compatible with OpenShift and as described in OpenShift Docs, we can use kubectl https://docs.openshift.com/container-platform/3.10/cli_reference/differences_oc_kubectl.html
But it would mean that user need OC to log in on the cluster, and then kubectl to copy/paste instruction. Or manually replace kubectl to oc before execution. So, just tell me and I'll introduce variable to cli depending on the article context

It's done in scope of the https://github.com/eclipse/che/issues/18518

7.23 and even earlier

### PR Checklist

- [x] Changed article references are updated where they are used (or a redirect has been set up on the docs side):
    - [x] Dashboard [branding.json](https://github.com/eclipse/che-dashboard/blob/master/src/components/branding/branding.json)


themr0c (approved):

themr0c:
Great improvement, thanks!

modules/administration-guide/partials/proc_verification-custom-registries.adoc
```diff
@@ -0,0 +1,73 @@
...
+ "name":"my-plug-in",
+ "description":"This plugins shows that we are able to add plugins to the registry...",
+ "publisher":"my-org",
+ "links": \{"self":"/v3/plugins/my-org/my-plug-in/latest" }
+}
+----
+
+. Verify that the {prod-short} server points to the URL of the registry. To do this, compare the value of the `pass:[CHE_WORKSPACE_PLUGIN__REGISTRY__URL]` parameter in the `che` ConfigMap (or `pass:[CHE_WORKSPACE_DEVFILE__REGISTRY__URL]` for the devfile registry):
++
+[subs="+quotes,+attributes"]
```
l0rd:
```suggestion
[subs="+attributes"]
```

or otherwise CHE_WORKSPACE_DEVFILE__REGISTRY__URL is rendered CHE_WORKSPACE_DEVFILE*REGISTRY*URL

modules/administration-guide/partials/proc_verification-custom-registries.adoc
```diff
@@ -0,0 +1,73 @@
...
+. To verify that the new plug-in is correctly published to the plug-in registry, make a request to the registry path `/v3/plugins/index.json` (or `/devfiles/index.json` for the devfile registry).
++
+[subs="+quotes,+attributes"]
+----
+$ URL=$({orch-cli}  get -o 'custom-columns=URL:.spec.rules[0].host' \
+  -l app=che-plugin-registry {platform-routing} --no-headers)
+$ INDEX_JSON=$(curl -sSL http://$\{URL}/v3/plugins/index.json)
+$ echo $\{INDEX_JSON} | grep -A 4 -B 5 "\"name\":\"my-plug-in\""
+,\{
+ "id": "my-org/my-plug-in/1.0.0",
```
l0rd, line 16: This command doesn't work for me:

```
bash-5.0$ echo ${INDEX_JSON} | grep -A 4 -B 5 "\"name\":\"my-plug-in\""
bash-5.0$ ,\{
bash: ,{: command not found
(...)
```

modules/administration-guide/partials/proc_verification-custom-registries.adoc
```diff
@@ -0,0 +1,73 @@
...
+[id="verification-custom_registries-_{context}"]
+
+.Verification steps
+
+. To verify that the new plug-in is correctly published to the plug-in registry, make a request to the registry path `/v3/plugins/index.json` (or `/devfiles/index.json` for the devfile registry).
++
+[subs="+quotes,+attributes"]
+----
+$ URL=$({orch-cli}  get -o 'custom-columns=URL:.spec.rules[0].host' \
+  -l app=che-plugin-registry {platform-routing} --no-headers)
```
l0rd: routes do not have `.spec.rules[0].host` but `.spec.host`

l0rd: and `-l app=che-plugin-registry` doesn't select anything: the route is not labelled. It would be better to use the name, at least the command would work.

modules/administration-guide/partials/proc_deploying-registries-in-openshift.adoc
```diff
@@ -18,7 +18,7 @@ oc new-app -f openshift/che-plugin-registry.yml \
  -n "$\{NAMESPACE}" \
  -p IMAGE="$\{IMAGE_NAME}" \
  -p IMAGE_TAG="$\{IMAGE_TAG}" \
- -p PULL_POLICY="IfNotPresent"
+ -p PULL_POLICY="Always"
```
l0rd, line 21: For helm the pull policy of the registries is still IfNotPresent. Is there a reason for that difference?

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: Hmm,it seems to work fine for me, except that there is space between field name and value. Maybe it's better to use JQ instead. WDYT?

```bash
~/projects/che-docs runningCustomRegistries > INDEX_JSON=$(curl -sSL http://che-plugin-registry.openshift.io/v3/plugins/index.json)                                17:30:44
~/projects/che-docs runningCustomRegistries > echo ${INDEX_JSON} | grep -A 4 -B 5 "\"name\": \"che-theia\""                                                    INT 17:30:46
    "displayName": "theia-ide",
    "id": "eclipse/che-theia/7.21.1",
    "links": {
      "self": "/v3/plugins/eclipse/che-theia/7.21.1/"
    },
    "name": "che-theia",
    "publisher": "eclipse",
    "type": "Che Editor",
    "version": "7.21.1"
  },
--
    "displayName": "theia-ide",
    "id": "eclipse/che-theia/7.21.2",
    "links": {
      "self": "/v3/plugins/eclipse/che-theia/7.21.2/"
    },
    "name": "che-theia",
    "publisher": "eclipse",
    "type": "Che Editor",
    "version": "7.21.2"
  },
--
```

l0rd:
A couple of other typos I have spotted on the doc. Not really related to this PR but if you have the chance please do fix them...

```diff
- (or `app=che-plugin-registry` for the devfile registry)
+ (or `app=che-devfile-registry` for the devfile registry)
```

```diff
- An OpenShift template to deploy the plug-in registry is available in the `openshift/` directory of the GitHub repository.
+ An OpenShift template to deploy the plug-in registry is available in the `deploy/openshift/` directory of the GitHub repository.
```

modules/administration-guide/partials/proc_deploying-registries-in-openshift.adoc
sleshchenko: I believe it may cause issues, typically when I rebuild registry, I use the same tag. And IfNotPresent may lead to situation where my registry is not deployed but image is right. So, I believe always is safer

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: My fault. I'll fix it!

modules/administration-guide/partials/proc_deploying-registries-in-openshift.adoc
sleshchenko: I'll update helm chart to Always as well, if you don't mind

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: @l0rd Could you test

```bash
$ INDEX_JSON=$(curl -sSL http://che-plugin-registry.openshift.io/v3/plugins/index.json)
$ echo ${INDEX_JSON} | yq '.[] | select(.name == "my-plug-in")'                                                     
```

on your end?

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: > and -l app=che-plugin-registry doesn't select anything: the route is not labelled. It would be better to use the name, at least the command would work.

Good catch! The issue is that helm chart and openshift template uses different objects names. 
I'll take a look at the best option how to align them

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: Also, I've noticed that it may be not easy for user to adapt that command to use for devfile registry.
So, I wonder if it's better to distinct articles for devfile and plugin registries.
With templates mechanism we have, we'll have 0 duplicated code, but users will get ready to copy/paste article )

sleshchenko: @themr0c WDYT?

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: @tolusha Do you have any preference here?
Helm charts use `che-plugin-registry` while openshift template and che operator use plugin registry https://github.com/eclipse/che-operator/blob/cc937352745b8202a1f69d99b483b5ae0cf0374e/pkg/deploy/const.go#L4

Personally, I prefer che-plugin-registry a bit more because I would say plugin-registry is too general, but it may be OK if che leaves in a dedicated namespace which should be the case.

modules/administration-guide/partials/proc_deploying-registries-in-openshift.adoc
sleshchenko: Update to Always for helm chart as well. Thanks for catching it.

modules/administration-guide/partials/proc_verification-custom-registries.adoc
l0rd: > @l0rd Could you test
> 
> ```shell
> $ INDEX_JSON=$(curl -sSL http://che-plugin-registry.openshift.io/v3/plugins/index.json)
> $ echo ${INDEX_JSON} | yq '.[] | select(.name == "my-plug-in")'                                                     
> ```
> 
> on your end?

That works

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: Anatoliy brought a good point: During Click Party Deploy team discovered that custom registries docs are not working at all, apart from minor commands mistakes Mario found, there is also Http + Https mixing.
So, I've introduced one more variable that solves che-plugin-registry/plugin registry issue. There is an issue to review and update this article https://issues.redhat.com/browse/RHDEVDOCS-1762

tolusha: @sleshchenko 
I prefer to use labels that are used for operator and helm installer:
`app: che`, `component: plugin-registry`

sleshchenko: The PR on Che side is created https://github.com/eclipse/che/pull/18655
And docs are updated to use labels.

modules/administration-guide/partials/proc_verification-custom-registries.adoc
sleshchenko: did typo. So, changed yq to jq. Because there is two of yq that works fine and the right one is harder to install )

sleshchenko:
After this PR this article is just a bit improved but not ideal and not fully working. Everything else should be fixed in the scope of https://issues.redhat.com/browse/RHDEVDOCS-1762

PR #1762, opened 2020-12-22
Fixes for 7.20.x

themr0c:
Typo: additional t in an id


themr0c:
Wrong target. Misclick.

PR #1795, opened 2021-01-14
Resurrect the che-docs container

themr0c:
* Dockerfile with all tools relevant for che-docs workflow.
* Published on quay.io/eclipse/che-docs. Contribution to automate next publications is welcome.
* Use the image quay.io/eclipse/che-docs as single container for devfile, Jenkinsfile and podmanpreview.sh

Tested on che.openshift.io:

Screenshot from 2021-01-14 16-55-12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN dnf install --refresh -y --nodocs --setopt=install_weak_deps=False --best \
     findutils \
     git-core \
     jq \
-    npm \
+    nodejs \
     python3-jinja2-cli \
     python3-pip \
     python3-setuptools \
@@ -34,7 +34,7 @@ RUN dnf install --refresh -y --nodocs --setopt=install_weak_deps=False --best \
     && pip3 install --no-cache-dir --no-input git+https://github.com/linkchecker/linkchecker.git yq\
     && curl -sfL https://install.goreleaser.com/github.com/ValeLint/vale.sh | sh \
     && mv ./bin/vale /usr/local/bin/vale \
-    && yarnpkg global add --ignore-optional --non-interactive @antora/cli@latest @antora/site-generator-default@latest gulp gulp-cli gulp-connect \
+    && yarnpkg global add --ignore-optional --non-interactive @antora/cli@latest @antora/site-generator-default@latest gulp gulp-connect \
     && rm -rf $(yarnpkg cache dir)/* \
     && rm -rf /tmp/* \
     && antora --version \
@@ -55,5 +55,6 @@ EXPOSE 35729
 VOLUME /projects
 WORKDIR /projects
 ENV HOME /projects
+ENV NODE_PATH /usr/local/share/.config/yarn/global/node_modules
 
 USER 1001
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:latest
+FROM quay.io/fedora/fedora:33-x86_64
 
 LABEL description="Tools to build Eclipse Che documentation: antora, bash, curl, findutils, git, gulp, linkchecker, vale"
 LABEL io.k8s.description="Tools to build Eclipse Che documentation: antora, bash, curl, findutils, git, gulp, jq, linkchecker, newdoc, vale, yq"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,19 @@
-FROM mrksu/newdoc as newdoc
+FROM jdkato/vale:v2.8.0 as vale
 
-FROM jdkato/vale as vale
+FROM rust:1.49.0-alpine3.12 as newdoc-builder
+RUN cargo install newdoc
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM alpine:3.13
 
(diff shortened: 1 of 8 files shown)


themr0c:
Still a draft, need to validate usage in Che.

benoitf:
hello, in the description I don't see the benefits of building a custom image from different tools that have their own images ?

Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM quay.io/fedora/fedora:33-x86_64
```
benoitf: could we use much more smaller images there ?

yhontyk (approved):

Dockerfile
themr0c: Built on purpose on fedora to reduce the delta between container and local build. As we run 1 image instead of 4, I believe benefit is still there.

themr0c:
> hello, in the description I don't see the benefits of building a custom image from different tools that have their own images ?

My desire to use the  official images from the upstream tools is still very high, but I had to accept that:

* Linkchecker: the official image is outdated. I have published a fresh one as a temporary workaround and under my namespace https://quay.io/repository/themr0c/linkchecker. On the long term: Releasing a new version would help to have a refreshed package in fedora and an official image update. Linkchecker project need help to release a new version, and to maintain the package, see https://github.com/linkchecker/linkchecker/issues/359. 

* Antora: the official image doesn't contain gulp. We have used the official container and  add gulp on the spot. Then also as a temporary workaround to because Docker Hub:  https://quay.io/repository/themr0c/antora. 

* Vale: we build an image for the vale extension. It is currently broken, and need to wait before a new version. We need to have some vale binary available *now*. In the future it stays interesting as a fallback.

* Antora container doesn't contain bash, and we have some bash scripts that are difficult to convert to POSIX shell. As a consequence building the docs need multiple containers, and this orchestration can't  be automated in Che. Using a single container will make it easier to streamline the steps: launch the script generating system-variables.adoc, and maybe other scripts as part of the gulp pipeline.

So it means 
* development: maintaining one single image instead of three.
* to start a workspace: pulling and starting one container instead of three. Using 512M of RAM instead of 512 * 3.

In that context I am fine with a larger image :).

Dockerfile
benoitf: For dogfooding I don't still see the benefits ? it looks like more a "I don't use che" use-case ?

benoitf: > Built on purpose on fedora to reduce the delta between container and local build. As we run 1 image instead of 4, I believe benefit is still there.

We're trying to have multi-arch images as well on che, here it seems it's not multi-arch then

themr0c: These are the constraints I saw when choosing a base image:
* Contains glibc (for vale) -> rules out alpine.
* As I understood, the license of ubi8 makes it unsuitable for this project?
* Maximize the number of available system packages:  newdoc is available on fedora
* No system package (on fedora) for linkchecker, yq, antora, asciidoctor.js, gulp, vale.
* To build on Quay without rate-limiting issue: base image should not be published on Docker Hub -> rules out the fedora images published on Docker Hub https://hub.docker.com/_/fedora.
* Fedora is also present on quay: https://quay.io/repository/fedora/fedora?tag=latest&tab=tags: multiarch in only for F34. I opted for stability (F33, but not multiarch). Of course we can discuss this choice. Is multiarch really a blocker?

benoitf: > Contains glibc (for vale) -> rules out alpine.

Official image of vale is using Alpine : https://github.com/errata-ai/vale/blob/v2/Dockerfile

license of ubi8 is fine, but ubi8 is still bigger

themr0c:
Additionally: using latest version of official Antora container led us to unexpected issues after Antora 2.3.4 upgrade made unforeseen changes to the CSS. If we use our own container or set the versions, we will have less surprises like:

* https://github.com/eclipse/che/issues/18660
* https://github.com/eclipse/che-docs/pull/1793

themr0c:
* Switched to alpine as baseimage (approx. 200M instead of 400M).
* Building vale and newdoc to have multiarch binaries.

Dockerfile
themr0c: Switched to alpine as baseimage (approx. 200M instead of 400M).
    Building vale and newdoc to have multiarch binaries.

Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM golang:1.15.6-alpine3.12 as vale-builder
```
benoitf, line 12: Please add copyright headers

benoitf (approved):

Dockerfile
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2018-2020 Red Hat, Inc.
```
themr0c:
```suggestion
# Copyright (c) 2018-2021 Red Hat, Inc.
```

Dockerfile
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2018-2021 Red Hat, Inc.
```
themr0c:
```suggestion
# Copyright (c) 2021 Red Hat, Inc.
```

PR #1804, opened 2021-01-21
Add configuration for htmltest

themr0c:
Add configuration files for https://github.com/wjdp/htmltest link checker.
Not for general usage yet, therefore no references in devfile or Dockerfile.
Just the minimum configuration to enable anyone to study the behavior of the tool in a local environment.
May later replace or complete linkchecker.


benoitf (approved):

tools/podmanhtmltest.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
```
benoitf, line 1: probably need to have copyright headers

tools/podmanhtmltest.sh
```diff
@@ -0,0 +1,7 @@
...
+set -ex
+
+podman run --rm -ti \
```
benoitf: would be nice to have docker as a tool as well
(for example on macos I don't have a working podman)

tools/podmanhtmltest.sh
```diff
@@ -0,0 +1,7 @@
...
+  -v "$(pwd):/test:Z" \
+  "${CHE_DOCS_IMAGE:-wjdp/htmltest}" \
```
benoitf: where is referenced CHE_DOCS_IMAGE ?

themr0c: It is not. It gives the ability to a user to change the image used.
But here it's a leftover of a copy-paste from another script.

tools/podmanhtmltest.sh
themr0c: Done.

tools/podmanhtmltest.sh
themr0c: Fixed

tools/podmanhtmltest.sh
themr0c: Done.

yhontyk (approved):

PR #1814, opened 2021-01-27
Update version to 7.24.2 for 7.24.x branch

themr0c:
Update version to 7.24.2 for 7.24.x branch


themr0c:
Vale error is not legit here. Error on a file that is not touched by the final commit.

PR #1829, opened 2021-02-01
Table regen

MichalMaler:

### PR Checklist


modules/installation-guide/examples/system-variables.adoc
```diff
@@ -226,15 +226,16 @@
...
  `+CHE_WORKSPACE_PROVISION_SECRET_LABELS+`,"`+app.kubernetes.io/part-of=che.eclipse.org,app.kubernetes.io/component=workspace-secret+`","Defines comma-separated list of labels for selecting secrets from a user namespace, which will be mount into workspace containers as a files or env variables. Only secrets that match ALL given labels will be selected." 
- `+CHE_WORKSPACE_DEVFILE_ASYNC_STORAGE_PLUGIN+`,"`+eclipse/che-async-pv-plugin/nightly+`","Plugin is added in case async storage feature will be enabled in workspace config and supported by environment" 
- `+CHE_INFRA_KUBERNETES_ASYNC_STORAGE_IMAGE+`,"`+quay.io/eclipse/che-workspace-data-sync-storage:latest+`","Docker image for the {prod-short} async storage" 
+ `+CHE_WORKSPACE_DEVFILE_ASYNC_STORAGE_PLUGIN+`,"`+eclipse/che-async-pv-plugin/latest+`","Plugin is added in case async storage feature will be enabled in workspace config and supported by environment" 
+ `+CHE_INFRA_KUBERNETES_ASYNC_STORAGE_IMAGE+`,"`+quay.io/eclipse/che-workspace-data-sync-storage:0.0.1+`","Docker image for the {prod-short} async storage"
```
nickboldt, line 233: latest is newer than 0.0.1. Why use old one and have to update it manually? 

https://quay.io/repository/eclipse/che-workspace-data-sync-storage?tag=latest&tab=tags

themr0c: Then we need to update it in the source in: https://github.com/eclipse/che/blob/7.25.x/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties

themr0c:
> Also what is "RedHat Che" ? Is that something like Eclipse Che?

That's exactly the leftover we need to remove.

See:https://github.com/eclipse/che/pull/18938

rkratky:
> What is "RedHat" ? Is that something like "Red Hat" (with space) ?

Yes.

> Also what is "RedHat Che" ? Is that something like Eclipse Che?

No. It's some weird remnant of a past misunderstanding. The purpose of this PR is to get rid of the reference because it was offending people to no end.

themr0c (approved):

modules/installation-guide/examples/system-variables.adoc
themr0c: @nickboldt target is 7.25.x, are you sure the version should be updated in  https://github.com/eclipse/che/blob/7.25.x/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties ?

nickboldt: no idea, I was just pointing out a discrepancy. @skabashnyuk might be able to help you here

nickboldt (approved):
LGTM, but not an expert in specific versions vs. `latest`

PR #1834, opened 2021-02-03
Run build-pr-docs and link-checker in che-docs container

themr0c:
Run build-pr-docs and link-checker in che-docs container


benoitf (approved):

themr0c:
For linkchecker, we could probably do better using a container with a static webserver instead of using gulp.

yhontyk (approved):

PR #1842, opened 2021-02-08
Release fix

mkuznyetsov:

### What does this PR do?
Add various fixes & improvements to the release workflow

### PR Checklist


.github/workflows/release.yml
```diff
@@ -6,16 +6,43 @@ on:
       version:
         description: 'The version that is going to be released. Should be in format 7.y.z'
         required: true
-        default: '7.y.z'
+        default: ''
+      forceRecreateTags:
```
nickboldt, line 10: I thought @themr0c said we would never need to retag or rebranch docs, since docs only live in master branch.

make-release.sh
```diff
@@ -148,15 +148,14 @@ else
   # bump the z digit
   [[ $VERSION =~ ^([0-9]+)\.([0-9]+)\.([0-9]+) ]] && BASE="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"; NEXT="${BASH_REMATCH[3]}"; # for VERSION=7.25.2, get BASE=7.25, NEXT=2 [DO NOT BUMP]
   NEXTVERSION_Z="${BASE}.${NEXT}"
-  bump_version ${NEXTVERSION_Z} ${BASEBRANCH}
+  bump_version ${NEXTVERSION_Z} "master"
```
nickboldt: Should this be hardcoded? We're setting BASEBRANCH on line 113, so why not just use that so it's more obvious what needs refactoring when the project evenutally moves from `master to `main`? https://github.com/eclipse/che-docs/pull/1842/files#diff-be36842270d47fd0f7dd297abbdf5c5ef7916efe78fd4617f5ca6a5fdcba421cR113

make-release.sh
```diff
@@ -29,11 +29,11 @@ Options:
 "
 }
 
+set -e
+
 if [[ ! ${VERSION} ]]; then
   usage
   exit 1
-else # clone into a temp folder so we don't collide with local changes to this script
```
nickboldt, line 35: I added this so that it would be easier to test local changes to this script w/o a checkout failing because of dirty workspace. But I guess for a GH action it's a redundant checkout as the code's already present from an earlier action step.

.github/workflows/release.yml
```diff
@@ -6,16 +6,43 @@ on:
...
+        run: |
+          RECREATE_TAGS=${{ github.event.inputs.forceRecreateTags }}
+          git ls-remote --exit-code origin refs/tags/${{ github.event.inputs.version}}
+          TAG_EXISTS=$?
+          if [[ $TAG_EXISTS -eq 0 ]]; then
+            if [[ $RECREATE_TAGS == "true" ]]; then
+              echo "[INFO] Recreating tags for ${{ github.event.inputs.version}} version"
+              git push origin :${{ github.event.inputs.version}}
+            else
+              echo "[ERROR] Cannot proceed with release - tag already exists"
```
nickboldt, line 33: if the tag already exists... doesn't that mean that we have already released che operator for this $VERSION and therefore it's a GOOD thing that the docs are already done? Therefore this is an INFO or perhaps WARN, but not an ERROR. WDYT ?

Also should this bit of bash code be inside `make-release.sh` so we can run it locally w/o it needing to be run via an action? For this specific case it's probably not urgent, but I'm wondering if best practices here would be to minimize bash code embedded in yaml when we can simply call a bash script w/ params, so that it's easier to run that same script should ther ebe a GH outage.

make-release.sh
nickboldt: maybe it's better to implement this as an option, so we can run locally in /tmp if we want, but otherwise just use current folder's contents

make-release.sh
```diff
@@ -82,9 +83,7 @@ bump_version() {
   git checkout ${BUMP_BRANCH}
 
   echo "[INFO] Update project version to ${NEXTVERSION}"
-  updateYaml ${NEXTVERSION}
-
-  if [[ ${docommit} -eq 1 ]]; then
+  if updateYaml ${NEXTVERSION} && [[ ${docommit} -eq 1 ]]; then
```
nickboldt, line 93: so now none of the commit/push/pull/PR will happen if there's nothing changed in the yaml. Nice improvement!

nickboldt:
Some questions on this one, but overall +1.

nickboldt (approved):

.github/workflows/release.yml
themr0c: Can you clarify what this setting does?

For a `7.__<something>__.0`
release, creating a new `7.__<something>__.x` branch is fine and desired.

For subsequent releases (`7.__<something>__.1`, `7.__<something>__.2`, ...):

* We are actively maintaining the  `7.__<something>__.x` branches corresponding to downstream releases. They include a lot of backportes and fixes, we can't afford any overwrite.

* Recreating / overwriting a `7.__<something>__.x` branch that already exist is a no-go.

* Updating the `antora-playbook.yml` is desired. (Branching from `7.__<something>__.x` + PR to `7.__<something>__.x`)
We (as docs team) are not using tags at all at the moment, but I understand they can make sense for historical or forensics purpose, to retrieve the status of the docs at the day of a release.

mkuznyetsov: it is just to recreate tags, should release have to be rerun several times for some reason, caused by some unexpected error. Bugfix branch will not be recreated

.github/workflows/release.yml
mkuznyetsov: in future, perhaps we could move this to special reusable github action

make-release.sh
mkuznyetsov: sure

themr0c (approved):

themr0c:
We have enough approval to merge. One rebase to put the PR in a mergeable state. Can we merge now @mkuznyetsov?

make-release.sh
mkuznyetsov: basebranch meant something different, as it cane be major or bugfix branch. So I'll rework it, since we have to push in both branches for both release cases

nickboldt (approved):

PR #1854, opened 2021-02-15
fixing podman error

MichalMaler:

### What does this PR do?

When I was running the ./tools/podmanpreview.sh script, the following errors occurred:
* The following tasks did not complete: default, generate
* Did you forget to signal async completion?

@themr0c advised me to add the `async` part for the `function generate` function in the following code:

async function generate (done) {
  generator(antoraArgs, process.env)
    .then(() => done())
    .catch((err) => {
      console.log(err)
      done()
    })
}

Fixing the following errors:

* The following tasks did not complete: default, generate
* Did you forget to signal async completion?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,7 +19,7 @@ const watchPatterns = playbook.content.sources.filter((source) => !source.url.in
 }, [])
 
 
-function generate (done) {
+async function generate (done) {
   generator(antoraArgs, process.env)
     .then(() => done())
     .catch((err) => {


gulpfile.js
```diff
@@ -19,7 +19,7 @@ const watchPatterns = playbook.content.sources.filter((source) => !source.url.in
 }, [])
 
 
-function generate (done) {
+async function generate (done) {
```
benoitf, line 22: not sure how it's fixing the error as it's using callback here

caller is calling generate with a done callback that is invoked when it's finished.
Using async won't solve issue

async would be useful if 
we had for example the code

```javascript
async function generate(): Promise<void> {
  generator(antoraArgs, process.env)
}
```

and then the caller could do
```javascript
await generate();
```

MichalMaler: I am not sure, but it fixed those errors. @themr0c Fabrice, please, can you provide the feedback about what you saw? :)

themr0c: @benoitf This is copy-paste programming. I couldn't reproduce the error experienced by @MichalMaler. Search on the error message proposed many solutions, we tried prepending the function with async and it fixed his problem. It has a nice secondary effect: the web server is starting before the generation of the files is complete. Any further improvement on the script much appreciated, we reached our limits as tech writers here.

PR #1877, opened 2021-03-02
Remove yarn.lock and package.json, docker support for tools.

themr0c:
Ignore and remove unused yarn.lock file to receive less Dependabot errors.

Local installation is not supported or maintained, only usage of the che-docs container is supported and recommended.


benoitf:
if you remove yarn.lock you've to remove package.json file

tools/podmanhtmltest.sh
```diff
@@ -6,23 +6,20 @@
...
-  exit 1
+if command -v podman
+  then RUNNER=podman
+elif command -v docker
+  then RUNNER=docker
+else echo "No installation of podman or docker found in the PATH" ; exit 1
 fi
 
-${RUNNER} run --rm -ti \
+${RUNNER}  run --rm -ti \
```
benoitf:
```suggestion
${RUNNER} run --rm -ti \
```

tools/podmanlinkchecker.sh
```diff
@@ -1,8 +1,22 @@
...
-  che-docs \
-  "./tools/linkchecker.sh" \
+if command -v podman
+  then RUNNER=podman
+elif command -v docker
+  then RUNNER=docker
+else echo "No installation of podman or docker found in the PATH" ; exit 1
+fi
+
+${RUNNER} exec -ti che-docs "./tools/linkchecker.sh"
```
benoitf, line 22: Hello, why not use the official image name there ? `quay.io/eclipse/che-docs:latest` ?
so by default people may not have to build image before

themr0c: It's reusing the image launched by podmanpreview (launched with name che-docs) as it is a prerequisite to have the the site built and running.

themr0c: (So there is no image build involved)

tools/podmanpreview.sh
```diff
@@ -1,7 +1,23 @@
...
 
-podman run --rm -ti \
+if command -v podman
+  then RUNNER=podman
+elif command -v docker
+  then RUNNER=docker
+else echo "No installation of podman or docker found in the PATH" ; exit 1
+fi
+
+${RUNNER}  run --rm -ti \
```
themr0c:
```suggestion
${RUNNER} run --rm -ti \
```

themr0c:
```suggestion
${RUNNER} run --rm -ti \
```

tools/podmanpreview.sh
```diff
@@ -1,7 +1,23 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
 set -ex
 
-podman run --rm -ti \
+if command -v podman
```
benoitf, line 13: it looks like all this commands have almost 95% of copy/paste

would it be possible to include most of the code in a shared file ?

or what about utility where you provide the command you want
./che-docs <linkchecker|preview|>


or even more generic way of running it:
${RUNNER} exec -ti che-docs "./tools/<command>.sh"

where command is like $1 parameter, etc

tools/podmanlinkchecker.sh
benoitf: OK understood about exec vs run but why not having a single command ? like not having to have container running all the time.

benoitf (approved):
removing copy/paste across all .sh files could be done in a follow-up step

yhontyk (approved):

PR #1888, opened 2021-03-05
fix: Use attributes for Kubernetes occurences

themr0c:
Use attributes for Kubernetes occurences

### PR Checklist

- [x] `vale` has been run successfully against the PR branch
 => expected many vale errors due to the number of files touched
- [x] Documentation describes a scenario that is already covered by QE tests, otherwise an issue has been created and acknowledged by Che QE team
- [x] Changed article references are updated where they are used (or a redirect has been set up on the docs side):
    - [x] Dashboard [branding.json](https://github.com/eclipse/che-dashboard/blob/master/src/components/branding/branding.json)


themr0c:
@davidfestal Please help me spot the occurrences that should stay "Kubernetes". In that case we should use the `{kubernetes}` attribute.

modules/installation-guide/examples/checluster-properties.adoc
```diff
@@ -6,7 +6,7 @@
...
 airGapContainerRegistryOrganization: Optional repository name of an alternate container registry to pull images from. This value overrides the container registry organization defined in all the default container images involved in a Che deployment. This is particularly useful to install {prod-short} in a restricted environment.
-allowUserDefinedWorkspaceNamespaces: Defines that a user is allowed to specify a Kubernetes namespace, or an OpenShift project, which differs from the default. It's NOT RECOMMENDED to set to `true` without OpenShift OAuth configured. The OpenShift infrastructure also uses this property.
+allowUserDefinedWorkspaceNamespaces: Defines that a user is allowed to specify a {platforms-namespace}, or an OpenShift project, which differs from the default. It's NOT RECOMMENDED to set to `true` without OpenShift OAuth configured. The OpenShift infrastructure also uses this property.
```
themr0c:
```suggestion
allowUserDefinedWorkspaceNamespaces: Defines that a user is allowed to specify a {platforms-namespace}, which differs from the default. It's NOT RECOMMENDED to set to `true` without OpenShift OAuth configured. The OpenShift infrastructure also uses this property.
```

modules/installation-guide/pages/configuring-labels-for-ingresses.adoc
```diff
@@ -1,6 +1,6 @@
 [id="configuring-labels-for-ingresses"]
 // = Configuring Labels
-:navtitle: Configuring labels for Kubernetes Ingress
+:navtitle: Configuring labels for {orch-name} Ingress
```
themr0c:
```suggestion
:navtitle: Configuring labels for {platforms-ingress}
```

modules/installation-guide/partials/proc_configuring-namespace-strategies.adoc
```diff
@@ -25,7 +25,7 @@ spec:
 
 
 ifeval::["{project-context}" == "che"]
-With **Helm** installer, Kubernetes namespaces strategies are configured using `global.cheWorkspacesNamespace` property.
+With **Helm** installer, {platforms-namespace}s strategies are configured using `global.cheWorkspacesNamespace` property.
```
themr0c:
```suggestion
With **Helm** installer, {platforms-namespace} strategies are configured using `global.cheWorkspacesNamespace` property.
```

modules/administration-guide/pages/viewing-kubernetes-events.adoc
```diff
@@ -1,6 +1,6 @@
 [id="viewing-kubernetes-events"]
-// = Accessing Kubernetes events on OpenShift
-:navtitle: Accessing Kubernetes events on OpenShift
+// = Accessing {kubernetes} events on OpenShift
```
yhontyk, line 2: Could we just call them cluster events?

modules/administration-guide/partials/proc_configuring-bitbucket-server-oauth1.adoc
```diff
@@ -41,7 +41,7 @@ openssl rand -base64 24 > __<bitbucket_server_consumer_key>__
 openssl rand -base64 24 > __<bitbucket_shared_secret>__
 ----
 
-. Create a Kubernetes Secret in {prod-short} namespace containing the consumer and private keys.
+. Create a {orch-name} Secret in {prod-short} namespace containing the consumer and private keys.
```
yhontyk: Will resolve as `a OpenShift`... Can we get away with `your {orch-name} secret`?

modules/end-user-guide/partials/proc_mounting-a-secret-as-an-environment-variable-into-a-workspace-container.adoc
```diff
@@ -13,7 +13,7 @@ The following section describes how to mount a {platforms-name} secret from the
 
 .Procedure
 
-. In the Kubernetes {orch-namespace} where a {prod-short} workspace will be created, generate a new {platforms-name} secret .
+. In the {platforms-namespace} where a {prod-short} workspace will be created, generate a new {platforms-name} secret .
```
yhontyk:
```suggestion
. In the {platforms-namespace} where a {prod-short} workspace will be created, generate a new {platforms-name} secret.
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -10,7 +10,7 @@
...
  `+CHE_WEBSOCKET_ENDPOINT+`,"`+ws://${CHE_HOST}:${CHE_PORT}/api/websocket+`","{prod-short} websocket major endpoint. Provides basic communication endpoint for major websocket interactions and messaging." 
  `+CHE_WORKSPACE_PROJECTS_STORAGE+`,"`+/projects+`","Your projects are synchronized from the {prod-short} server into the machine running each workspace. This is the directory in the machine where your projects are placed." 
- `+CHE_WORKSPACE_PROJECTS_STORAGE_DEFAULT_SIZE+`,"`+1Gi+`","Used when Kubernetes or OpenShift-type components in a devfile request project PVC creation (Applied in case of 'unique' and 'per workspace' PVC strategy. In case of the 'common' PVC strategy, it is rewritten with the value of the `che.infra.kubernetes.pvc.quantity` property.)" 
+ `+CHE_WORKSPACE_PROJECTS_STORAGE_DEFAULT_SIZE+`,"`+1Gi+`","Used when {orch-name} or OpenShift-type components in a devfile request project PVC creation (Applied in case of 'unique' and 'per workspace' PVC strategy. In case of the 'common' PVC strategy, it is rewritten with the value of the `che.infra.kubernetes.pvc.quantity` property.)"
```
yhontyk:
```suggestion
 `+CHE_WORKSPACE_PROJECTS_STORAGE_DEFAULT_SIZE+`,"`+1Gi+`","Used when {kubernetes} or OpenShift-type components in a devfile request project PVC creation (Applied in case of 'unique' and 'per workspace' PVC strategy. In case of the 'common' PVC strategy, it is rewritten with the value of the `che.infra.kubernetes.pvc.quantity` property.)" 
```

 In downstream will resolve as "OpenShift or OpenShift-type"

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -24,13 +24,13 @@
...
- `+CHE_WORKSPACE_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit for each container that has no CPU settings in its environment. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_DEFAULT__MEMORY__REQUEST__MB+`,"`+200+`","RAM request for each container that has no explicit RAM settings in its environment. This amount is allocated when the workspace container is created. This property may not be supported by all infrastructure implementations. Currently it is supported by {orch-name}. A memory request exceeding the memory limit is ignored, and only the limit size is used. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit for each container that has no CPU settings in its environment. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit."
```
yhontyk: I'm not sure it can be referred to as "OpenShift format", it is an inherently k8s thing - https://www.noqcks.io/notes/2016/12/14/kubernetes-understanding-millicores/

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -24,13 +24,13 @@
...
- `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
- `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit."
```
yhontyk: Same as the comment above.

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -24,13 +24,13 @@
...
- `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit."
```
yhontyk: Same as the comment above.

Btw, we'll need to think of a way to automatically apply these attributes to this file. -_-

modules/administration-guide/pages/viewing-kubernetes-events.adoc
themr0c: This navtitle is upstream only. It makes sense to keep Kubernetes here.
(The title is in a comment, we could even remove it)

modules/administration-guide/partials/proc_configuring-bitbucket-server-oauth1.adoc
themr0c: one?

themr0c:
```suggestion
. Create one {orch-name} Secret in {prod-short} namespace containing the consumer and private keys.
```

modules/installation-guide/examples/system-variables.adoc
themr0c: Removed system-variables.adoc  from the scope of this PR.

modules/installation-guide/examples/system-variables.adoc
themr0c: Removed system-variables.adoc  from the scope of this PR.

modules/installation-guide/examples/system-variables.adoc
themr0c: Removed system-variables.adoc  from the scope of this PR.

modules/installation-guide/examples/system-variables.adoc
themr0c: Removed system-variables.adoc  from the scope of this PR.

yhontyk (approved):

modules/installation-guide/examples/checluster-properties.adoc
```diff
@@ -127,7 +127,7 @@ workspacePVCStorageClassName: Storage class for the Persistent Volume Claims ded
 :=== 
  Property: Description 
 ingressClass: Ingress class that will define the which controller will manage ingresses. Defaults to `nginx`. NB\: This drives the `kubernetes.io/ingress.class` annotation on Che-related ingresses.
-ingressDomain: Global ingress domain for a Kubernetes cluster. This MUST be explicitly specified\: there are no defaults.
+ingressDomain: Global ingress domain for a {orch-name} cluster. This MUST be explicitly specified\: there are no defaults.
```
davidfestal: Since this section of the `CheCluster` custom resource is for raw Kubernetes clusters, (non-OpenShift), it might make sense to keep `{kubernetes]` here.

davidfestal: By the way, it seems there's an error at [line 124](https://github.com/eclipse/che-docs/pull/1888/files#diff-eb21f1372e73e4cac48f5299e610e09d815c3d10cc59a21dbb8efdd092b45d1aR124). It should be `Kubernetes` as well afaik. The current docs are misleading here afaict.

modules/installation-guide/examples/system-variables.adoc
davidfestal: I don't really understand why we keep `{orch-name} or OpenShift-type ` here ?
In the case of CRW, will it finally end up as `OpenShift or Openshift-type components...` ? That would look odd, no ?

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -83,39 +83,39 @@
...
- `+CHE_INFRA_KUBERNETES_NAMESPACE_ALLOW__USER__DEFINED+`,"`+false+`","Defines if a user is able to specify Kubernetes namespace (or OpenShift project) different from the default. It's NOT RECOMMENDED to configured true without OAuth configured. This property is also used by the OpenShift infra." 
- `+CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME+`,"`+NULL+`","Defines Kubernetes Service Account name which should be specified to be bound to all workspaces pods. Note that Kubernetes Infrastructure won't create the service account and it should exist. OpenShift infrastructure will check if project is predefined(if `che.infra.openshift.project` is not empty):  - if it is predefined then service account must exist there  - if it is 'NULL' or empty string then infrastructure will create new OpenShift project per workspace    and prepare workspace service account with needed roles there" 
+ `+CHE_INFRA_KUBERNETES_NAMESPACE_ALLOW__USER__DEFINED+`,"`+false+`","Defines if a user is able to specify {platforms-namespace} (or OpenShift project) different from the default. It's NOT RECOMMENDED to configured true without OAuth configured. This property is also used by the OpenShift infra."
```
davidfestal: Do we need to keep `(or OpenShift project)` here ?

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -83,39 +83,39 @@
...
+ `+CHE_INFRA_KUBERNETES_PVC_PRECREATE__SUBPATHS+`,"`+true+`","Defines whether to run a job that creates workspace's subpath directories in persistent volume for the 'common' strategy before launching a workspace. Necessary in some versions of {orch-name} as workspace subpath volume mounts are created with root permissions, and thus cannot be modified by workspaces running as a user (presents an error importing projects into a workspace in {prod-short}). The default is 'true', but should be set to false if the version of {orch-name} creates subdirectories with user permissions. Relevant issue: \https://github.com/kubernetes/kubernetes/issues/41638 Note that this property has effect only if the 'common' PVC strategy used."
```
davidfestal, line 119: > Necessary in some versions of {orch-name} as workspace subpath volume mounts are created with root permissions

Are we sure it is also the case on OpenShift ? cc @skabashnyuk

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -124,18 +124,18 @@
...
- `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
- `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
- `+CHE_INFRA_KUBERNETES_POD_TERMINATION__GRACE__PERIOD__SEC+`,"`+0+`","Defines grace termination period for pods that will be created by Kubernetes / OpenShift infrastructures Grace termination period of Kubernetes / OpenShift workspace's pods defaults '0', which allows to terminate pods almost instantly and significantly decrease the time required for stopping a workspace. Note: if `terminationGracePeriodSeconds` have been explicitly set in Kubernetes / OpenShift recipe it will not be overridden." 
+ `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by {orch-name} Infra This is ignored by OpenShift infra"
```
davidfestal: Here we're presenting a contrast betwen Kubernetes and OpenShift infrastructure (ignored on OpenShift infra). So I assume we should use `kubernetes`

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -124,18 +124,18 @@
...
- `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
- `+CHE_INFRA_KUBERNETES_POD_TERMINATION__GRACE__PERIOD__SEC+`,"`+0+`","Defines grace termination period for pods that will be created by Kubernetes / OpenShift infrastructures Grace termination period of Kubernetes / OpenShift workspace's pods defaults '0', which allows to terminate pods almost instantly and significantly decrease the time required for stopping a workspace. Note: if `terminationGracePeriodSeconds` have been explicitly set in Kubernetes / OpenShift recipe it will not be overridden." 
+ `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by {orch-name} Infra This is ignored by OpenShift infra" 
+ `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by {orch-name} Infra This is ignored by OpenShift infra"
```
davidfestal: Here we're presenting a contrast betwen Kubernetes and OpenShift infrastructure (ignored on OpenShift infra). So I assume we should use `kubernetes`

modules/installation-guide/partials/proc_configuring-workspace-exposure-strategies-using-a-helm-chart.adoc
```diff
@@ -34,9 +34,9 @@ The supported values for `global.serverStrategy` are:
 * xref:default-host-workspace-exposure-strategy_{context}[`default-host`]
 
 
-.Gateway single-host on Kubernetes
+.Gateway single-host on {orch-name}
```
davidfestal: Shouldn't we keep `Kubernetes` here since we're in the context of installing Che from helm charts, which is upstream-only and not targeted at OpenShift installations ?

davidfestal (dismissed):

modules/installation-guide/examples/checluster-properties.adoc
themr0c:
```suggestion
ingressDomain: Global ingress domain for a {kubernetes} cluster. This MUST be explicitly specified\: there are no defaults.
```

modules/installation-guide/partials/proc_configuring-workspace-exposure-strategies-using-a-helm-chart.adoc
themr0c:
```suggestion
.Gateway single-host on {kubernetes}
```

modules/installation-guide/examples/checluster-properties.adoc
themr0c: Do you mean the name of the parameter is `kubernetes` instead of `k8s`?

modules/installation-guide/examples/system-variables.adoc
themr0c:
```suggestion
 `+CHE_WORKSPACE_PROJECTS_STORAGE_DEFAULT_SIZE+`,"`+1Gi+`","Used when {orch-name}-type components in a devfile request project PVC creation (Applied in case of 'unique' and 'per workspace' PVC strategy. In case of the 'common' PVC strategy, it is rewritten with the value of the `che.infra.kubernetes.pvc.quantity` property.)" 
```

modules/installation-guide/examples/system-variables.adoc
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_NAMESPACE_ALLOW__USER__DEFINED+`,"`+false+`","Defines if a user is able to specify {platforms-namespace} different from the default. It's NOT RECOMMENDED to configured true without OAuth configured. This property is also used by the OpenShift infra." 
```

modules/installation-guide/examples/system-variables.adoc
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by {kubernetes} Infra This is ignored by OpenShift infra" 
```

modules/installation-guide/examples/system-variables.adoc
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by {kubernetes} Infra This is ignored by OpenShift infra" 
```

themr0c:
Thanks for the review, @davidfestal. I applied all your suggestions. We still have 2 open questions:
* https://github.com/eclipse/che-docs/pull/1888#discussion_r598691308 for @davidfestal 
* https://github.com/eclipse/che-docs/pull/1888#discussion_r598563101 for @skabashnyuk

modules/installation-guide/examples/system-variables.adoc
skabashnyuk: I believe - yes.

themr0c: OK, thanks.

themr0c:
Vale: 203 errors, 54 warnings and 785 suggestions in 25 files.

Now that fixing Vale errors are mandatory, this PR will never get through approval in one go. We'll have to chunk it.

yhontyk:
> Vale: 203 errors, 54 warnings and 785 suggestions in 25 files.
> 
> Now that fixing Vale errors are mandatory, this PR will never get through approval in one go. We'll have to chunk it.

More like 286 errors, 56 warnings and 946 suggestions in 25 files. now  =\

modules/installation-guide/partials/proc_configuring-namespace-strategies.adoc
```diff
@@ -25,7 +25,7 @@ spec:
 
 
 ifeval::["{project-context}" == "che"]
-With **Helm** installer, Kubernetes namespaces strategies are configured using `global.cheWorkspacesNamespace` property.
+With **Helm** installer, {platforms-namespace} strategies are configured using `global.cheWorkspacesNamespace` property.
```
themr0c:
```suggestion
With **Helm** installer, {orch-namespace} strategies are configured using `global.cheWorkspacesNamespace` property.
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
 ,=== 
 
 [id="kubernetes-infra-parameters"]
-= Kubernetes Infra parameters
+= {orch-name} Infra parameters
```
themr0c:
```suggestion
= Kubernetes Infra parameters
```

themr0c:
```suggestion
= Kubernetes Infra parameters
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
 
-.Kubernetes Infra parameters 
+.{orch-name} Infra parameters
```
themr0c:
```suggestion
.Kubernetes Infra parameters 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
+= {orch-name} Infra parameters
 
-.Kubernetes Infra parameters 
+.{orch-name} Infra parameters 
 ,=== 
  Environment Variable Name,Default value, Description 
  
- `+CHE_INFRA_KUBERNETES_MASTER__URL+`,"","Configuration of Kubernetes client that Infra will use" 
- `+CHE_INFRA_KUBERNETES_TRUST__CERTS+`,"","Configurationof Kubernetes client that Infra will use" 
+ `+CHE_INFRA_KUBERNETES_MASTER__URL+`,"","Configuration of {orch-name} client that Infra will use"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_MASTER__URL+`,"","Configuration of Kubernetes client that Infra will use" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
- `+CHE_INFRA_KUBERNETES_TRUST__CERTS+`,"","Configurationof Kubernetes client that Infra will use" 
+ `+CHE_INFRA_KUBERNETES_MASTER__URL+`,"","Configuration of {orch-name} client that Infra will use" 
+ `+CHE_INFRA_KUBERNETES_TRUST__CERTS+`,"","Configurationof {orch-name} client that Infra will use" 
  `+CHE_INFRA_KUBERNETES_SERVER__STRATEGY+`,"`+multi-host+`","Defines the way how servers are exposed to the world in {orch-name} infra. List of strategies implemented in {prod-short}: default-host, multi-host, single-host" 
- `+CHE_INFRA_KUBERNETES_SINGLEHOST_WORKSPACE_EXPOSURE+`,"`+native+`","Defines the way in which the workspace plugins and editors are exposed in the single-host mode. Supported exposures: - 'native': Exposes servers using {orch-name} Ingresses. Works only on Kubernetes. - 'gateway': Exposes servers using reverse-proxy gateway." 
+ `+CHE_INFRA_KUBERNETES_SINGLEHOST_WORKSPACE_EXPOSURE+`,"`+native+`","Defines the way in which the workspace plugins and editors are exposed in the single-host mode. Supported exposures: - 'native': Exposes servers using {orch-name} Ingresses. Works only on {kubernetes}. - 'gateway': Exposes servers using reverse-proxy gateway."
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_SINGLEHOST_WORKSPACE_EXPOSURE+`,"`+native+`","Defines the way in which the workspace plugins and editors are exposed in the single-host mode. Supported exposures: - 'native': Exposes servers using Kubernetes Ingresses. Works only on {kubernetes}. - 'gateway': Exposes servers using reverse-proxy gateway." 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
+ `+CHE_INFRA_KUBERNETES_NAMESPACE+`,"","DEPRECATED - please do not change the value of this property otherwise the existing workspaces will loose data. Do not set it on new installations. Defines {platforms-namespace} in which all workspaces will be created. If not set, every workspace will be created in a new namespace, where namespace = workspace id It's possible to use <username> and <userid> placeholders (e.g.: che-workspace-<username>). In that case, new namespace will be created for each user. Service account with permission to create new namespace must be used. Ignored for OpenShift infra. Use `che.infra.openshift.project` instead If the namespace pointed to by this property exists, it will be used for all workspaces. If it does not exist, the namespace specified by the che.infra.kubernetes.namespace.default will be created and used."
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_NAMESPACE+`,"","DEPRECATED - please do not change the value of this property otherwise the existing workspaces will loose data. Do not set it on new installations. Defines Kubernetes namespace in which all workspaces will be created. If not set, every workspace will be created in a new namespace, where namespace = workspace id It's possible to use <username> and <userid> placeholders (e.g.: che-workspace-<username>). In that case, new namespace will be created for each user. Service account with permission to create new namespace must be used. Ignored for OpenShift infra. Use `che.infra.openshift.project` instead If the namespace pointed to by this property exists, it will be used for all workspaces. If it does not exist, the namespace specified by the che.infra.kubernetes.namespace.default will be created and used." 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
  `+CHE_INFRA_KUBERNETES_NAMESPACE_CREATION__ALLOWED+`,"`+true+`","Indicates whether {prod-short} server is allowed to create namespaces/projects for user workspaces, or they're intended to be created manually by cluster administrator. This property is also used by the OpenShift infra." 
- `+CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT+`,"`+<username>-che+`","Defines Kubernetes default namespace in which user's workspaces are created if user does not override it. It's possible to use <username>, <userid> and <workspaceid> placeholders (e.g.: che-workspace-<username>). In that case, new namespace will be created for each user (or workspace). Is used by OpenShift infra as well to specify Project" 
+ `+CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT+`,"`+<username>-che+`","Defines {orch-name} default namespace in which user's workspaces are created if user does not override it. It's possible to use <username>, <userid> and <workspaceid> placeholders (e.g.: che-workspace-<username>). In that case, new namespace will be created for each user (or workspace). Is used by OpenShift infra as well to specify Project"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT+`,"`+<username>-che+`","Defines Kubernetes default namespace in which user's workspaces are created if user does not override it. It's possible to use <username>, <userid> and <workspaceid> placeholders (e.g.: che-workspace-<username>). In that case, new namespace will be created for each user (or workspace). Is used by OpenShift infra as well to specify Project" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
+ `+CHE_INFRA_KUBERNETES_NAMESPACE_ALLOW__USER__DEFINED+`,"`+false+`","Defines if a user is able to specify {platforms-namespace} different from the default. It's NOT RECOMMENDED to configured true without OAuth configured. This property is also used by the OpenShift infra." 
+ `+CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME+`,"`+NULL+`","Defines {orch-name} Service Account name which should be specified to be bound to all workspaces pods. Note that {orch-name} Infrastructure won't create the service account and it should exist. OpenShift infrastructure will check if project is predefined(if `che.infra.openshift.project` is not empty):  - if it is predefined then service account must exist there  - if it is 'NULL' or empty string then infrastructure will create new OpenShift project per workspace    and prepare workspace service account with needed roles there"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_SERVICE__ACCOUNT__NAME+`,"`+NULL+`","Defines Kubernetes Service Account name which should be specified to be bound to all workspaces pods. Note that {orch-name} Infrastructure won't create the service account and it should exist. OpenShift infrastructure will check if project is predefined(if `che.infra.openshift.project` is not empty):  - if it is predefined then service account must exist there  - if it is 'NULL' or empty string then infrastructure will create new OpenShift project per workspace    and prepare workspace service account with needed roles there" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -89,39 +89,39 @@ pass:[<!-- vale off -->]
...
+ `+CHE_INFRA_KUBERNETES_PVC_STRATEGY+`,"`+common+`","Defined which strategy will be used while choosing PVC for workspaces. Supported strategies: - 'common'        All workspaces in the same {platforms-namespace} will reuse the same PVC.        Name of PVC may be configured with 'che.infra.kubernetes.pvc.name'.        Existing PVC will be used or new one will be created if it doesn't exist. - 'unique'        Separate PVC for each workspace's volume will be used.        Name of PVC is evaluated as '{che.infra.kubernetes.pvc.name} + '-' + {generated_8_chars}'.        Existing PVC will be used or a new one will be created if it doesn't exist. - 'per-workspace'        Separate PVC for each workspace will be used.        Name of PVC is evaluated as '{che.infra.kubernetes.pvc.name} + '-' + {WORKSPACE_ID}'.        Existing PVC will be used or a new one will be created if it doesn't exist."
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_PVC_STRATEGY+`,"`+common+`","Defined which strategy will be used while choosing PVC for workspaces. Supported strategies: - 'common'        All workspaces in the same Kubernetes namespace will reuse the same PVC.        Name of PVC may be configured with 'che.infra.kubernetes.pvc.name'.        Existing PVC will be used or new one will be created if it doesn't exist. - 'unique'        Separate PVC for each workspace's volume will be used.        Name of PVC is evaluated as '{che.infra.kubernetes.pvc.name} + '-' + {generated_8_chars}'.        Existing PVC will be used or a new one will be created if it doesn't exist. - 'per-workspace'        Separate PVC for each workspace will be used.        Name of PVC is evaluated as '{che.infra.kubernetes.pvc.name} + '-' + {WORKSPACE_ID}'.        Existing PVC will be used or a new one will be created if it doesn't exist." 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -130,18 +130,18 @@ pass:[<!-- vale off -->]
...
- `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
- `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
- `+CHE_INFRA_KUBERNETES_POD_TERMINATION__GRACE__PERIOD__SEC+`,"`+0+`","Defines grace termination period for pods that will be created by Kubernetes / OpenShift infrastructures Grace termination period of Kubernetes / OpenShift workspace's pods defaults '0', which allows to terminate pods almost instantly and significantly decrease the time required for stopping a workspace. Note: if `terminationGracePeriodSeconds` have been explicitly set in Kubernetes / OpenShift recipe it will not be overridden." 
+ `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by {kubernetes} Infra This is ignored by OpenShift infra"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -130,18 +130,18 @@ pass:[<!-- vale off -->]
...
- `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
- `+CHE_INFRA_KUBERNETES_POD_TERMINATION__GRACE__PERIOD__SEC+`,"`+0+`","Defines grace termination period for pods that will be created by Kubernetes / OpenShift infrastructures Grace termination period of Kubernetes / OpenShift workspace's pods defaults '0', which allows to terminate pods almost instantly and significantly decrease the time required for stopping a workspace. Note: if `terminationGracePeriodSeconds` have been explicitly set in Kubernetes / OpenShift recipe it will not be overridden." 
+ `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_RUN__AS__USER+`,"`+NULL+`","Defines security context for pods that will be created by {kubernetes} Infra This is ignored by OpenShift infra" 
+ `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by {kubernetes} Infra This is ignored by OpenShift infra"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_POD_SECURITY__CONTEXT_FS__GROUP+`,"`+NULL+`","Definessecurity context for pods that will be created by Kubernetes Infra This is ignored by OpenShift infra" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -130,18 +130,18 @@ pass:[<!-- vale off -->]
...
- `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_MAX__IDLE+`,"`+5+`","Max number of idle connections in the connection pool of the Kubernetes-client shared http client" 
- `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_KEEP__ALIVE__MIN+`,"`+5+`","Keep-alive timeout of the connection pool of the Kubernetes-client shared http client in minutes" 
+ `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_ASYNC__REQUESTS_MAX__PER__HOST+`,"`+1000+`","Numberof maximum concurrent async web requests (http requests or ongoing  web socket calls) supported in the underlying shared http client of the `{orch-name}Client` instances. Default values are 64, and 5 per-host, which doesn't seem correct for multi-user scenarios knowing that {prod-short} keeps a number of connections opened (e.g. for command or ws-agent logs)"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_ASYNC__REQUESTS_MAX__PER__HOST+`,"`+1000+`","Numberof maximum concurrent async web requests (http requests or ongoing  web socket calls) supported in the underlying shared http client of the `KubernetesClient` instances. Default values are 64, and 5 per-host, which doesn't seem correct for multi-user scenarios knowing that {prod-short} keeps a number of connections opened (e.g. for command or ws-agent logs)" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -130,18 +130,18 @@ pass:[<!-- vale off -->]
...
- `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_MAX__IDLE+`,"`+5+`","Max number of idle connections in the connection pool of the Kubernetes-client shared http client" 
- `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_KEEP__ALIVE__MIN+`,"`+5+`","Keep-alive timeout of the connection pool of the Kubernetes-client shared http client in minutes" 
+ `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_ASYNC__REQUESTS_MAX__PER__HOST+`,"`+1000+`","Numberof maximum concurrent async web requests (http requests or ongoing  web socket calls) supported in the underlying shared http client of the `{orch-name}Client` instances. Default values are 64, and 5 per-host, which doesn't seem correct for multi-user scenarios knowing that {prod-short} keeps a number of connections opened (e.g. for command or ws-agent logs)" 
+ `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_MAX__IDLE+`,"`+5+`","Max number of idle connections in the connection pool of the {orch-name}-client shared http client"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_MAX__IDLE+`,"`+5+`","Max number of idle connections in the connection pool of the Kubernetes-client shared http client" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -130,18 +130,18 @@ pass:[<!-- vale off -->]
...
- `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_MAX__IDLE+`,"`+5+`","Max number of idle connections in the connection pool of the Kubernetes-client shared http client" 
- `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_KEEP__ALIVE__MIN+`,"`+5+`","Keep-alive timeout of the connection pool of the Kubernetes-client shared http client in minutes" 
+ `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_ASYNC__REQUESTS_MAX__PER__HOST+`,"`+1000+`","Numberof maximum concurrent async web requests (http requests or ongoing  web socket calls) supported in the underlying shared http client of the `{orch-name}Client` instances. Default values are 64, and 5 per-host, which doesn't seem correct for multi-user scenarios knowing that {prod-short} keeps a number of connections opened (e.g. for command or ws-agent logs)" 
+ `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_MAX__IDLE+`,"`+5+`","Max number of idle connections in the connection pool of the {orch-name}-client shared http client" 
+ `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_KEEP__ALIVE__MIN+`,"`+5+`","Keep-alive timeout of the connection pool of the {orch-name}-client shared http client in minutes"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_CLIENT_HTTP_CONNECTION__POOL_KEEP__ALIVE__MIN+`,"`+5+`","Keep-alive timeout of the connection pool of the Kubernetes-client shared http client in minutes" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -130,18 +130,18 @@ pass:[<!-- vale off -->]
...
+ `+CHE_INFRA_KUBERNETES_RUNTIMES__CONSISTENCY__CHECK__PERIOD__MIN+`,"`+-1+`","Defines the period with which runtimes consistency checks will be performed. If runtime has inconsistent state then runtime will be stopped automatically. Value must be more than 0 or `-1`, where `-1` means that checks won't be performed at all. It is disabled by default because there is possible {prod-short} Server configuration when {prod-short} Server doesn't have an ability to interact with {orch-name} API when operation is not invoked by user. It DOES work on the following configurations: - workspaces objects are created in the same namespace where {prod-short} Server is located; - cluster-admin service account token is mount to {prod-short} Server pod; It DOES NOT work on the following configurations: - {prod-short} Server communicates with {orch-name} API using token from OAuth provider;"
```
themr0c:
```suggestion
 `+CHE_INFRA_KUBERNETES_RUNTIMES__CONSISTENCY__CHECK__PERIOD__MIN+`,"`+-1+`","Defines the period with which runtimes consistency checks will be performed. If runtime has inconsistent state then runtime will be stopped automatically. Value must be more than 0 or `-1`, where `-1` means that checks won't be performed at all. It is disabled by default because there is possible {prod-short} Server configuration when {prod-short} Server doesn't have an ability to interact with Kubernetes API when operation is not invoked by user. It DOES work on the following configurations: - workspaces objects are created in the same namespace where {prod-short} Server is located; - cluster-admin service account token is mount to {prod-short} Server pod; It DOES NOT work on the following configurations: - {prod-short} Server communicates with Kubernetes API using token from OAuth provider;" 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -26,13 +26,13 @@ pass:[<!-- vale off -->]
...
- `+CHE_WORKSPACE_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit for each container that has no CPU settings in its environment. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_DEFAULT__MEMORY__REQUEST__MB+`,"`+200+`","RAM request for each container that has no explicit RAM settings in its environment. This amount is allocated when the workspace container is created. This property may not be supported by all infrastructure implementations. Currently it is supported by {orch-name}. A memory request exceeding the memory limit is ignored, and only the limit size is used. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit for each container that has no CPU settings in its environment. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit."
```
themr0c:
```suggestion
 `+CHE_WORKSPACE_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit for each container that has no CPU settings in its environment. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
```

themr0c:
```suggestion
 `+CHE_WORKSPACE_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit for each container that has no CPU settings in its environment. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -26,13 +26,13 @@ pass:[<!-- vale off -->]
...
- `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
- `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit."
```
themr0c:
```suggestion
 `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
```

modules/installation-guide/examples/system-variables.adoc
```diff
@@ -26,13 +26,13 @@ pass:[<!-- vale off -->]
...
- `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__LIMIT__CORES+`,"`+-1+`","CPU limit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
+ `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the {orch-name} format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit."
```
themr0c:
```suggestion
 `+CHE_WORKSPACE_SIDECAR_DEFAULT__CPU__REQUEST__CORES+`,"`+-1+`","CPUlimit and request default for each sidecar that has no CPU settings in the {prod-short} plug-in configuration. Specify either in floating point cores number, for example, `0.125`, or using the Kubernetes format, integer millicores, for example, `125m`. Value less or equal to 0 is interpreted as disabling the limit." 
```

themr0c:
Reverted most changes from system-variables.adoc. I believe this PR is ready to merge now. @davidfestal @skabashnyuk can you review?

As a followup: the plan is to revamp completely how we document Che server parameters https://github.com/eclipse/che/issues/19630

skabashnyuk (approved):

PR #1892, opened 2021-03-08
Fix build error: level 0 sections can only be used when doctype is book

themr0c:
Fix build error: level 0 sections can only be used when doctype is book


nickboldt (approved):
+1, assuming this works as expected (no clue)

yhontyk (approved):

yhontyk:
@themr0c doesn't seem like the error's been fixed :(

PR #1919, opened 2021-03-18
Fix id containing a dot

themr0c:
Fix id containing a dot


benoitf (approved):

nickboldt (approved):
smells like a typo fix so +1 from me, dawg

PR #1921, opened 2021-03-18
Fix - use attribute in id

themr0c:
Fix - use attribute in id


nickboldt (approved):
+ :100: -- stop ebedding "che" everywhere when {prod-*} variables exist.

azatsarynnyy (approved):

MichalMaler (approved):

PR #1923, opened 2021-03-18
A new procedure + fixes

MichalMaler:

### What does this PR do?

Documents how to change the plugin registry via preferences

https://github.com/eclipse/che/issues/19237

Current master; crw 2.7

### PR Checklist


yhontyk (approved):

vinokurig:
Actually there is one more way to add a new pluign-registry url: by executing the `Add Registry` command. I think this should be mentioned as well.

yhontyk (approved):

MichalMaler:
> Actually there is one more way to add a new pluign-registry url: by executing the `Add Registry` command. I think this should be mentioned as well.

I mention this in the last sentence:
`This check is also functional for plug-ins added using the Command palette command `Add plugin registry`.`

vinokurig:
@MichalMaler We have two ways of adding a plugin-registry URL
* Add via the command
* Add or edit via the `settings.json` file

The first way is more user friendly, because it uses input messages, but it doesn't allow to edit URLs.
The second way is not so user friendly because it is just a JSON file where user have to find a proper place and set an item in the required format, but it gives an ability to edit the items.
So as the topic name sounds as `Adding a custom plug-in registry ...` I think it would be better to describe the first way as a main way to add a registry URL, and then add that if you want to edit entered item you can use the second way.  WDYT?

MichalMaler:
> @MichalMaler We have two ways of adding a plugin-registry URL
> 
> * Add via the command
> * Add or edit via the `settings.json` file
> 
> The first way is more user friendly, because it uses input messages, but it doesn't allow to edit URLs.
> The second way is not so user friendly because it is just a JSON file where user have to find a proper place and set an item in the required format, but it gives an ability to edit the items.
> So as the topic name sounds as `Adding a custom plug-in registry ...` I think it would be better to describe the first way as a main way to add a registry URL, and then add that if you want to edit entered item you can use the second way. WDYT?

Ok, will redo it tomorrow. I am on sick day today. Will get back to you.

vinokurig:
@MichalMaler Get well soon

vinokurig (approved):

PR #1927, opened 2021-03-25
GitHub Workflows reorganization.

themr0c:
This PR is a mitigation for super slow GitHub Actions runs.

Looking at the behavior of GitHub Actions, what takes the most time is to wait for a runner for each job in each workflow.

The workflow stays as "queued" as long as the last job has not found a runner.

As a consequence, it makes sense to limit the number of *jobs* as much as possible. As a consequence, limit also the number of workflows.

This commit is an attempt to have:
1. build and validate on PR, 
2. publish using netlify after PR build (need a specific workflow because secrets are not available in forks, and PR are run in the security context of forks) 
3. build and validate on push.

It implies some duplication of code, and less visibility for which checks are failing but I guess this is the price to pay for less wait time.

When we find another solution to speedup GitHub actions, we should use steps instead of jobs to make review more readable.


MichalMaler (approved):

benoitf (approved):

benoitf (approved):

benoitf:
you'll need to ping Eclipse webmaster to remove mandatory check 'link checker' as it's being removed

themr0c:
It's not removed, it is just moved in a different file ;).

themr0c:
In action: only 2 workflows scheduled for one PR update:

Screenshot from 2021-03-25 17-29-00

themr0c:
Now that we have workflows with multiple jobs: the situation is even more complex. A workflow that ran 2 jobs successfully appears still as queued. 

So a solution would be to minimize also the number of jobs? But how to keep a sustainable visibility in the PR checks section?

As this pr is now anyway, minimizing the number of workflows is not a solution. 

Screenshot_20210325-193337

PR #1928, opened 2021-03-26
Fix htmltest errors for new pages

themr0c:
Fix htmltest errors for new pages


themr0c:
I apologize, intended content is in https://github.com/eclipse/che-docs/pull/1929

PR #1930, opened 2021-03-26
One job on push. Count vale errors.

themr0c:
* Use also one unique job for the on push workflow.
* Count vale errors, without failing the job.


benoitf (approved):

benoitf:
I think you should add the if failure on steps so you still have them executed

MichalMaler (approved):

PR #1943, opened 2021-04-01
Report breakdown of vale errors per module

themr0c:
Report breakdown of vale errors per module

* Script in the tools directory
* Workflow on push using this script

Current output:

```
./tools/count_vale_errors.sh                                                                                                                                                                             
vale version 2.10.2
# Breakdown of vale errors per module:
modules/administration-guide: 0
modules/contributor-guide: 11
modules/end-user-guide: 0
modules/extensions: 40
modules/glossary: 0
modules/hosted-che: 8
modules/installation-guide: 193
modules/overview: 0
```

### DIFF
--- a/tools/count_vale_errors.sh
+++ b/tools/count_vale_errors.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+vale -v 
+echo "# Breakdown of vale errors per module:"
+
+for module in modules/*
+    do
+    echo -n "$module: "
+    vale --minAlertLevel=error --output=line "$module" | wc -l
+done
--- a/tools/count_vale_errors.sh
+++ b/tools/count_vale_errors.sh
@@ -1,9 +1,19 @@
 #!/bin/sh
+#
+# Copyright (c) 2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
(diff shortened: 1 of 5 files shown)


tools/count_vale_errors.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
benoitf, line 1: you should add license on top of the file

tools/count_vale_errors.sh
```diff
@@ -0,0 +1,9 @@
...
+vale -v 
+echo "# Breakdown of vale errors per module:"
+
+for module in modules/*
+    do
+    echo -n "$module: "
```
benoitf: it doesn't pass shellcheck, so probably use bash instead of sh

benoitf (approved):

tools/count_vale_errors.sh
themr0c: Oh that's strange, shell check was passing locally. I'll fix it after some PTO.

themr0c:
The vale report on the yaml is unexpected. I'll have to fix this also.

boczkowska:
@themr0c we need to disable vale for this file.

tools/count_vale_errors.sh
themr0c: Done.

tools/count_vale_errors.sh
themr0c: Indeed. Last minute change introduced the error. Nicely spotted. Fixed.

yhontyk (approved):

PR #2092, opened 2021-08-09
Create try-pr-on-eclipse-che-hosted-by-red-hat.yaml

ibuziuk:

### What does this PR do?
Create try-pr-on-eclipse-che-hosted-by-red-hat.yaml

Generates a link for starting a workspace on workspaces.openshift.com based on the PR branch in order to try PR on Eclipse Che hosted by Red Hat

image

Related https://github.com/eclipse/che/issues/20155

latest

### PR Checklist

- [x] `vale` has been run successfully against the PR branch
- [ ] Changed article references are updated where they are used (or a redirect has been configured on the docs side):
    - [ ] Dashboard [default branding data](https://github.com/eclipse-che/che-dashboard/blob/main/src/services/bootstrap/branding.constant.ts)


MichalMaler (approved):

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  example_comment_pr:
+    runs-on: ubuntu-latest
+    name: Try the PR on Eclipse Che hosted by Red Hat
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@master
```
benoitf: Hi,
I would not use master as it may break at any time
Also probably we could just add a custom che or redhat action that do the work (so it could be easy to add that action on other repositories)

instead of copying/pasting the same message everywhere (let say we want to update title or badge, etc we would need to update on every repositories)

ibuziuk: +1 for not using master / main - we can switch to `https://github.com/thollander/actions-comment-pull-request/tree/1.0.1`

Could you clarify a bit the custom che / GH action approach? I though that the only way to do it is to create a yaml per repo, similar to what is currently done for the CI builds based on GH actions

benoitf: sure,

for example for happy path tests running with github actions, instead of having each repositories having lot of copy/paste instructions, there is a custom action that can be added

in your example we could replace

```yaml
- name: Comment PR
        uses: thollander/actions-comment-pull-request@1.0.1
        with:
          message: 'Try the PR on Eclipse Che hosted by Red Hat [![Contribute](https://img.shields.io/badge/Eclipse_Che-Hosted%20by%20Red%20Hat-525C86?logo=eclipse-che&labelColor=FDB940)](https://workspaces.openshift.com#https://github.com/${{ github.event.pull_request.head.repo.full_name }}/tree/${{ github.head_ref }})'
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
by

```yaml
- name: Hosted Che by Red Hat
        uses: <che-incubator or redhat-developer>/comment-pr@next
        with:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

(the whole message to add, icons, etc is managed by the action)

so for example we see we should swap icon, text etc we just bump our action and it's updated everywhere
We could add a status check as well

example:
https://github.com/che-incubator/che-deploy-action
https://github.com/che-incubator/happy-path-tests-action


you may check https://github.com/benoitf/che-pr-check-gh-action

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
```diff
@@ -0,0 +1,15 @@
+on: pull_request_target
+
+jobs:
+  example_comment_pr:
+    runs-on: ubuntu-latest
+    name: Try the PR on Eclipse Che hosted by Red Hat
+    steps:
+      - name: Checkout
```
benoitf: do we need to checkout the project ?

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
```diff
@@ -0,0 +1,15 @@
+on: pull_request_target
+
+jobs:
+  example_comment_pr:
```
benoitf: better to cut the 'example'

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-latest
```
benoitf: we should stick to a given ubuntu version

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
ibuziuk: This is nice, but don't you think it is overkill to create a separate action and maintain it on our own vs just having ~ 10 line file per-repo which is unlikely to be changed in the future?

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
ibuziuk: looks like yes according to the usage - https://github.com/ibuziuk/actions-comment-pull-request#usage / examples - https://github.com/thollander/actions-comment-pull-request/pulls

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
benoitf: well it also add status check, you have properties to enable/disable, etc

ibuziuk: well, but this is a significant support burden for a simple badge generation that could be accomplished with an existing action. Not sure if it is really worth reinventing the action for this particular case e.g. - https://github.com/benoitf/che-pr-check-gh-action

ibuziuk:
build/validation failures are related to Service degradation in all web properties  https://status.redhat.com/

.github/workflows/try-pr-on-eclipse-che-hosted-by-red-hat.yaml
ibuziuk: Indeed, looks like the checkout step is not needed

ibuziuk: Removed

benoitf (approved):
approving as step 0, but then we'll probably need a dedicated action

PR #2111, opened 2021-09-06
fix: example value of `CHE_WORKSPACE_POD_TOLERATIONS__JSON` to double-quote

int128:
I set a single-quoted JSON (as the example in doc) to `CHE_WORKSPACE_POD_TOLERATIONS__JSON` but got the following error:

```
Caused by: com.fasterxml.jackson.core.JsonParseException: Unexpected character (''' (code 39)): was expecting double-quote to start field name
```

It seems the value of `CHE_WORKSPACE_POD_TOLERATIONS__JSON` must be a double-quoted JSON. I'd like to fix the example value in doc.

None.

I'm using Eclipse Che 7.34.0.

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


themr0c:
Thank you for the contribution.

As this file is single-sourced from upstream sources, the changes will be overwritten after the next run of the scipt. 

The fix need to happen in the upstream source files:

*  https://github.com/eclipse-che/che-server/blob/main/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
* https://github.com/eclipse-che/che-server/blob/main/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
 
@int128  can you provide the same patch as a pull request to the che-server repository?

int128:
@themr0c Thank you for your explanation.

I found the example at https://github.com/eclipse-che/che-server/blob/6d4ee34fe8eefcbe686cdd1e400779ea19f7c60d/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L754-L758 but it is already double-quoted JSON as follows:

```properties
# Example: `[{"effect":"NoExecute","key":"aNodeTaint","operator":"Equal","value":"aValue"}]`
che.workspace.pod.tolerations_json=NULL
```

Also I found that the script replaces `"` with `'` as follows:
https://github.com/eclipse/che-docs/blob/e3d0a0eca9a45f2f0c736de476f24f8508f53ace/tools/environment_docs_gen.sh#L99

I think the replacement is needed for [CSV table syntax in Asciidoctor](https://docs.asciidoctor.org/asciidoc/latest/tables/data-format/). It would be possible to use `"` in [Basic table syntax](https://docs.asciidoctor.org/asciidoc/latest/tables/build-a-basic-table/).

themr0c:
@int128 If you manage to find a solution to update the script, that would be awesome!

PR #2130, opened 2021-10-07
fix: Jenkinsfile

themr0c:
Fix daily build on Eclipse Jenkins.
Refactor Jenkinsfile.


Jenkinsfile
```diff
@@ -88,99 +91,50 @@ spec:
...
-                sh '''
-                git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
-                git fetch
-                ./tools/publication.sh
-                '''
-            }
+            sh './tools/build-for-publication.sh'
+          }
         }
+        milestone 22
```
nickboldt, line 110: what does this do?

nickboldt (approved):
seems ok but didn't test/run anything.

MichalMaler (approved):

Jenkinsfile
themr0c: It's a fail fast feature in Jenkins: if a later build is running on the same branch, and reached a higher milestone, then the current build stops directly at milestone check. 

See: https://www.jenkins.io/doc/pipeline/steps/pipeline-milestone-step/#pipeline-milestone-step

PR #2145, opened 2021-10-27
fix: Fix Che Variables update script

mkuznyetsov:
Signed-off-by: Mykhailo Kuznietsov <mkuznets@redhat.com>

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/.github/workflows/che-variables-update.yml
+++ b/.github/workflows/che-variables-update.yml
@@ -25,7 +25,7 @@ jobs:
           git checkout ${BASE_BRANCH}
           ./tools/environment_docs_gen.sh
           
-          if git status --porcelain; then
+          if git diff --quiet; then
             PR_BRANCH="pr-update-base-images-$(date +%s)"
             git checkout -b ${PR_BRANCH}
             git add -A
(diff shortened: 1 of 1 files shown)


themr0c (approved):

nickboldt (changes requested):
Doesn't do what you think it does:

```
$➔ git diff
diff --git a/LICENSE b/LICENSE
index d3087e4c..a3ed1807 100644
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,5 @@
-Eclipse Public License - v 2.0
+  Eclipse Public License - v 2.0
 
-    THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE
     PUBLIC LICENSE ("AGREEMENT"). ANY USE, REPRODUCTION OR DISTRIBUTION
     OF THE PROGRAM CONSTITUTES RECIPIENT'S ACCEPTANCE OF THIS AGREEMENT.
```
So this should tell me a file has changed:
``` 
$➔ if git diff --quiet; then echo "changed"; else echo "not changed"; fi
not changed
```

But this works:

```
$➔ if [[ $(git diff --quiet; echo $?) -gt 0 ]]; then echo "changed"; else echo "not changed"; fi
changed
```

As does this:

```
$➔ if git diff --name-only; then echo "changed"; else echo "not changed"; fi
LICENSE
changed
```

nickboldt (approved):

PR #2146, opened 2021-10-27
docs: storage strategies concept update

themr0c:

* PVC strategies table should be a regular content
* PVC strategies concept update, required as we have now only one unique namespace strategy: `per-user`.

https://issues.redhat.com/browse/CRW-2400

## Pull Request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:
    - [x] Dashboard [default branding data](https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/services/bootstrap/branding.constant.ts)


modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -7,7 +7,30 @@
...
+[width="100%",cols="^,^,^,^",options="header"]
+|===
+|Strategy |Details |Pros |Cons
+|*common (default)* |
+One PVC for all workspaces in one {platforms-namespace}
+
+
+| Easy to manage and control storage | If PV does not support `ReadWriteMany` (RWX) access mode, the configuration will work only if one or both of the following apply:
+
+* Each workspace is in a separate {platforms-namespace}
```
ibuziuk: could you clarify what exactly is meant here? any details on how is it possible to configure running each workspace in a separate namespace?
Why a user would opt for the common PVC if each workspace is isolated in a separate namespace? The whole essence of the common PVC is sharing a single persistent volume across all the user's workspaces

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -7,7 +7,30 @@
 
 Workspace Pods use Persistent Volume Claims (PVCs), which are bound to the physical Persistent Volumes (PVs) with link:https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes[ReadWriteOnce access mode]. It is possible to configure how the {prod-short} server uses PVCs for workspaces. The individual methods for this configuration are called PVC strategies:
 
-include::example${project-context}-pvc-strategies.adoc[]
+
+[width="100%",cols="^,^,^,^",options="header"]
+|===
+|Strategy |Details |Pros |Cons
+|*common (default)* |
+One PVC for all workspaces in one {platforms-namespace}
```
ibuziuk:
```suggestion
One PVC is shared across all workspaces of a given user
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -7,7 +7,30 @@
...
+
+
+| Easy to manage and control storage | If PV does not support `ReadWriteMany` (RWX) access mode, the configuration will work only if one or both of the following apply:
+
+* Each workspace is in a separate {platforms-namespace}
+
+* Only one workspace is running per {platforms-namespace} at the same time. See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
```
ibuziuk: this statement is missing details that were part of the https://github.com/eclipse/che-docs/pull/1902

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: You mean: remove this line, as it is meaningless?

themr0c: Can it be some leftover from the time when multiple namespace strategies existed?

ibuziuk: I do not know what is meant, so I'm asking the author of PR ;-)
for me the statement does not make sense

themr0c: Indeed: All namespace strategies other than per-user are deprecated for 7.30 and removed for 7.32. We removed direct mentions of these strategies, but here we can simplify the docs as we can assert that one user == one namespace.

themr0c: @ibuziuk I added substantial refactoring, does it address your concern?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: @ibuziuk I added substantial refactoring, does it address your concern?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,113 @@
...
+
+| Control storage
+| easy
+| complex 
+| most complex
+
+| User count
+| < 10
+| > 100
+| < 100
```
ibuziuk: where have those figures come from? Look pretty ambiguous for me

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,113 @@
...
+
+| Cluster size
+| single node cluster
+| multinode cluster
+| multinode cluster
```
ibuziuk: where have you got that info? 
common PVC can be also used on the multi-node clusters

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,113 @@
...
+
+| Limitations
+| Maximum one running workspace per user when the PV is in `ReadWriteOnce` access mode
+| Somewhat unpredictable PV count
```
ibuziuk: why is it unpredictable? number of PVs = number of workspaces

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: The previous content is even more ambiguous: "great number of users", "lesser amount of users". Let's find something measurable. I will ask around for real numbers.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,113 @@
...
 
-When the `common` strategy is used, and a workspace PVC access mode is ReadWriteOnce (RWO), only one node can simultaneously use the PVC.
-
-If there are several nodes, you can use the `common` strategy, but:
+`ReadWriteOnce` access mode::
+The `ReadWriteOnce` access mode limits each user to run one unique concurrent workspace. See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
 
-* The workspace PVC access mode must be reconfigured to `ReadWriteMany` (RWM), so multiple nodes can use this PVC simultaneously.
+Scalability::
+Consider using this strategy in single-node clusters.
```
ibuziuk: should be removed

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,113 @@
...
+The `common` PVC strategy is not suitable for large multinode clusters.
```
ibuziuk:
```suggestion
Scalability::
The `common` PVC strategy is not suitable for multinode clusters with 'ReadWriteOnce' access mode for PVs if the number of concurrently running workspaces per user is more than 1
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: Is it now advised? Current docs say it is not recommended.

Or should we change for "Cluster size recommendation" to make it more explicit?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: Current docs say: "PV count is not known and depends on the number of workspaces". Is the number of workspaces predictable?  It's a real question, how do you do capacity planning for workspaces.openshift.com?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: Can we now recommend the `common` strategy for large clusters and "many" users?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: Current docs says this (but the `per-workspace` namespace strategy has been deprecated and removed, so it can't stay as it is): 

> The `common` PVC strategy is not suitable for large multi-node clusters. Therefore, it is best to use it in single-node clusters. However, in combination with the `per-workspace` {orch-namespace} strategy, the `common` PVC strategy is usable for clusters with not more than 75 nodes. The PVC used with this strategy must be large enough to accommodate all projects to prevent a situation in which one project depletes the resources of others.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -71,7 +122,12 @@ User-defined PVCs are created with the following specifics:
 
 * Subpaths of the mounted Physical persistent volumes that reference user-defined PVCs are prefixed with `_<workspace-ID>_` or `__<PVC-name>__`. This ensures that the same PV data structure is configured with different PVC strategies. For details, see xref:how-subpaths-are-used-in-pvcs_{context}[].
 
-The `unique` PVC strategy is suitable for larger multi-node clusters with a lesser amount of users. Since this strategy operates with separate PVCs for each volume in a workspace, vastly more PVCs are created.
+
+Scalability::
+Consider using this strategy on larger multinode clusters with a lesser amount of users.
```
ibuziuk: should be removed IMO

ibuziuk (commented):
Looks good in general
Thanks for adding back the missing info from https://github.com/eclipse/che-docs/pull/1902

A couple of statements that need to be clarified:

1. User count. I believe there is no strict correlation between the number of users and the PVC strategy. However, there is a dependency between the strategy and number of PVc required per user:
- common -> one per user
- per-workspace -> one per workspace
- unique -> one per volume mount defined in a devfile

 2. Cluster size. All PVC strategies can work on both single and multi-node clusters

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: @ibuziuk or can we entirely remove these user count restrictions?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c, line 126:
```suggestion
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -71,7 +121,12 @@ User-defined PVCs are created with the following specifics:
...
```
themr0c:
```suggestion
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
ibuziuk: I would suggest removing the user-count section at all. @l0rd wdyt?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
ibuziuk: I would remove cluster size section also - all strategises can work on single and multinode clusters

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
ibuziuk: exactly, so let's change `Somewhat unpredictable PV count` to `PV count depends on the number of workspaces`

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
l0rd: I agree. Let's remove this line.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,112 @@
...
+|===
+| Storage strategy name
+| `common`
+| `per-workspace`
+| `unique`
+
+| PVC count
+| One per user
+| One per non-ephemeral workspace
+| One per workspace volume and user-defined PVC
```
l0rd: I don't understand what this sentence means. Isn't `unique` just "One"? What's the "workspace volume" what's the "user-defined PVC"?

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
l0rd: I agree with @ibuziuk, please remove it.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,112 @@
...
-include::example${project-context}-pvc-strategies.adoc[]
+.Storage strategies comparison
+[width="100%",cols="1,1,1,1",options="header"]
+|===
+| Storage strategy name
+| `common`
+| `per-workspace`
+| `unique`
+
+| PVC count
```
l0rd:
```suggestion
| PV count
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
ibuziuk: yes, what is more, there is a discussion around having only 'common' supported strategy in the future.
For example Developer Sandbox / workspaces.openshift.com is using common PVC strategy

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,112 @@
...
+| One per user
+| One per non-ephemeral workspace
```
l0rd: Do we really need to mention "non-ephemaral". A sentence that is easy to read ("One per workspace") becomes hard ("what the heck is a non-ephemeral workspace"?). 

Moreover this is a guide for admins. For admins that need to do capacity planning. And since they cannot control if developers will use ephemeral or not, they should consider the case that costs more (non-ephemeral). Thus we should remove "non-ephemeral" here.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
l0rd: A PVC is the request for storage, a PV is a chunk of storage. PV count and PVC should be the same. But as an admin I probably want to know how many PVs I have to provision rather than the number of PVC that Che is going to issue. That's not critical but it may be worth reviewing this article usage of PVC vs PV.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
l0rd: Ok I got it now. I think that "Multiple PVs per workspace" would be simpler here.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,62 +5,112 @@
...
+| One per workspace volume and user-defined PVC
+
+| Default
+| yes
+| no
+| no
+
+| Storage isolation
+| no
```
l0rd: I think that's misleading. "storage not isolated"  may be interpreted as "users have access to others storage" which is not true and may be a reason for customers to not adopt Che. I am not sure what we wanted to say here but I would remove this line.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,131 @@
...
+| Default
+| yes
+| no
+| no
 
-All workspaces inside a {platforms-namespace} use the same Persistent Volume Claim (PVC) as the default data storage when storing data such as the following in their declared volumes:
+| Control storage
+| easy
+| complex 
+| most complex
```
ibuziuk: I would probably remove this section because it is more ambiguous then infortative

themr0c:
```suggestion
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,127 @@
 [id="storage-strategies-for-{prod-id-short}-workspaces_{context}"]
 = Storage strategies for {prod-id-short} workspaces
 
-Workspace Pods use Persistent Volume Claims (PVCs), which are bound to the physical Persistent Volumes (PVs) with link:https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes[ReadWriteOnce access mode]. It is possible to configure how the {prod-short} server uses PVCs for workspaces. The individual methods for this configuration are called PVC strategies:
+Storage strategy::
+A configurable method defining how {prod-short} workspaces uses persistent volume claims (PVCs) and persistent volumes (PVs).
```
ibuziuk:
```suggestion
A configurable method defining how {prod-short} workspaces uses persistent volume claims (PVCs) which are bound to the physical persistent volumes (PVs).
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,127 @@
...
 
-If there are several nodes, you can use the `common` strategy, but:
+When the user deletes a workspace, the workspace engine deletes the corresponding subdirectory (`$\{ws-id}`) in the persistent volume directory.
 
-* The workspace PVC access mode must be reconfigured to `ReadWriteMany` (RWM), so multiple nodes can use this PVC simultaneously.
+When the user deletes the last workspace, the workspace engine removes the common persistent volume. 
 
-* Only one workspace in the same {orch-namespace} may be running. See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
+Restrictions on the `common` storage strategy with `ReadWriteOnce` access mode::
+The `ReadWriteOnce` access mode limits each user to run one unique concurrent workspace.
```
ibuziuk:
```suggestion
The `ReadWriteOnce` access mode limits each user to run only one concurrent workspace.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,127 @@
...
-
+When a user defines volumes for components in the devfile, all components that define the volume of the same name will be backed by the same directory in the persistent volume as `__<persistent-volume-name>__`, `__<workspace-ID>__`, or `__<original-volume-name>__`.
+Each component can have this location mounted on a different path in its containers.
 
 .Example
-Using the `common` PVC strategy, user-defined PVCs are replaced with subpaths on the common PVC. When the user references a volume as `my-volume`, it is mounted in the common-pvc with the `/workspace-id/my-volume` subpath.
+With the `common` storage strategy, the workspace engine replaces user-defined volumes with subpaths on the common persistent volume.
+When the user references the `__<my-volume>__` volume, the workspace engine mounts it on the common persistent volume with the `/__<workspace-id>__/__<my-volume>__` subpath.
```
ibuziuk: this section should be part of the 'common' strategy imo

ibuziuk (commented):
looks good in general, left a couple of suggestions

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: I think it does not make much sense to the user to introduce details on the relationship between PV and PVC.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: The common strategy already has it, let's remove it.

themr0c:
(fun fact: most of this concept file will go away with devworkspaces, as we will document only the `common` storage strategy)

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+| no
+| no
 
-All workspaces inside a {platforms-namespace} use the same Persistent Volume Claim (PVC) as the default data storage when storing data such as the following in their declared volumes:
 
-* projects
-* workspace logs
-* additional Volumes defined by a use
+| Limitations
+| Maximum one running workspace per user when the persistent volume is in `ReadWriteOnce` access mode
```
l0rd:
```suggestion
| Set a limit of one running workspace per user to avoid issues when mounting volumes
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+For each user, all workspaces use the same Persistent Volume (PV) for the default data storage.
 
-.Restrictions on using the `common` PVC strategy
+When the user starts a first non-ephemeral workspace, the workspace engine creates the common persistent volume.
 
-When the `common` strategy is used, and a workspace PVC access mode is ReadWriteOnce (RWO), only one node can simultaneously use the PVC.
+When the user starts a non-ephemeral workspace, the workspace engine uses the common persistent volume.
+The workspace only binds simultaneously to one node in the {orch-name} cluster.
+The workspace engine ignores user-defined volumes. 
+The workspace engine replaces volumes related to user-defined volumes with a subpath in the common volume.
```
l0rd: The 2 rows above are contradictory. First we say that user-defined volumes are ignored. Then that they are "replaced" with a subpath in the common volume.

The "common" PVC strategy is implemented using [subPaths](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath) of a "common to all workspaces" PV. That's not specific of user defined volumes but applies to any file that needs to be persisted through the restarts of a workspace (projects included).

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: Avoid "recommends". See:  https://redhat-documentation.github.io/supplementary-style-guide/#recommend

l0rd: 👍

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
Srivaralakshmi:
```suggestion
A configurable method defining how {prod-short} workspaces use persistent volume claims (PVCs) and persistent volumes (PVs). 
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
 [id="storage-strategies-for-{prod-id-short}-workspaces_{context}"]
 = Storage strategies for {prod-id-short} workspaces
 
-Workspace Pods use Persistent Volume Claims (PVCs), which are bound to the physical Persistent Volumes (PVs) with link:https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes[ReadWriteOnce access mode]. It is possible to configure how the {prod-short} server uses PVCs for workspaces. The individual methods for this configuration are called PVC strategies:
+Storage strategy::
+A configurable method defining how {prod-short} workspaces uses persistent volume claims (PVCs) and persistent volumes (PVs). 
+It defines the storage for workspace data: projects, workspace logs, or additional Volumes defined by a user.
```
Srivaralakshmi:
```suggestion
This method defines the storage for workspace data, for example, projects, workspace logs, or additional volumes defined by a user.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
Srivaralakshmi:
```suggestion
| Maximum one running workspace per user when the PV is in the `ReadWriteOnce` access mode
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
-The {prod-short} Volume name is identical to the name of the user-defined PVC. It means that if a machine is configured to use a {prod-short} volume with the same name as the user-defined
-PVC has, they will use the same shared folder in the common PVC.
 
-When a workspace is deleted, a corresponding subdirectory (`$\{ws-id}`) is deleted in the PV directory.
+[id="the-common-storage-strategy_{context}"]
+== The `common` storage strategy
 
-NOTE: The common PVC will be removed when a user's workspaces are all deleted. The PVC will be re-created when a non-ephemeral workspace is started.
+This is the default storage strategy.
+For each user, all workspaces use the same Persistent Volume (PV) for the default data storage.
```
Srivaralakshmi:
```suggestion
For each user, all workspaces use the same PV for the default data storage.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-.Restrictions on using the `common` PVC strategy
+When the user starts a first non-ephemeral workspace, the workspace engine creates the common persistent volume.
```
Srivaralakshmi:
```suggestion
When the user starts a first non-ephemeral workspace, the workspace engine creates a common PV.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-When the `common` strategy is used, and a workspace PVC access mode is ReadWriteOnce (RWO), only one node can simultaneously use the PVC.
+When the user starts a non-ephemeral workspace, the workspace engine uses the common persistent volume.
```
Srivaralakshmi:
```suggestion
When the user starts another non-ephemeral workspace, the workspace engine uses the same common PV.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
l0rd: I have updated the suggestion

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+.Storage strategies comparison
+[width="100%",cols="1,1,1,1",options="header"]
+|===
+| Storage strategy name
+| `common`
+| `per-workspace`
+| `unique`
 
-{prod-short} uses the `common` PVC strategy in combination with the "one {orch-namespace} per user" {orch-namespace} strategy when all {prod-short} workspaces operate in the user's {orch-namespace}, sharing one PVC.
+| Persistent volume count
```
Srivaralakshmi:
```suggestion
| PVC
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+| One per user
+| One per workspace
+| Multiple persistent volumes per workspace
```
Srivaralakshmi:
```suggestion
| Multiple PVs per workspace
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+| no
 
-All workspaces inside a {platforms-namespace} use the same Persistent Volume Claim (PVC) as the default data storage when storing data such as the following in their declared volumes:
 
-* projects
-* workspace logs
-* additional Volumes defined by a use
+| Limitations
+| Maximum one running workspace per user when the persistent volume is in `ReadWriteOnce` access mode
+| Persistent volume count depends on the number of workspaces
```
Srivaralakshmi:
```suggestion
| PVC depends on the number of workspaces
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+| Unpredictable persistent volume count
```
Srivaralakshmi:
```suggestion
| Unpredictable PVC
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-When the `common` PVC strategy is in use, user-defined PVCs are ignored, and volumes related to these user-defined PVCs are replaced with a volume that refers to the common PVC.
-In this strategy, all {prod-short} workspaces use the same PVC. When the user runs one workspace, it only binds to one node in the cluster at a time.
+|===
 
-The corresponding containers volume mounts link to a common volume, and sub-paths are prefixed with `_<workspace-ID>_` or `__<original-PVC-name>__`. For more details, see xref:how-subpaths-are-used-in-pvcs_{context}[].
+Persistent volumes (PVs) access mode::
+The nature of the persistent volume determines the available access mode: `ReadWriteMany` or `ReadWriteOnce`.
```
Srivaralakshmi:
```suggestion
The nature of the PV determines the available access mode: `ReadWriteMany` or `ReadWriteOnce`.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
Srivaralakshmi: These seem like restrictions. Suggest putting a suitable sub-heading such as limitations/restrictions. It will be helpful from the reader's perspective to scan the information easily and quickly.

Srivaralakshmi:
```suggestion
The workspace engine replaces volumes related to user-defined volumes with a subPath in the common volume. 
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-.Restrictions on using the `common` PVC strategy
+When the user starts a first non-ephemeral workspace, the workspace engine creates the common persistent volume.
 
-When the `common` strategy is used, and a workspace PVC access mode is ReadWriteOnce (RWO), only one node can simultaneously use the PVC.
+When the user starts a non-ephemeral workspace, the workspace engine uses the common persistent volume.
+The workspace only binds simultaneously to one node in the {orch-name} cluster.
+The workspace engine ignores user-defined volumes. 
+The workspace engine replaces volumes related to user-defined volumes with a subpath in the common volume. 
+Subpaths have a `_<workspace-ID>_` or `__<original-volume-name>__` prefix.
```
Srivaralakshmi:
```suggestion
SubPaths have a `_<workspace-ID>_` or `__<original-volume-name>__` prefix.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+See xref:how-subpaths-are-used-in-persistent-volumes_{context}[].
+The {prod-short} Volume name is identical to the name of the user-defined persistent volume.
```
Srivaralakshmi:
```suggestion
The {prod-short} volume name is identical to the name of the user-defined PV.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+Therefore, if a workspace container uses a {prod-short} volume with the same name as the user-defined persistent volume, they will use the same shared folder in the common persistent volume.
```
Srivaralakshmi:
```suggestion
Therefore, if a workspace container uses a {prod-short} volume with the same name as the user-defined PV, they will use the same shared folder in the common PV.
```

Srivaralakshmi: Question: 
In this sentence: "**they** will use the same shared folder", what do you mean by "they"?

- Workspace containers ?
- {prod-short} volumes?
- user-defined PVs?
- or both {prod-short} volumes and user-defined PVs ?

It is not clear to the reader. Suggest rewriting to provide more clarity.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-If there are several nodes, you can use the `common` strategy, but:
+When the user deletes a workspace, the workspace engine deletes the corresponding subdirectory (`$\{ws-id}`) in the persistent volume directory.
```
Srivaralakshmi:
```suggestion
When the user deletes a workspace, the workspace engine deletes the corresponding subdirectory (`$\{ws-id}`) in the PV directory.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-* The workspace PVC access mode must be reconfigured to `ReadWriteMany` (RWM), so multiple nodes can use this PVC simultaneously.
+When the user deletes the last workspace, the workspace engine removes the common persistent volume.
```
Srivaralakshmi:
```suggestion
When the user deletes the last workspace, the workspace engine removes the common PV. 
```

Srivaralakshmi: Suggestion: Would it be a good idea to come up with a table with columns such as user action and Behavior of the workspace engine, with an appropriate table heading? This will simplify the content and improve the readability as the reader can scan, map, and understand the content easily and quickly.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+Therefore, if a workspace container uses a {prod-short} volume with the same name as the user-defined persistent volume, they will use the same shared folder in the common persistent volume.
 
-If there are several nodes, you can use the `common` strategy, but:
+When the user deletes a workspace, the workspace engine deletes the corresponding subdirectory (`$\{ws-id}`) in the persistent volume directory.
 
-* The workspace PVC access mode must be reconfigured to `ReadWriteMany` (RWM), so multiple nodes can use this PVC simultaneously.
+When the user deletes the last workspace, the workspace engine removes the common persistent volume. 
 
-* Only one workspace in the same {orch-namespace} may be running. See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
+Restrictions on the `common` storage strategy with `ReadWriteOnce` access mode::
```
Srivaralakshmi:
```suggestion
Restrictions on the `common` storage strategy with the `ReadWriteOnce` access mode::
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+The `ReadWriteOnce` access mode limits each user to run only one concurrent workspace.
+See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
 
-The `common` PVC strategy is not suitable for large multi-node clusters. Therefore, it is best to use it in single-node clusters. However, in combination with the `per-workspace` {orch-namespace} strategy, the `common` PVC strategy is usable for clusters with not more than 75 nodes. The PVC used with this strategy must be large enough to accommodate all projects to prevent a situation in which one project depletes the resources of others.
+Scalability::
+The `common` storage strategy is not suitable for multinode clusters with 'ReadWriteOnce' access mode for persistent volumes if the number of concurrently running workspaces per user is more than 1
```
Srivaralakshmi:
```suggestion
The `common` storage strategy is not suitable for multi-node clusters with the 'ReadWriteOnce' access mode for PVs if the number of concurrently running workspaces per user is more than 1.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-[id="the-per-workspace-pvc-strategy_{context}"]
-== The `per-workspace` PVC strategy
+Persistent volumes (PV) provisioning::
+Create large enough persistent volume to accommodate all projects to prevent a situation in which one project depletes the resources of others.
```
Srivaralakshmi:
```suggestion
Create a large enough PV to accommodate all projects to prevent a situation in which one project depletes the resources of others.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-The `per-workspace` strategy is similar to the `common` PVC strategy. The only difference is that all workspace Volumes, but not all the workspaces, use the same PVC as the default data storage for:
+[id="the-per-workspace-storage-strategy_{context}"]
+== The `per-workspace` storage strategy
 
-* projects
-* workspace logs
-* additional Volumes defined by a user
+Each workspace uses one dedicated persistent volume.
```
Srivaralakshmi:
```suggestion
Each workspace uses one dedicated PV. 
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-The `per-workspace` strategy is similar to the `common` PVC strategy. The only difference is that all workspace Volumes, but not all the workspaces, use the same PVC as the default data storage for:
+[id="the-per-workspace-storage-strategy_{context}"]
+== The `per-workspace` storage strategy
 
-* projects
-* workspace logs
-* additional Volumes defined by a user
+Each workspace uses one dedicated persistent volume. 
+All {prod-short} Volumes defined in one workspace use the same persistent volume.
```
Srivaralakshmi:
```suggestion
All {prod-short} volumes defined within a single workspace use the same PV.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-With this strategy, {prod-short} keeps its workspace data in assigned PVs that are allocated by a single PVC.
+Scalability::
+Consider enabling this strategy on large multinode clusters with a higher amount of users.
```
Srivaralakshmi:
```suggestion
Consider enabling this strategy on large multi-node clusters with a higher amount of users.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-The `per-workspace` PVC strategy is the universal strategy out of the PVC strategies available and acts as a proper option for large multi-node clusters with a higher amount of users. Using the `per-workspace` PVC strategy, users can run multiple workspaces simultaneously, which results in more PVCs being created.
+Persistent volumes (PV) provisioning::
+Users can run multiple workspaces simultaneously. It results in more persistent volumes.
```
Srivaralakshmi:
```suggestion
Users can run multiple workspaces simultaneously. This action results in more PVs.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-[id="the-unique-pvc-strategy_{context}"]
-== The `unique` PVC strategy
+[id="the-unique-storage-strategy_{context}"]
+== The `unique` storage strategy
 
-Using the `unique `PVC strategy, every {prod-short} Volume of a workspace has its own PVC. The workspace PVCs are then:
+Each {prod-short} Volume defined in a workspace has its own persistent volume.
```
Srivaralakshmi:
```suggestion
Each {prod-short} volume defined in a workspace has its own PV.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-* Created when a workspace starts for the first time.
-* Deleted when a corresponding workspace is deleted.
+When the user starts a workspace, the workspace engine creates the workspace persistent volumes.
```
Srivaralakshmi:
```suggestion
When the user starts a workspace, the workspace engine creates the workspace PVs.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-User-defined PVCs are created with the following specifics:
+The workspace engine generates an unique name for each persistent volume to prevent naming conflicts with other persistent volumes in the same {orch-namespace}.
```
Srivaralakshmi:
```suggestion
The workspace engine generates a unique name for each PV to prevent name conflicts with other PVs in the same {orch-namespace}.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-* They are provisioned with generated names to prevent naming conflicts with other PVCs in a {orch-namespace}.
+To ensure that different storage strategies use the same persistent volume data structure, subpaths of the mounted persistent volumes that reference user-defined persistent volumes are prefixed with `_<workspace-ID>_` or `__<persistent-volume-name>__`.
```
Srivaralakshmi:
```suggestion
To ensure that different storage strategies use the same PV data structure, subPaths of the mounted PVs that reference user-defined PVs are prefixed with `_<workspace-ID>_` or `__<persistent-volume-name>__`. 
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-User-defined PVCs are created with the following specifics:
+The workspace engine generates an unique name for each persistent volume to prevent naming conflicts with other persistent volumes in the same {orch-namespace}.
 
-* They are provisioned with generated names to prevent naming conflicts with other PVCs in a {orch-namespace}.
+To ensure that different storage strategies use the same persistent volume data structure, subpaths of the mounted persistent volumes that reference user-defined persistent volumes are prefixed with `_<workspace-ID>_` or `__<persistent-volume-name>__`. 
+See xref:how-subpaths-are-used-in-persistent-volumes_{context}[].
 
-* Subpaths of the mounted Physical persistent volumes that reference user-defined PVCs are prefixed with `_<workspace-ID>_` or `__<PVC-name>__`. This ensures that the same PV data structure is configured with different PVC strategies. For details, see xref:how-subpaths-are-used-in-pvcs_{context}[].
+When the user deletes a workspace, the workspace engine deletes all workspace persistent volumes.
```
Srivaralakshmi:
```suggestion
When the user deletes a workspace, the workspace engine deletes all workspace PVs.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-The `unique` PVC strategy is suitable for larger multi-node clusters with a lesser amount of users. Since this strategy operates with separate PVCs for each volume in a workspace, vastly more PVCs are created.
+Persistent volumes (PV) provisioning::
+This is the strategy creating the highest volumes count.
```
Srivaralakshmi:
```suggestion
This is the strategy, which creates the highest volume counts.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
 
-[id="how-subpaths-are-used-in-pvcs_{context}"]
-== How subpaths are used in PVCs
+[id="how-subpaths-are-used-in-persistent-volumes_{context}"]
+== How subpaths are used in persistent volumes
```
Srivaralakshmi:
```suggestion
== How subPaths are used in persistent volumes (PVs)
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-* They are provisioned with generated names to prevent naming conflicts with other PVCs in a {orch-namespace}.
+To ensure that different storage strategies use the same persistent volume data structure, subpaths of the mounted persistent volumes that reference user-defined persistent volumes are prefixed with `_<workspace-ID>_` or `__<persistent-volume-name>__`. 
+See xref:how-subpaths-are-used-in-persistent-volumes_{context}[].
 
-* Subpaths of the mounted Physical persistent volumes that reference user-defined PVCs are prefixed with `_<workspace-ID>_` or `__<PVC-name>__`. This ensures that the same PV data structure is configured with different PVC strategies. For details, see xref:how-subpaths-are-used-in-pvcs_{context}[].
+When the user deletes a workspace, the workspace engine deletes all workspace persistent volumes.
 
-The `unique` PVC strategy is suitable for larger multi-node clusters with a lesser amount of users. Since this strategy operates with separate PVCs for each volume in a workspace, vastly more PVCs are created.
+Persistent volumes (PV) provisioning::
```
Srivaralakshmi:
```suggestion
PV provisioning::
```

Srivaralakshmi: ^^ This is a global change.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+This is the strategy creating the highest volumes count.
 
 
-[id="how-subpaths-are-used-in-pvcs_{context}"]
-== How subpaths are used in PVCs
+[id="how-subpaths-are-used-in-persistent-volumes_{context}"]
+== How subpaths are used in persistent volumes
 
 Subpaths illustrate the folder hierarchy in the Persistent Volumes (PV).
```
Srivaralakshmi:
```suggestion
SubPaths illustrate the folder hierarchy in the PV.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+    /__<user-defined-volume-name-3|volume3>__
     ...
 ----
 
-When a user defines volumes for components in the devfile, all components that define the volume of the same name will be backed by the same directory in the PV as `__<PV-name>__`, `__<workspace-ID>__, or `__<original-PVC-name>__`. Each component can have this location mounted on a different path in its containers.
-
-
-.Example
-Using the `common` PVC strategy, user-defined PVCs are replaced with subpaths on the common PVC. When the user references a volume as `my-volume`, it is mounted in the common-pvc with the `/workspace-id/my-volume` subpath.
+When a user defines volumes for components in the devfile, all components that define the volume of the same name will be backed by the same directory in the persistent volume as `__<persistent-volume-name>__`, `__<workspace-ID>__`, or `__<original-volume-name>__`.
```
Srivaralakshmi:
```suggestion
When a user defines volumes for components in the devfile, all components that define the volume of the same name will be backed by the same directory in the PV as `__<persistent-volume-name>__`, `__<workspace-ID>__`, or `__<original-volume-name>__`.
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
+Each component can have this location mounted on a different path in its containers.
```
Srivaralakshmi, line 125: Question: What do you mean by "this location"? 
Suggestion: Rewrite to provide more clarity. An example will help to better understand what you are saying in the preceding sentence.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
Srivaralakshmi: > volume of the same name will be backed by the same directory

Suggestion: Rewrite to simplify the meaning and bring in more clarity; an example will definitely be helpful. It will also prevent confusion. As a result, for the reader, it will be easier and quicker to read and understand the concept and process.

Srivaralakshmi (changes requested):
@themr0c Left some suggestions. PTAL. Thanks!

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c:
```suggestion
| PV count
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c:
```suggestion
| PV count depends on the number of workspaces
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c:
```suggestion
| Unpredictable PV count
```

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c: I believe we can keep it as is because this complete piece of documentation will go way soon.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,123 @@
...
 
-* projects
-* workspace logs
-* additional Volumes defined by a user
+Each workspace uses one dedicated PV. 
+All {prod-short} volumes defined within a single workspace use the same PV.
 
-With this strategy, {prod-short} keeps its workspace data in assigned PVs that are allocated by a single PVC.
+Scalability::
+Consider enabling this strategy on large multi-node clusters with a higher amount of users.
```
ibuziuk: I would probably drop this scalability recommendation at all. For me, it does not make much sense and it is much more ambiguous than useful. The scalability recommendation would be related to the number of PVs, which is already covered above

ibuziuk (commented):
LGTM, Just lets drop the scalability recommendation for per-workspace strategy

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
themr0c:
```suggestion
```

ibuziuk (approved):

rkratky (approved):

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,121 @@
...
+This is the default storage strategy.
+For each user, all workspaces use the same PV for the default data storage.
 
-.Restrictions on using the `common` PVC strategy
+When the user starts a first non-ephemeral workspace, the workspace engine creates a common PV.
 
-When the `common` strategy is used, and a workspace PVC access mode is ReadWriteOnce (RWO), only one node can simultaneously use the PVC.
+When the user starts another non-ephemeral workspace, the workspace engine uses the same common PV.
+The workspace only binds simultaneously to one node in the {orch-name} cluster.
+The workspace engine ignores user-defined volumes.
```
l0rd, line 54: I don't think this is right. In this paragraph you are assuming that a devfile `volume` is a Kubernetes `PersistentVolume`. But that's not accurate. We do not mention that in the devfile specification. The only garantee is that the devfile volume will be translated in a volume (escaping the Copy-on-Write container filesystem) and that it will be persisted after a workspace restart (unless `attributes.persistVolumes` is set to `false`). The kind of volume that will be used by Che is just an implementation detail: saying that "user-defined" volumes are ignoered or that "user-defined volumes are replaced with volumes with subPath" is not right. I would remove these 2 sentences.

l0rd (approved):
I have added a minor comment but in general LGTM. Thanks @themr0c for your updates.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,121 @@
...
+SubPaths have a `_<workspace-ID>_` or `__<original-volume-name>__` prefix.
+See xref:how-subpaths-are-used-in-persistent-volumes_{context}[].
+The {prod-short} volume name is identical to the name of the user-defined PV.
+Therefore, if a workspace container uses a {prod-short} volume with the same name as the user-defined PV, they will use the same shared folder in the common PV.
 
-If there are several nodes, you can use the `common` strategy, but:
+When the user deletes a workspace, the workspace engine deletes the corresponding subdirectory (`$\{ws-id}`) in the PV directory.
 
-* The workspace PVC access mode must be reconfigured to `ReadWriteMany` (RWM), so multiple nodes can use this PVC simultaneously.
+When the user deletes the last workspace, the workspace engine removes the common PV.
```
Srivaralakshmi, line 63: Suggestion: Group the related items in a list and have these listed as bullet points. Seeing all these different possibilities and reading them in one paragraph is difficult as a user/reader to understand and process the content in your mind.

I hope it is not just me. See what is helpful from the user/reader's perspective and do the needful.

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,121 @@
...
-If there are several nodes, you can use the `common` strategy, but:
+When the user deletes a workspace, the workspace engine deletes the corresponding subdirectory (`$\{ws-id}`) in the PV directory.
 
-* The workspace PVC access mode must be reconfigured to `ReadWriteMany` (RWM), so multiple nodes can use this PVC simultaneously.
+When the user deletes the last workspace, the workspace engine removes the common PV. 
 
-* Only one workspace in the same {orch-namespace} may be running. See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
+Restrictions on the `common` storage strategy with the `ReadWriteOnce` access mode::
+The `ReadWriteOnce` access mode limits each user to run only one concurrent workspace.
+See xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc[].
```
Srivaralakshmi, line 67: Suggestion: Please fill in a friendly text inside [ ]. At present it is not saying anything here: xref:installation-guide:configuring-the-number-of-workspaces-that-a-user-can-run.adoc**[]**

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,121 @@
...
 
-The `common` PVC strategy is not suitable for large multi-node clusters. Therefore, it is best to use it in single-node clusters. However, in combination with the `per-workspace` {orch-namespace} strategy, the `common` PVC strategy is usable for clusters with not more than 75 nodes. The PVC used with this strategy must be large enough to accommodate all projects to prevent a situation in which one project depletes the resources of others.
+Scalability::
+The `common` storage strategy is not suitable for multi-node clusters with the 'ReadWriteOnce' access mode for PVs if the number of concurrently running workspaces per user is more than 1.
 
-[id="the-per-workspace-pvc-strategy_{context}"]
-== The `per-workspace` PVC strategy
+Persistent volumes (PV) provisioning::
```
Srivaralakshmi, line 72: Suggestion: Choose any one: Persistent volume (PV)  or Persistent volumes (PV)s - global change (for lines 82 and 99 as well).

modules/installation-guide/partials/con_persistent-volume-configuration.adoc
```diff
@@ -5,95 +5,121 @@
...
+Users can run multiple workspaces simultaneously. This action results in more PVs.
 
-[id="the-unique-pvc-strategy_{context}"]
-== The `unique` PVC strategy
+[id="the-unique-storage-strategy_{context}"]
+== The `unique` storage strategy
 
-Using the `unique `PVC strategy, every {prod-short} Volume of a workspace has its own PVC. The workspace PVCs are then:
+Each {prod-short} volume defined in a workspace has its own PV.
```
Srivaralakshmi, line 88: Suggestion: has its own/has its own persistent volume (PV).

Srivaralakshmi (commented):
@themr0c Thanks for addressing my suggestions. I have left a few more suggestions. PTAL. Thanks for your patience :)

PR #2171, opened 2021-11-23
chore: update Dockerfile

themr0c:

* Update Dockerfile.
* New version for all tools: antora, newdoc, shellcheck, vale ...

* Quay reports container image has security issues.
* Container image doesn't build.

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,36 +9,37 @@
 #   Red Hat, Inc. - initial implementation
 #
 
-FROM rust:1.49.0-alpine3.12 as newdoc-builder
-RUN cargo install newdoc
+FROM docker.io/library/rust:1.51-alpine3.13 as newdoc-builder
+RUN apk add --no-cache --update  musl-dev \
+    && cargo install newdoc
 
-FROM golang:1.15.6-alpine3.12 as htmltest-builder
+FROM docker.io/library/golang:1.15-alpine3.13 as htmltest-builder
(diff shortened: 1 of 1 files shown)


nickboldt (approved):
Would be better to use UBI images but +1

benoitf (approved):

Dockerfile
```diff
@@ -9,36 +9,37 @@
...
 #
 
-FROM rust:1.49.0-alpine3.12 as newdoc-builder
-RUN cargo install newdoc
+FROM docker.io/library/rust:1.51-alpine3.13 as newdoc-builder
+RUN apk add --no-cache --update  musl-dev \
+    && cargo install newdoc
 
-FROM golang:1.15.6-alpine3.12 as htmltest-builder
+FROM docker.io/library/golang:1.15-alpine3.13 as htmltest-builder
```
benoitf, line 16: 1.15 is a rolling tag so usually we stick to a tag that is not changing

themr0c: You mean to make sure we can rebuild exactly the same? 
I see this here as a feature to enable to build newer images including security fixes without changing the Dockerfile in master.

benoitf: yes, to have something that is reproduced all the time

about security fixes updates, probably that dependabot can provide such enhancements ?

PR #2200, opened 2021-12-15
chore: externalize Vale styles (for the master branch)

themr0c:

Externalize Vale styles. Use Vale styles maintained in external repositories:

- https://github.com/vale-at-red-hat/vale-at-red-hat/releases/latest/download/RedHat.zip
-  https://github.com/vale-at-red-hat/chedocs/releases/latest/download/CheDocs.zip 

Single source for the Vale style used for all projects using the RedHat Style Guide, and the Vale style used for other Eclipse Che projects (the blog).

7.41.0

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


.vale.ini
```diff
@@ -43,17 +37,17 @@ properties = md
...
 
 [*.ini]
-BasedOnStyles = CheDocs,IBM
-IBM.Slash = NO
-IBM.Spacing = NO
-CheDocs.Spelling = NO
+BasedOnStyles = RedHat
+RedHat.Slash = NO
+RedHat.Spacing = NO
+RedHat.Spelling = NO
```
themr0c, line 53:
```suggestion
RedHat.Spelling = NO
RedHat.url = NO

```

.vale.ini
```diff
@@ -43,17 +37,19 @@ properties = md
...
+RedHat.url = NO
```
themr0c, line 54:
```suggestion
RedHat.url = NO
RedHat.ConfigMap  = NO
```

.vale.ini
```diff
@@ -43,17 +37,20 @@ properties = md
...
+RedHat.ConfigMap  = NO
```
themr0c, line 55:
```suggestion
RedHat.ConfigMap  = NO
RedHat.write = NO
```

benoitf (approved):
I think it might be good to check sha before always downloading the same file when we're launching the command but could be an improvement

Also styles could be cached in github action as well using sha as a key as well to be more robust/faster

tolusha (approved):

themr0c:
Good ideas @benoitf. Maybe for the next iteration. Let's see first if downloading the styles is consuming too much time.

themr0c:
> I think it might be good to check sha before always downloading the same file when we're launching the command but could be an improvement
> 
> Also styles could be cached in github action as well using sha as a key as well to be more robust/faster

=> https://issues.redhat.com/browse/RHDEVDOCS-3601

themr0c:
https://git-scm.com/ is down today :(

PR #2213, opened 2022-01-18
chore: Update Dockerfile.

themr0c:
Major changes:

- Upgrading `antora` to 3.0.0, as Antora 2.3 will enter maintenance on January 25, 2022 and reach end of life (EOL) on February 25, 2022.
- Adding Antora Lunr Extension. See https://gitlab.com/antora/antora-lunr-extension.
- Upgrading `vale` to 2.14.0
- Adding `diagrams`
- Removing `linkckecker`

We need to merge https://github.com/eclipse/che-docs/pull/2212 before publishing the new container.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,0 +1,43 @@
+#
+# Copyright (c) 2021 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+name: Build and publish container
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to docker.io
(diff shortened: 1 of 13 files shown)


benoitf (approved):

.gitignore
```diff
@@ -40,3 +40,4 @@ linkchecker-out.html
 .local
 .yarnrc
 yarn.lock
+package.json
```
benoitf, line 43: looks odd to ignore this file ?

themr0c: We don't need it. It's presence is misleading because we don't use it or maintain it, but it could lead people to think they should install Antora locally rather than using the container.

themr0c:
It seems I will need to publish the container to enable the build to pass...

As the integrated search is a key feature we were waiting for since long, I will move forward quickly.

themr0c:
I finished my round of patching projects depending on the che-docs container image. I will publish the new image.

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) 2021 Red Hat, Inc.
```
benoitf:
```suggestion
# Copyright (c) 2022 Red Hat, Inc.
```

Dockerfile
```diff
@@ -76,27 +77,28 @@ RUN apk add --no-cache --update \
...
 
 VOLUME /projects
 WORKDIR /projects
 ENV HOME="/projects" \
-    NODE_PATH="/usr/local/share/.config/yarn/global/node_modules"
+    NODE_PATH="/usr/local/share/.config/yarn/global/node_modules" \
+    USER_NAME=che-docs
 
-USER 1001
+USER 1000
```
themr0c:
```suggestion
USER 1001
```

PR #2216, opened 2022-01-24
chore: location for eclipse-che/che-docs-vale-style

themr0c:
New location: https://github.com/eclipse-che/che-docs-vale-style


benoitf (approved):

benoitf:
just curious: why the vale style of che-docs is not part of che-docs repository ?

themr0c:
> just curious: why the vale style of che-docs is not part of che-docs repository ?

We use the "Release" feature in GitHub to distribute the CheDocs style (for use in Eclipse Che blog). It would look awkward in the che-docs repository, where we would expect a release to be a release of the Che docs. Using a dedicated repository leaves no place for ambiguity.

Additionally, we can have a different release workflow on this repository: we don't require a double review.

rkratky (approved):

PR #2265, opened 2022-03-30
feat: Provides 'Adding a Visual Studio Code Extension' module

benoitf:
Introduce `Adding a Visual Studio Code Extension` module

https://github.com/eclipse/che/issues/21296

master

## Pull request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:
    - [x] Dashboard [default branding data](https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/services/bootstrap/branding.constant.ts)

Change-Id: Ide2764fe5bb63fcd5d7592d5c75d67622a7b1cac
Signed-off-by: Florent Benoit <fbenoit@redhat.com>

reviewer: @deerskindoll

### DIFF
--- a/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
+++ b/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
@@ -9,7 +9,7 @@ Devfiles in the devfile v2 format don't contain any IDE-specific plug-ins or Vis
 [id="plug-in-registries"]
 == {prod-short} plug-in registries overview
 
-{prod-short} has a default plug-in and extensions registry installed in every {prod-short} instance. The Che-Theia IDE searches and installs the plug-ins and extensions from that location.
+Every {prod-short} instance has a registry of default plug-ins and extensions. The Che-Theia IDE gets information about these plug-ins and extensions from the registry and installs them.
 
 Check this {prod-short} https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of default plug-ins, extensions, and source code. An online instance that is refreshed after every commit to the main branch, is located https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air gap mode: only the default registry is available there.
 
--- a/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
+++ b/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
@@ -65,7 +65,7 @@ Define the plug-ins to be installed in the workspace.
 - id: redhat/java  
 ----
 
-=== Changing default memory limit
+=== Changing the default memory limit
 
 Override container settings such as the memory limit.
 
--- a/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
+++ b/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
@@ -16,9 +16,9 @@ Check this {prod-short} https://github.com/eclipse-che/che-plugin-registry[regis
 The plug-in and extension overview for Che-Code Visual Studio Code editor is located in the https://www.open-vsx.org/[OpenVSX instance]. Air gap mode is not yet supported for this editor.
 
 [id="visual-studio-code-extensions-json"]
-== Adding an extension to .vscode/extensions.json
+== Adding an extension to `.vscode/extensions.json`
 
-The easiest way to add a Visual Studio Code extension to a workspace is to add it to the .vscode/extensions.json file. The main advantage of this method is that it works with all supported {prod-short} IDEs.
+The easiest way to add a Visual Studio Code extension to a workspace is to add it to the `.vscode/extensions.json` file. The main advantage of this method is that it works with all supported {prod-short} IDEs.
 
 If you use the Che-Theia IDE, the extension is installed and configured automatically. If you use a different supported IDE with the Che-Code Visual Studio Code fork, a pop-up with a recommendation to install the extension shows when you start the workspace.
 
@@ -26,7 +26,7 @@ If you use the Che-Theia IDE, the extension is installed and configured automati
 . You have the `.vscode/extensions.json` file in the root of the GitHub repository.
 
 .Procedure
-. Add the extension id to .vscode/extensions.json file. Use a . sign to separate the publisher and extension names:
+. Add the extension id to `.vscode/extensions.json` file. Use a `.` sign to separate the publisher and extension names:
 +
 *EXAMPLE*
 [source,json,subs="+quotes"]
@@ -43,7 +43,7 @@ If the specified set of extension ids isn't available in the {prod-short} regist
 
 
 [id="che-theia-plug-ins-YAML"]
-== Adding plug-in parameters to .che/che-theia-plugins.yaml
+== Adding plug-in parameters to `.che/che-theia-plugins.yaml`
 You can add extra parameters to a plug-in by modifying the `.che/che-theia-plugins.yaml` file. These modifications include:
 * Defining the plug-ins for workspace installation.
 * Changing default memory limit.
@@ -117,9 +117,9 @@ Override the default preferences of the Visual Studio Code extension for the wor
 [id="visual-studio-code-extensions-in-devfile"]
 == Defining Visual Studio Code extension attributes in the devfile
 
-In case it's not possible to add extra files in the GitHub repository, you can define some of the plug-in or extension attributes by inlining them in the devfile. You can use this procedure with both .vscode/extensions.json and .che/che-theia-plugins.yaml file contents.
+In case it's not possible to add extra files in the GitHub repository, you can define some of the plug-in or extension attributes by inlining them in the devfile. You can use this procedure with both `.vscode/extensions.json` and `.che/che-theia-plugins.yaml` file contents.
 
-=== Inlining .vscode/extensions.json file
+=== Inlining `.vscode/extensions.json` file
 Use `.vscode/extensions.json` file contents to inline extension attributes in the devfile.
 
 .Procedure
@@ -142,7 +142,7 @@ attributes:
     }
 ----
 
-=== Inlining .che/che-theia-plugins.yaml file
+=== Inlining `.che/che-theia-plugins.yaml` file
 Use `.che/che-theia-plugins.yaml` file contents to inline plug-in attributes in the devfile.
 
 .Procedure
--- a/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
+++ b/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
@@ -46,7 +46,7 @@ If the specified set of extension ids isn't available in the {prod-short} regist
 == Adding plug-in parameters to `.che/che-theia-plugins.yaml`
 You can add extra parameters to a plug-in by modifying the `.che/che-theia-plugins.yaml` file. These modifications include:
 * Defining the plug-ins for workspace installation.
-* Changing default memory limit.
+* Changing the default memory limit.
 * Overriding default preferences.
 
 
--- a/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
+++ b/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
@@ -4,14 +4,14 @@
 [id="adding-visual-studio-code-extension_{context}"]
 = Adding a Visual Studio Code extension to a workspace
 
-Previously, with the devfiles v1 format, you used the devfile to specify IDE-specific plug-ins and Visual Studio Code extensions. Now, with devfiles v2, you use a specific meta-folder instead of the devfile to specify the plug-ins and extensions.
+Previously, with the devfiles v1 format, you used the devfile to specify IDE-specific plug-ins and Visual Studio Code extensions. Now, with devfiles v2, you use a specific meta-folder rather than the devfile to specify the plug-ins and extensions.
 
 [id="plug-in-registries"]
 == {prod-short} plug-in registries overview
 
 Every {prod-short} instance has a registry of default plug-ins and extensions. The Che-Theia IDE gets information about these plug-ins and extensions from the registry and installs them.
 
-Check this {prod-short} link:https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of the default plug-ins, extensions, and source code. An online instance that refreshes fter every commit to the main branch, is located link:https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air-gapped environment: only the default registry is available there.
+Check this {prod-short} link:https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of the default plug-ins, extensions, and source code. An online instance that refreshes after every commit to the main branch, is located link:https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air-gapped environment: only the default registry is available there.
 
 The plug-in and extension overview for Che-Code Visual Studio Code editor is located in the link:https://www.open-vsx.org/[OpenVSX instance]. Air gap is not yet supported for this editor.
 
--- a/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
+++ b/modules/end-user-guide/pages/adding-a-vscode-extension.adoc
@@ -11,24 +11,23 @@ Previously, with the devfiles v1 format, you used the devfile to specify IDE-spe
 
 Every {prod-short} instance has a registry of default plug-ins and extensions. The Che-Theia IDE gets information about these plug-ins and extensions from the registry and installs them.
 
-Check this {prod-short} https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of default plug-ins, extensions, and source code. An online instance that is refreshed after every commit to the main branch, is located https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air gap mode: only the default registry is available there.
+Check this {prod-short} link:https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of the default plug-ins, extensions, and source code. An online instance that refreshes fter every commit to the main branch, is located link:https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air-gapped environment: only the default registry is available there.
 
-The plug-in and extension overview for Che-Code Visual Studio Code editor is located in the https://www.open-vsx.org/[OpenVSX instance]. Air gap mode is not yet supported for this editor.
+The plug-in and extension overview for Che-Code Visual Studio Code editor is located in the link:https://www.open-vsx.org/[OpenVSX instance]. Air gap is not yet supported for this editor.
 
 [id="visual-studio-code-extensions-json"]
 == Adding an extension to `.vscode/extensions.json`
 
 The easiest way to add a Visual Studio Code extension to a workspace is to add it to the `.vscode/extensions.json` file. The main advantage of this method is that it works with all supported {prod-short} IDEs.
 
-If you use the Che-Theia IDE, the extension is installed and configured automatically. If you use a different supported IDE with the Che-Code Visual Studio Code fork, a pop-up with a recommendation to install the extension shows when you start the workspace.
+If you use the Che-Theia IDE, the extension is installed and configured automatically. If you use a different supported IDE with the Che-Code Visual Studio Code fork, the IDE displays a pop-up with a recommendation to install the extension. 
 
 .Prerequisites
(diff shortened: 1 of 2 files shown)


ibuziuk (approved):

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,167 @@
...
+  .che/che-theia-plugins.yaml: |  
+    - id: redhat/java
+----
+
+[id="plug-in-registries"]
+== Plug-in registries
+
+{prod-short} provides a default plug-in registry installed on every {prod-short} instance.
+
+When using `Che-Theia` as IDE, it will search and install the extensions from there.
```
ibuziuk: my understanding that `che-theia` does not support the OpenVSX registry, right?

benoitf: correct

benoitf:
@deerskindoll it seems that after your changes, the Pull Request is not passing anymore the vale checks

deerskindoll:
requesting peer review

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
+:navtitle: Adding a Visual Studio Code extension to a workspace
+:keywords: Visual Studio Code extension, user-guide
+
+[id="adding-visual-studio-code-extension_{context}"]
+= Adding a Visual Studio Code extension to a workspace
+
+Devfiles in the devfile v2 format don't contain any IDE-specific plug-ins or Visual Studio Code extensions. Instead, the plug-ins and extensions are described in a specific meta-folder.
```
rolfedh: When I read this phrase, the current wording raises some questions about devfiles:
- Are you specifying `v2` because other versions _can_ contain these plug-ins and extensions?
- As a user, what agency do I have in the things these phrases describe? 
- The first and second phrases juxtapose "contain" with "describe." Would it be clearer to use "describe" in both cases? 
- What is a "meta-folder?" I don't see that term anywhere else in the CHE docs content. Is there a more familiar or obvious term?

For example:
```suggestion
With devfiles that use the devfile v2 format, you don't describe IDE-specific plug-ins or Visual Studio Code extensions. Instead, you describe them in a specific folder.
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+[id="plug-in-registries"]
+== {prod-short} plug-in registries overview
+
+{prod-short} has a default plug-in and extensions registry installed in every {prod-short} instance. The Che-Theia IDE searches and installs the plug-ins and extensions from that location.
```
rolfedh:
```suggestion
Every {prod-short} instance has a registry of default plug-ins and extensions. The Che-Theia IDE gets information about these plug-ins and extensions from the registry and installs them.
```
Are the plug-ins and extensions actually _in_ the registry, or does the registry point to their locations? This suggestion supposes the latter.

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+Check this {prod-short} https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of default plug-ins, extensions, and source code. An online instance that is refreshed after every commit to the main branch, is located https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air gap mode: only the default registry is available there.
```
rolfedh: In some che-docs topics, `link:` precedes URLs like this, but not in this PR. Is that okay?

rolfedh:
```suggestion
Check this {prod-short} https://github.com/eclipse-che/che-plugin-registry[registry project] for an overview of [the?] default plug-ins, extensions, and source code. An online instance that refreshes after every commit to the main branch is located https://eclipse-che.github.io/che-plugin-registry/main/v3/plugins/[here]. You can use a different plug-in or extension registry for Che-Theia if you don't work in air gap mode: only the default registry is available there.
```

rolfedh: I'm having trouble making sense of the final sentence. 
- I don't see an "air gap mode" mentioned in the che-docs. Would it help to say "air-gapped environment", link to a topic, or specify the key parameters for setting up an air-gapped environment, such as `airGapContainerRegistryHostname`?
- @benoitf, does "only the default registry is available there" mean "only the default registry is available in an air-gapped environment?" This seems like the opposite of what I would expect. I expect the user would need to specify a custom (non-default) plug-in or extension registry _because_ they are in an air-gapped environment. The default URLs of the plug-ins and extensions might not be reachable from an air-gapped system.

rolfedh: Using the definite article, "the default plug-ins", makes more sense if the "registry project" topic describes specific plug-ins. Otherwise, if it's an abstract discussion of "plug-ins, extensions, and source code," the current wording is fine.

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+The plug-in and extension overview for Che-Code Visual Studio Code editor is located in the https://www.open-vsx.org/[OpenVSX instance]. Air gap mode is not yet supported for this editor.
+
+[id="visual-studio-code-extensions-json"]
+== Adding an extension to `.vscode/extensions.json`
+
+The easiest way to add a Visual Studio Code extension to a workspace is to add it to the `.vscode/extensions.json` file. The main advantage of this method is that it works with all supported {prod-short} IDEs.
+
+If you use the Che-Theia IDE, the extension is installed and configured automatically. If you use a different supported IDE with the Che-Code Visual Studio Code fork, a pop-up with a recommendation to install the extension shows when you start the workspace.
```
rolfedh:
```suggestion
If you use the Che-Theia IDE, the extension is installed and configured automatically. If you use a different supported IDE with the Che-Code Visual Studio Code fork, the IDE [displays a pop-up with a recommendation/promts you] to install the extension when you start the workspace.
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+.Prerequisites
+. You have the `.vscode/extensions.json` file in the root of the GitHub repository.
+
+.Procedure
+. Add the extension id to `.vscode/extensions.json` file. Use a `.` sign to separate the publisher and extension names:
```
rolfedh:
```suggestion
. Add the extension ID to the `.vscode/extensions.json` file. Use a `.` sign to separate the publisher and extension names:
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+[source,json,subs="+quotes"]
+----
+  {
+    "recommendations": [
+      "redhat.java"
+    ]
+  }
+----
+
+*NOTE*
```
rolfedh: For comparison, I searched the che-docs repo for admonitions like one this but didn't find any. You can use the asciidoc formatting for admonitions, which looks like this:

```
NOTE: adlfjakdfj
```
or 
```
[NOTE]
====
afdasdf
====
```
The openshift-docs repo standardized on this later format... 
I would recommend asking your team which one to use.

rolfedh: This is how asciidoc renders the original format and the two proposed formats: 
image

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+----
+  {
+    "recommendations": [
+      "redhat.java"
+    ]
+  }
+----
+
+*NOTE* 
+If the specified set of extension ids isn't available in the {prod-short} registry, the workspace starts without the extension.
```
rolfedh: s/id/ID/ throughout.

benoitf:
who should handle the suggested changes from @rolfedh , @deerskindoll or myself ?

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+
+[id="che-theia-plug-ins-YAML"]
+== Adding plug-in parameters to `.che/che-theia-plugins.yaml`
+You can add extra parameters to a plug-in by modifying the `.che/che-theia-plugins.yaml` file. These modifications include:
```
rolfedh, line 45: Bulleted lists don't render as such unless they're preceded by a blank line.

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+* Defining the plug-ins for workspace installation.
+* Changing default memory limit.
```
rolfedh:
```suggestion
* Changing the default memory limit.
```

deerskindoll:
> 

I'll need your help if that's ok with you. I can handle asciidoc fixes but I can't answer the questions regarding devfile v2, air gap mode etc.

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+
+=== Defining the plug-ins for workspace installation
+Define the plug-ins to be installed in the workspace.
+
+.Prerequisites
+. You have the `.che/che-theia-plugins.yaml` file in the root of the GitHub repository.
+
+.Procedure
+. Add the id of the plug-in in the  `.che/che-theia-plugins.yaml` file. Use the `/` sign to separate the publisher and plug-in name:
```
rolfedh:
```suggestion
. Add the id of the plug-in to the  `.che/che-theia-plugins.yaml` file. Use the `/` sign to separate the publisher and plug-in name:
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
++
+*EXAMPLE*
+[source,yaml,subs="+quotes"]
+----
+- id: redhat/java  
+----
+
+=== Changing default memory limit
```
rolfedh:
```suggestion
=== Changing the default memory limit
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+Override container settings such as the memory limit.
```
rolfedh, line 70: It sounds like this procedure might  demonstrate how to override other settings besides the memory limit, but then it doesn't mention what they are. Maybe make this phrase more precise or mention the other settings below?

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+.Prerequisites
+. You have the `.che/che-theia-plugins.yaml` file in the root of the GitHub repository.
+
+.Procedure
+. Add an `override` section in the `.che/che-theia-plugins.yaml` file under the `id` of the plug-in.
```
rolfedh: add...to

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+=== Overriding default preferences
+
+Override the default preferences of the Visual Studio Code extension for the workspace.
+
+.Prerequisites
+. You have the `.che/che-theia-plugins.yaml` file in the root of the GitHub repository.
+
+.Procedure
+. Add an `override` section in the `.che/che-theia-plugins.yaml` file under the `id` of the extension.
```
rolfedh: "add...to" here and elsewhere

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+*EXAMPLE*
+[source,yaml,subs="+quotes"]
+----
+  - id: redhat/java  
+    override:  
+      preferences:
+        java.server.launchMode: LightWeight
+----
+
+*NOTE* You can also define the preferences in the `.vscode/settings.json` file, either by changing the preferences in the UI of your IDE or by manually adding them to the `.vscode/settings.json` file:
```
rolfedh:
```suggestion
*NOTE* You can also define the preferences in the `.vscode/settings.json` file, either by changing the preferences in the UI of your IDE or by adding them to the `.vscode/settings.json` file:
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+----
+{
+"my.preferences": "my-value"
+}
+----
+
+[id="visual-studio-code-extensions-in-devfile"]
+== Defining Visual Studio Code extension attributes in the devfile
+
+In case it's not possible to add extra files in the GitHub repository, you can define some of the plug-in or extension attributes by inlining them in the devfile. You can use this procedure with both `.vscode/extensions.json` and `.che/che-theia-plugins.yaml` file contents.
```
rolfedh: For minimalism:
- In case > If ?
- it's not possible to > you can't ?

Is there a more commonly understood equivalent for "inlining" that preserves the meaning?

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+If you use the Che-Theia IDE, the extension is installed and configured automatically. If you use a different supported IDE with the Che-Code Visual Studio Code fork, a pop-up with a recommendation to install the extension shows when you start the workspace.
+
+.Prerequisites
+. You have the `.vscode/extensions.json` file in the root of the GitHub repository.
+
+.Procedure
+. Add the extension ID to the `.vscode/extensions.json` file. Use a `.` sign to separate the publisher and extension names:
++
+*EXAMPLE*
```
rolfedh: Didn't find any other instances of `*EXAMPLE*` in che-docs. Consider adopting the standard code block titles throughout, like this:
```
.Example <optional descriptive text>
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
benoitf: > Are you specifying v2 because other versions can contain these plug-ins and extensions?

yes

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
benoitf: point to the location in Che (but can be sometimes stored in the registry)

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
benoitf: 'air gap' is also named 'restricted environment' sometimes. But you can find definition there https://en.wikipedia.org/wiki/Air_gap_(networking)

And it's used in Che (and docs)

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
benoitf: GitHub suggestion implies that we can commit directly the change.
Here it looks that there are two suggestions so it should be two separate GitHub PR suggestions (and people can approve one of them)

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
```diff
@@ -0,0 +1,163 @@
...
+}
+----
+
+[id="visual-studio-code-extensions-in-devfile"]
+== Defining Visual Studio Code extension attributes in the devfile
+
+In case it's not possible to add extra files in the GitHub repository, you can define some of the plug-in or extension attributes by inlining them in the devfile. You can use this procedure with both `.vscode/extensions.json` and `.che/che-theia-plugins.yaml` file contents.
+
+=== Inlining `.vscode/extensions.json` file
+Use `.vscode/extensions.json` file contents to inline extension attributes in the devfile.
```
rolfedh: Part of the issue with using the adjective, "inline," as a verb is that when you use it next to a noun, as shown here, the reader might experience some semantic noise:
raising the question, "Is a verb missing between "to" and "inline extension attributes?" 
Perhaps subsitute it with "add" or "insert?" WDYT?

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
rolfedh: Consider making that difference more explicit:
```suggestion
Previously, with the devfiles v1 format, you used the devfile to specify IDE-specific plug-ins and Visual Studio Code extensions. Now, with devfiles v2, you use a specific meta-folder instead of the devfile to specify the plug-ins and extensions.
```

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
rolfedh: > 'air gap' is also named 'restricted environment' sometimes. But you can find definition there https://en.wikipedia.org/wiki/Air_gap_(networking)
> 
> And it's used in Che (and docs)

Although we should strive to use the terminology consistently, I'm not highlighting "air gap" as a problem. I'm concerned that calling it a "mode" might prompt users to search for an actual software *mode* when it seems like there might not be one. I'm guessing you mean "an air-gapped or restricted environment" instead.

Otherwise, if there is something like a mode, it's not called a "mode" in the software or documentation and actually consists of one or more parameters that the user must set when operating the software in a restricted environment.

benoitf: I'll accept any suggested changes but here suggested changes have questions so I never saw this kind of suggested changes.

>
Does "only the default registry is available there" mean "only the default registry is available in an air-gapped environment?" This seems like the inverse of what I would expect: That the user probably requires use of a custom plug-in or extension registry because they are in an air-gapped environment. The default URLs of the plug-ins and extensions might not be reachable from an air-gapped system.

There is no need to use a custom plug-in or extension registry. Che deploy a registry on the air-gapped environment and in that case it includes all the binaries so there is no need to fetch something externally. 

> The default URLs of the plug-ins and extensions might not be reachable from an air-gapped system.

everything is reachable in that case (we build and deploy an offline / air-gapped environment registry)

rolfedh:
> who should handle the suggested changes from @rolfedh , @deerskindoll or myself ?

My role here is to serve as peer reviewer. I believe @deerskindoll, as the technical writer, should make the updates for language-related issues. I'll at-mention you, @benoitf, for any questions about the software functionality.

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: wording fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: wording fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: formatting fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: formatting fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: formatting fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: wording fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: wording fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: wording fixed

modules/end-user-guide/pages/adding-a-vscode-extension.adoc
deerskindoll: reworded as "inlining the extension/plug-in attributes"

deerskindoll (approved):
approved

benoitf:
Merging as all checks are green and it has been approved by Jana and Ilya

PR #2274, opened 2022-04-01
fix: update prod-url

max-cx:
The current {prod-url} value doesn't render correctly with `runnerpreview.sh`

## Pull request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


themr0c:
I apologize for the erroneous closing. My mistake.

max-cx:
If I understand it correctly, Fabrice's plan is to continue using the current value with `pass:c,a,q[{prod-url}]` everywhere.

PR #2301, opened 2022-04-21
procedures: Configuring workspace target namespace, Provisioning namespaces

themr0c:
Configuring workspace target namespace, Provisioning namespaces

references: https://github.com/eclipse/che/issues/21284

Update configuring-namespace-strategies

Firefox_Screenshot_2022-04-26T14-06-33 568Z

Firefox_Screenshot_2022-04-26T14-06-49 187Z

Firefox_Screenshot_2022-04-26T14-07-04 905Z

Remove "namespace strategy" concept.
Refresh with current situation.

## Pull request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,36 @@
...
+* Set the `server.workspaceNamespaceDefault` property in CheCluster Custom Resource:
++
+[subs="+quotes,+attributes"]
+----
+...
+spec:
+  server:
+    workspaceNamespaceDefault: __{prod-workspace}__-__<username>__
+...
+----
```
ibuziuk: ~~I do not think we support this. My understanding is that now we only have one namespace per-user strategy~~

ibuziuk: oh, sorry it is probably the same as 
```
server:
      customCheProperties:
        CHE_INFRA_KUBERNETES_NAMESPACE_DEFAULT: <username>-dev
```
which is supported

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,36 @@
...
+For each user, {prod-short} isolates the workspaces in a {orch-namespace}.
+You can configure the {orch-namespace} name template.
+
+{prod-short} server checks usernames and IDs for compatibility with {orch-name} objects naming convention before creating a {orch-namespace} from a template:
+
+*  Incompatible usernames or IDs are reduced to the nearest valid name by replacing groups of unsuitable symbols with the `-` symbol. 
+* The addition of a random 6-symbol suffix prevents IDs from collisions. The result is stored in preferences for reuse.
+* {kubernetes} limits the length of a {orch-namespace} name to 63 characters (this includes the evaluated placeholders). Additionally, the names (after placeholder evaluation) must be valid DNS names.
+* OpenShift limits the length further to 49 characters.
+* The `_<userid>_` placeholder is evaluated into a 36 character long UUID string.
```
ibuziuk: Interestingly so we do really support the 36 chars postfix in the namespace names? was it verified?

modules/administration-guide/pages/configuring-workspaces-globally.adoc
```diff
@@ -6,7 +6,14 @@
 [id="configuring-workspaces-globally_{context}"]
 = Configuring workspaces globally
 
+This ssection describes how an administrator can configure workspaces globally.
```
ibuziuk:
```suggestion
This section describes how an administrator can configure workspaces globally.
```

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+This procedure describes how to provision workpaces {orch-namespace}s in advance, rather than relying on automatic provisioning. 
+
+.Prerequisites
+
+* The workspace user cannot create namespaces on {orch-name}.
+
+.Procedure
+
+. Create one {orch-namespace} for each user, with following labels and annotations. Do not create multiple namespaces for a single user: it may lead to undefined behavior.
```
ibuziuk: should be a WARNING: probably

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,112 @@
...
+    app.kubernetes.io/component: workspaces-namespace
+  annotations: <2>
+    che.eclipse.org/username: __<username>__ <3>
+----
+<1> `server.customCheProperties.CHE_INFRA_KUBERNETES_NAMESPACE_LABELS` property value
+<2> `server.customCheProperties.CHE_INFRA_KUBERNETES_NAMESPACE_ANNOTATIONS` property value
+<3> Workspace user username on {orch-name}
+
+
+. Ensure the workspace user has `admin` role privileges in the target namespace:
```
ibuziuk: edit permissions should be enough

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
themr0c: Meaning an admin cannot configure the namespace names?

themr0c: OK

modules/administration-guide/pages/provisioning-namespaces.adoc
ibuziuk: https://github.com/eclipse/che/issues/21331

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,112 @@
...
+----
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: admin
+  namespace: __<namespace>__ <1>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
```
ibuziuk: should be changed to edit permissions. Basically `oc adm policy add-role-to-user edit ${USER}` should be enough here

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,112 @@
...
+  server:
+    server.customCheProperties.CHE_INFRA_KUBERNETES_NAMESPACE_LABEL=false
+...
+----
+
+.Additional resources
+
+* xref:understanding-the-checluster-custom-resource.adoc[]
+* xref:advanced-configuration-options-for-the-che-server-component.adoc[]
+
```
ibuziuk: hope that @metlos could review this part taken into account there is a pending https://github.com/eclipse-che/che-server/pull/286

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
ibuziuk: in any case, as a user / admin, I could not imagine having fun with namespace setup like `dev-26c52e9d-780d-42d2-9481-869357844e53` 
However, I might be missing smth. and in some cases that could potentially be useful

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+[id="configuring-workspace-target-namespace_{context}"]
+= Configuring workspace target {orch-namespace}
+
+For each user, {prod-short} isolates the workspaces in a {orch-namespace}.
+You can configure the {orch-namespace} name template.
+
+{prod-short} server checks usernames and IDs for compatibility with {orch-name} objects naming convention before creating a {orch-namespace} from a template:
+
+*  Incompatible usernames or IDs are reduced to the nearest valid name by replacing groups of unsuitable symbols with the `-` symbol.
```
ibuziuk:
```suggestion
* Incompatible usernames or IDs are reduced to the nearest valid name by replacing groups of unsuitable symbols with the `-` symbol. 
```

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,36 @@
...
+[id="configuring-workspace-target-namespace_{context}"]
+= Configuring workspace target {orch-namespace}
+
+For each user, {prod-short} isolates the workspaces in a {orch-namespace}.
+You can configure the {orch-namespace} name template.
+
+{prod-short} server checks usernames and IDs for compatibility with {orch-name} objects naming convention before creating a {orch-namespace} from a template:
+
+*  Incompatible usernames or IDs are reduced to the nearest valid name by replacing groups of unsuitable symbols with the `-` symbol. 
+* The addition of a random 6-symbol suffix prevents IDs from collisions. The result is stored in preferences for reuse.
```
ibuziuk: id are openshift UIDs AFAIK, I do not think we need any suffix here

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
themr0c: Che on Minikube: `<userid>-aka-<username>-namespace-for-che` translating to

`cgezegvsb2nhba-aka-user3-namespace-for-che-6m2w2b`
 so more a 14 character string.

```
| `workspaceNamespaceDefault` template value | resulting {orch-namespace} example

| <userid>-aka-<username>-namespace | cgezegvsb2nhba-aka-user3-namespace-for-che-6m2w2b

| <username>-namespace | user4-namespace 
```

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: I couldn't find the `kubectl` variant for `oc adm policy add-role-to-user edit ${USER}` :/.

themr0c:
Tested with Che on Minikube.

However, I am not sure how to reproduce the prerequisite "The workspace user cannot create namespaces on {orch-name}." for the second procedure. (Meaning that step 1 is sufficient). Therefore, for the next steps, the current state reflects what is created in nominal conditions.

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
themr0c: I observed that Minikube adds the suffix. Therore, valid on Kubernetes too.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: Removing entirely as the procedure should assert what the user must do, not what he must not do.

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
...
+metadata:
+  name: __<{orch-namespace}_name>__ <1>
+  labels:
+    app.kubernetes.io/part-of: che.eclipse.org <2>
+    app.kubernetes.io/component: workspaces-namespace <2>
+  annotations:
+    che.eclipse.org/username: __<username>__ <3>
+----
+<1> Use a {orch-namespace} name at your convenance
+<2> `server.customCheProperties.CHE_INFRA_KUBERNETES_NAMESPACE_LABELS` property value
```
l0rd: Reading notes `<2>` and `<3>` it looks like the admin has to set those `customCheProperties`. 99% of the time instead the admin should not care at all. These customChePropoerties are for whitelabelling / rebranding: we should mention it in the white labelling guide and we can just say here that those labels can be customizable as described in the white labelling guide...

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
...
+:page-aliases: 
+
+[id="configuring-{orch-namespace}s_{context}"]
+= Provisioning {orch-namespace}s
+
+This procedure describes how to provision workpaces {orch-namespace}s in advance, rather than relying on automatic provisioning. Repeat the procedure for each user.
+
+.Prerequisites
+
+* The workspace user cannot create namespaces on {orch-name}.
```
l0rd: That seams wrong: and admin can provision the namespace even if the workspace can create the namespaces.

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
+:_content-type: CONCEPT
+:navtitle: Provisioning {orch-namespace}s
+:keywords: administration guide, provisioning, {orch-namespace}
+:page-aliases: 
+
+[id="configuring-{orch-namespace}s_{context}"]
+= Provisioning {orch-namespace}s
```
l0rd: This guide is something that we may move to "Preparing the installation"

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
...
+    app.kubernetes.io/part-of: che.eclipse.org <2>
+    app.kubernetes.io/component: workspaces-namespace <2>
+  annotations:
+    che.eclipse.org/username: __<username>__ <3>
+----
+<1> Use a {orch-namespace} name at your convenance
+<2> `server.customCheProperties.CHE_INFRA_KUBERNETES_NAMESPACE_LABELS` property value
+<3> `server.customCheProperties.CHE_INFRA_KUBERNETES_NAMESPACE_ANNOTATIONS` property value, replacing __<username>__ with workspace user username on {orch-name}.
+
+. Create a role binding to ensure the __<username>__ workspace user has `edit` role privileges in the __<{orch-namespace}_name>__ {orch-namespace}:
```
l0rd:
```suggestion
. Create a role binding to ensure the __<username>__ user has `edit` role privileges in the __<{orch-namespace}_name>__ {orch-namespace}:
```

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
...
+    name: __<username>__
+    namespace: __<{orch-namespace}_name>__
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: edit
+----
+
+
+. Create a cluster role and cluster role binding to ensure the {prod-short} ServiceAccount has a accurate permissions on the __<{orch-namespace}_name>__ {orch-namespace}.
```
l0rd: This looks like a pre-requisite: it's done once whereas the other steps needs to be done for every new user / namespace.

l0rd: Another consideration is about the default Che SA rolebindings: aren't the permissions to list/get/update the namespaces already provisioned during the deploy? And is the `update` verb required in this case (the admin manages the namespaces)? These are questions for @tolusha and @metlos more than @themr0c

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: {prod-id-short} <1>
+    namespace: {prod-namespace} <2>
+roleRef:
+  kind: ClusterRole
+  name: {prod-service-account}-workspaces-namespaces-clusterrole
+  apiGroup: rbac.authorization.k8s.io
+----
+<1> {prod-short} service account name. Default: `{prod-id-short}`.
```
l0rd: This is similar to the note above about namespaces labels: that's a default that is fine 99% of the case. And that may be changed for white labelling purposes. But here we don't say if we can customize it and we do not link to the doc that says how to do that.

modules/administration-guide/pages/provisioning-namespaces.adoc
```diff
@@ -0,0 +1,118 @@
...
+<2> {prod-short} installation {orch-namespace} name. Default: {prod-namespace}.
+
+
+. Optionally, configure {prod-short} to not update the {orch-namespace} labels when starting a workspace.
```
l0rd: This step too only needs to be done once, not for every namespace.

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,58 @@
+:_content-type: CONCEPT
+:navtitle: Configuring workspace target {orch-namespace}
+:keywords: administration guide, configuring, namespace
+:page-aliases: installation-guide:configuring-namespace-strategies, configuring-namespace-strategies
+
+[id="configuring-workspace-target-namespace_{context}"]
+= Configuring workspace target {orch-namespace}
+
+For each user, {prod-short} isolates the workspaces in a {orch-namespace}.
+You can configure the {orch-namespace} name template.
```
l0rd: Here we should say that the target {orch-namespace} can be pre-created (link to the doc) or {prod-short} can create it the first time the user starts a workspace and the {orch-namespace} name template can be configured by the admin (this doc).

l0rd (approved):
This is great @themr0c. It's clear and goes straight to the point. I have added a few comments.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: With a default install, creating the namespace with labels and annotations is enough. A saw all the role bindings being created.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: OK remove callouts.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: Removed.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: Created a new section in Configuring.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: Removing, as che handles the role binding creation. What we have was more analysis document for the implementation, we don't need it in user docs.

themr0c:
Applied suggestions. Ready for language review I think.

l0rd (approved):
Fantastic 👍

modules/administration-guide/pages/configuring-namespace-provisioning.adoc
```diff
@@ -0,0 +1,17 @@
+:_content-type: CONCEPT
+:navtitle: Configuring {orch-namespace}s
+:keywords: administration guide, configuring, namespace
+:page-aliases: installation-guide:configuring-namespace-strategies, configuring-namespace-strategies
+
+[id="configuring-user-{orch-namespace}-provisioning_{context}"]
+= Configuring user {orch-namespace} provisioning
+
+For each user, {prod-short} isolates the workspaces in a {orch-namespace}.
```
bburt-rh:
```suggestion
For each user, {prod-short} isolates workspaces in a {orch-namespace}. 
```
No need for a definite article because you're speaking about a general function.

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
+:_content-type: PROCEDURE
+:navtitle: Configuring {orch-namespace} name
+:keywords: administration guide, configuring, namespace
+:page-aliases: 
+
+[id="configuring-user-{orch-namespace}-name-for-automatic-provisioning_{context}"]
```
bburt-rh:
```suggestion
[id="configuring-a-user-{orch-namespace}-name-for-automatic-provisioning_{context}"]
```

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
...
+= Configuring user {orch-namespace} name for automatic provisioning
```
bburt-rh:
```suggestion
= Configuring a user {orch-namespace} name for automatic provisioning
```
Need to use the indefinite article here because you're providing instructions for a specific example.

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+You can configure the {orch-namespace} name template that {prod-short} is using to create the required {orch-namespace} when starting a workspace.
```
bburt-rh:
```suggestion
You can configure the {orch-namespace} name template that {prod-short} uses to create the required {orch-namespace} when starting a workspace.
```
- Continuous tense incorrectly implies that the use of the template is a continuous, ongoing process that doesn't end.
- Should `starting a workspace` be `creating a workspace`?

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+A valid {orch-namespace} name template follows this convention:
```
bburt-rh:
```suggestion
A valid {orch-namespace} name template follows these conventions:
```
- Plural needed because there are multiple conventions in the list
- Also, consider making the items in the list more parallel in structure. Maybe the problem is with the word "conventions".  Not all of the items in the list are conventions. Some are limitations, and some are descriptions of the system processes. Maybe `guidelines` would be more accurate? Or maybe split it into two lists -- one with the naming conventions, and one explaining what the system does with the names to ensure adherence with the conventions?

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+* The `<username>` or `<userid>` placeholder is mandatory.
+
+* {prod-short} reduces usernames or IDs incompatible with {orch-name} objects naming convention to the nearest valid name by replacing groups of unsuitable symbols with the `-` symbol.
```
bburt-rh:
```suggestion
* Usernames and IDs cannot contain invalid characters. If the formatting of a username or ID is incompatible with the naming conventions for {orch-name} objects, {prod-short} changes the username or ID to a valid name by replacing incompatible characters with the `-` symbol. 
```
- Consider revising the sentence as suggested. The current sentence is awkwardly constructed with some unclear antecedents and is too "noun heavy" at times. 
- Also consider replacing "reduce" with "change" (the change may not be a reduction but merely a replacement) and "unsuitable" with "incompatible", which in this case is more accurate.

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+You can configure the {orch-namespace} name template that {prod-short} is using to create the required {orch-namespace} when starting a workspace.
+
+A valid {orch-namespace} name template follows this convention:
+
+* The `<username>` or `<userid>` placeholder is mandatory.
+
+* {prod-short} reduces usernames or IDs incompatible with {orch-name} objects naming convention to the nearest valid name by replacing groups of unsuitable symbols with the `-` symbol. 
+
+* {prod-short} evaluates the `<userid>` placeholder into a 14 character long string, and adds a random 6 character long suffix to prevents IDs from collisions. The result is stored in the user preferences for reuse.
```
bburt-rh:
```suggestion
* {prod-short} evaluates the `<userid>` placeholder into a 14 character long string, and adds a random six character long suffix to prevent IDs from colliding. The result is stored in the user preferences for reuse.
```
- Per ISG: `In text, use words for zero through nine. In text, use numerals for 10 and greater.`

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+* {kubernetes} limits the length of a {orch-namespace} name to 63 characters. 
+
+* OpenShift limits the length further to 49 characters.
+
+* The resulting name is a valid DNS name.
```
bburt-rh, line 23: Again, this isn't really a convention as suggested by the intro paragraph. It's the result of the system processing a name.

modules/administration-guide/pages/provisioning-namespaces-in-advance.adoc
```diff
@@ -0,0 +1,33 @@
...
+:page-aliases: 
+
+[id="preprovisioning-{orch-namespace}s_{context}"]
+= Provisioning {orch-namespace}s in advance
+
+You can provision workpaces {orch-namespace}s in advance, rather than relying on automatic provisioning. Repeat the procedure for each user.
+
+.Procedure
+
+* Create the __<{orch-namespace}_name>__ {orch-namespace} for __<username>__ user, with following labels and annotations.
```
bburt-rh:
```suggestion
* Create the __<{orch-namespace}_name>__ {orch-namespace} for __<username>__ user with the following labels and annotations:
```

modules/administration-guide/pages/provisioning-namespaces-in-advance.adoc
```diff
@@ -0,0 +1,33 @@
...
+apiVersion: v1
+metadata:
+  name: __<{orch-namespace}_name>__ <1>
+  labels:
+    app.kubernetes.io/part-of: che.eclipse.org
+    app.kubernetes.io/component: workspaces-namespace
+  annotations:
+    che.eclipse.org/username: __<username>__
+----
+<1> Use a {orch-namespace} name at your convenance
```
bburt-rh:
```suggestion
<1> Use a {orch-namespace} name of your choosing.
```
- `convenance` should be `convenience` but `convenience` isn't the correct word for this context.

modules/administration-guide/pages/configuring-workspace-target-namespace.adoc
bburt-rh:
```suggestion
```

This isn't really a convention as suggested by the intro paragraph for the list. It's the result of following the conventions. You can safely omit it because you already said that the list would define the naming conventions.

modules/administration-guide/pages/provisioning-namespaces.adoc
themr0c: During tests, it worked with only the first step. After discussion with @l0rd, removing the other steps.

PR #2320, opened 2022-05-13
chore: Added description to Admin and End user guides

swatikarmarkar:
Signed-off-by: swkarmar <swkarmar@redhat.com>

## Pull request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


modules/administration-guide/examples/system-variables.adoc
```diff
@@ -658,7 +658,7 @@ Default::: `+<username>-che+`
 
 == `+CHE_INFRA_KUBERNETES_NAMESPACE_LABEL+`
 
-Defines whether che-server should try to label the workspace namespaces.
+Defines whether che-server should try to label the workspace namespaces. NOTE: It is strongly recommended to keep the value of this property set to true. If false, the new workspace namespaces will not be labeled automatically and therefore not recognized by the {prod-short} operator making some features of DevWorkspaces not working. If false, an administrator is required to label the namespaces manually using the labels specified in che.infra.kubernetes.namespace.labels. If you want to manage the namespaces yourself, make sure to follow \https://www.eclipse.org/che/docs/stable/administration-guide/provisioning-namespaces-in-advance/. Any additional labels present on the namespace are kept in place and do not affect the functionality. Also note that the the administrator is free to pre-create and label the namespaces manually even if this property is true. No updates to the namespaces are done if they already conform to the labeling requirements.
```
amisevsk, line 661: This section could use some rewording (e.g. two consecutive sentences start with `If false,`). Likely the absolute docs link should be replaced with a relative, in-repo link for downstreaming.

Also it's not clear to me the relation between `CHE_INFRA_KUBERNETES_NAMESPACE_LABEL` and  `che.infra.kubernetes.namespace.labels`:
> If false, an administrator is required to label the namespaces manually using the labels specified in che.infra.kubernetes.namespace.labels

```suggestion
Defines whether che-server should try to label the workspace namespaces. NOTE: It is strongly recommended to keep the value of this property set to true. If false, the new workspace namespaces will not be labeled automatically and therefore not recognized by the {prod-short} operator making some features of DevWorkspaces not work. If false, an administrator is required to label the namespaces manually using the labels specified in `che.infra.kubernetes.namespace.labels`. If you want to manage the namespaces yourself, make sure to follow [relative-link]. Any additional labels present on the namespace are kept in place and do not affect the functionality. Also note that the the administrator is free to pre-create and label the namespaces manually even if this property is true. No updates to the namespaces are done if they already conform to the labeling requirements.
```

modules/administration-guide/examples/system-variables.adoc
```diff
@@ -676,7 +676,7 @@ Default::: `+true+`
 
 == `+CHE_INFRA_KUBERNETES_NAMESPACE_LABELS+`
 
-List of labels to find {orch-namespace} that are used for {prod-short} Workspaces. They are used to:  - find prepared {orch-namespace} for users in combination with `che.infra.kubernetes.namespace.annotations`.  - actively label {orch-namespace} with any workspace.
+List of labels to find {orch-namespace} that are used for {prod-short} Workspaces. They are used to:  - find prepared {orch-namespace} for users in combination with `che.infra.kubernetes.namespace.annotations`.  - actively label {orch-namespace} with any workspace. NOTE: It is strongly recommended not to change the value of this property because the {prod-short} operator relies on these labels and their precise values when reconciling DevWorkspaces. If this configuration is changed, the namespaces will not be automatically recognized by the {prod-short} operator as workspace namespaces unless manually labeled as such using the default labels and values. Additional labels on the namespace do not affect the functionality.
```
amisevsk, line 679:
```suggestion
List of labels to find {orch-namespace} that are used for {prod-short} Workspaces. They are used to:  - find prepared {orch-namespace} for users in combination with `che.infra.kubernetes.namespace.annotations`.  - actively label {orch-namespace} with any workspace. NOTE: It is strongly recommended not to change the value of this property because the {prod-short} operator relies on these labels and their precise values when reconciling workspace resources. If this configuration is changed, the namespaces will not be automatically recognized by the {prod-short} operator as workspace namespaces unless manually labeled as such using the default labels and values. Additional labels on the namespace do not affect the functionality.
```

modules/administration-guide/examples/system-variables.adoc
themr0c: This file is single-sourced from che properties in the che repository. 
To fix this part of the documentation, raise a pull request there:

* https://github.com/eclipse-che/che-server/blob/main/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
* https://github.com/eclipse-che/che-server/blob/main/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties

It is out of the real scope of this pull-request and should not be committed here.

modules/administration-guide/examples/checluster-properties.adoc
```diff
@@ -11,7 +11,7 @@ airGapContainerRegistryOrganization: Optional repository name of an alternate co
...
 cheClusterRoles: A comma-separated list of ClusterRoles that will be assigned to Che ServiceAccount. Each role must have `app.kubernetes.io/part-of=che.eclipse.org` label. Be aware that the Che Operator has to already have all permissions in these ClusterRoles to grant them.
 cheDebug: Enables the debug mode for Che server. Defaults to `false`.
-cheFlavor: Deprecated. The value of this flag is ignored. Specifies a variation of the installation. The options are `che` for upstream Che installations, or `devspaces` for link\:https\://developers.redhat.com/products/codeready-workspaces/overview[Red Hat OpenShift Dev Spaces] installation. Override the default value only on necessary occasions.
+cheFlavor: Deprecated. The value of this flag is ignored. Specifies a variation of the installation. The options are  `che` for upstream Che installations or `devspaces` for Red Hat OpenShift Dev Spaces (formerly Red Hat CodeReady Workspaces) installation
```
themr0c, line 14: Don't commit this file as part of this pull request. The file is single-sourced from the code repository and the change belongs to a dedicated pull request.

PR #2352, opened 2022-06-07
procedures: artifact repos

max-cx:
The SME collaborator is @ScrewTSW

Adding this topic to the 7.46.x docs.

RHDEVDOCS-4120

7.46.x

## Pull request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:


max-cx:
Current previews:
image

[Enabling Maven artifact repositories Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8855539/Enabling.Maven.artifact.repositories.Eclipse.Che.Documentation.pdf)
[Enabling Gradle artifact repositories Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8855538/Enabling.Gradle.artifact.repositories.Eclipse.Che.Documentation.pdf)
[Enabling npm artifact repositories Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8855540/Enabling.npm.artifact.repositories.Eclipse.Che.Documentation.pdf)
[Enabling Python artifact repositories Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8855543/Enabling.Python.artifact.repositories.Eclipse.Che.Documentation.pdf)
[Enabling Go artifact repositories Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8855537/Enabling.Go.artifact.repositories.Eclipse.Che.Documentation.pdf)
[Enabling NuGet artifact repositories Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8855542/Enabling.NuGet.artifact.repositories.Eclipse.Che.Documentation.pdf)

modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
```diff
@@ -0,0 +1,108 @@
...
+:page-aliases: .:using-maven-artifact-repositories
+
+[id="enabling-maven-artifact-repositories"]
+= Enabling Maven artifact repositories
+
+To enable a Maven artifact repository in Maven workspaces that run in a restricted environment:
+
+.Prerequisites
+
+* You are not running any Maven workspace.
```
nickboldt, line 12: Should this be in the form of a directive, like:
* Ensure all maven based workspaces are stopped.

modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
```diff
@@ -0,0 +1,108 @@
...
+  labels:
+    controller.devfile.io/mount-to-devworkspace: 'true'
+    controller.devfile.io/watch-secret: 'true'
+data:
+  tls.cer: >-
+    __<Base64_encoded_content_of_public_cert>__ <1>
+----
+<1> Base64 encoding with disabled line wrapping.
+
+. Apply the ConfigMap to create the `settings.xml` file:
```
nickboldt, line 36: You have 6 examples with "Apply the ConfigMap to ..." and 4 examples with "Apply ... as a ConfigMap."

nickboldt (approved):
+1, but would be nice if the 10 places where a configmap is used are all PHRASED the same way, rather than the 2 different ways.

nickboldt (approved):

modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
max-cx: I believe this is resolved now.

modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
max-cx: `You are not running any workspaces` is a condition, whereas we put directives into the procedure.
It's also possible that the reader has no workspaces that must be stopped.

rkratky: I'd go with Nick's version here because there's no guarantee that it's the reader/user (i.e. "you") who's running those workspaces. A general condition ("no Maven WSs running" or "all Maven WSs stopped") seems more universal than saying "You are not running..."

max-cx: Yes, but this is the end-user guide, so the reader is the one performing the procedure and to whom the prerequisite applies.

max-cx: I'll check with Tibor.

modules/end-user-guide/pages/enabling-go-artifact-repositories.adoc
```diff
@@ -0,0 +1,55 @@
+:navtitle: Go
+:keywords: go, artifact-repository, artifact-repositories
+:page-aliases: .:using-go-artifact-repositories
+
+[id="enabling-go-artifact-repositories"]
+= Enabling Go artifact repositories
+
+To enable a Go artifact repository in Go workspaces that run in a restricted environment:
```
rkratky: I believe the IBM SG rule about lead-in sentences ending with a colon doesn't apply in this case because

a) This is (or should be) a context paragraph, not a lead-in sentence.
b) Even if we considered it a lead-in sentence, there's still another section (prerequisites) between the sentence and the actual procedure steps, so using a colon is out of place.

modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
max-cx: Discussed and resolved on Slack.

modules/end-user-guide/pages/enabling-go-artifact-repositories.adoc
max-cx: Thank you, Robert.
I've rewritten it for all procedures in this PR as a proper sentence that ends with a period.

rkratky (approved):

max-cx:
> +1, but would be nice if the 10 places where a configmap is used are all PHRASED the same way, rather than the 2 different ways.

@nickboldt, i fixed it everywhere yesterday, as discussed.

modules/end-user-guide/pages/enabling-artifact-repositories-in-a-restricted-environment.adoc
```diff
@@ -0,0 +1,17 @@
+:_content-type: ASSEMBLY
+:navtitle: Enabling artifact repositories in a restricted environment
+:description: Enabling artifact repositories in a restricted environment
+:keywords: artifact-repositories, artifact-repository, maven, gradle, nuget, python, go, npm
+:page-aliases: .:using-artifact-repositories-in-a-restricted-environment
+
+[id="enabling-artifact-repositories-in-a-restricted-environment_{context}"]
+= Enabling artifact repositories in a restricted environment
+
+By configuring technology stacks, you can work with artifacts from in-house repositories using self-signed certificates:
```
deerskindoll, line 10: (...) using the following self-signed certificates

this would make the sentence flow better imho

modules/end-user-guide/pages/enabling-go-artifact-repositories.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+.Prerequisites
+
+* You are not running any Go workspace.
+
+include::partial$snip_warning-about-applying-a-configmap-that-sets-environment-variables.adoc[]
+
+.Procedure
+
+. Apply the Secret for the TLS certificate:
```
deerskindoll, line 18: Secret or secret? I'm not sure the capital S is necessary here (same for the other occurrences below).

deerskindoll (commented):
peer review: mini-suggestions

modules/end-user-guide/pages/enabling-artifact-repositories-in-a-restricted-environment.adoc
max-cx: Thanks, Jana. I avoided `the following` because i aimed at a general meaning that applies to the whole class of such repositories in general, not just those listed below. Also, users will only need one or two from the list, so i wouldn't want to tell readers to use the following repositories.

modules/end-user-guide/pages/enabling-go-artifact-repositories.adoc
max-cx: The capitalized version is still found in the OpenShift docs, and it's the convention in the Kubernetes docs.

PR #2365, opened 2022-06-09
fix: rewrite the section about migrating from Che 7.41

max-cx:

See the changes.

Preview: [Upgrading Che v7.41 on OpenShift Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/8870257/Upgrading.Che.v7.41.on.OpenShift.Eclipse.Che.Documentation.pdf)

RHDEVDOCS-4132

7.46.x

## Pull request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
+++ b/modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
@@ -11,8 +11,8 @@ The workspace engine and authentication system used up to and including {prod-pr
 
 .Prerequisites
 
-* An instance of {prod-prev-short} deployed on one of the xref:supported-platforms.adoc[], using the default internal PostgreSQL database, and with OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
-* The following CLI tools are available:
+* An instance of {prod-prev-short} deployed on one of the xref:supported-platforms.adoc[]. The instance uses the default internal PostgreSQL database and has OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
+* The following command line tools are available:
 ** `oc`
 ** `curl`
 ** `jq`
@@ -62,15 +62,18 @@ export PRE_MIGRATION_PRODUCT_SUBSCRIPTION_NAME={pre-migration-prod-subscription}
 <1> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
 <2> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
 
+. Run `chmod` on the downloaded scripts:
++
+[source,terminal]
+----
+$ chmod +x ./1-prepare.sh ./2-migrate.sh ./3-subscribe.sh ./4-wait.sh
+----
+
 . Run the migration scripts:
 +
 [source,terminal]
 ----
-$ chmod +x ./1-prepare.sh ./2-migrate.sh ./3-subscribe.sh ./4-wait.sh /
-./1-prepare.sh /
-./2-migrate.sh /
-./3-subscribe.sh /
-./4-wait.sh
+$ ./1-prepare.sh && ./2-migrate.sh && ./3-subscribe.sh && ./4-wait.sh
 ----
 
 .Verification
@@ -101,13 +104,22 @@ export PRE_MIGRATION_PRODUCT_OLM_CATALOG_SOURCE={pre-migration-prod-catalog-sour
 <1> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
 <2> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
 
-. Download and run the xref:attachment$migration/rollback.sh[rollback.sh] script.
+. Download the xref:attachment$migration/rollback.sh[rollback.sh] script.
+
+. Run `chmod` on the `rollback.sh` script:
 +
 [source,terminal]
 ----
-$ chmod +x ./rollback.sh /
-./rollback.sh
+$ chmod +x ./rollback.sh
 ----
+
+. Run the `rollback.sh` script.
++
+[source,terminal]
+----
+$ ./rollback.sh
+----
+
 
 .Verification
 
--- a/modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
+++ b/modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
@@ -11,26 +11,47 @@ The workspace engine and authentication system used up to and including {prod-pr
 
 .Prerequisites
 
-* An instance of {prod-prev-short} deployed on xref:supported-platforms.adoc[], using the default internal PostgreSQL database, and with OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
-* Command line tools `oc`, `curl`, and `jq` are available.
+* An instance of {prod-prev-short} deployed on one of the xref:supported-platforms.adoc[], using the default internal PostgreSQL database, and with OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
+* The following CLI tools are available:
+** `oc`
+** `curl`
+** `jq`
 * The host running the upgrade commands is running on Linux.
+* Optional:
+.. All changes from all workspaces have been commited and pushed to their Git remotes.
+.. All workspaces have been stopped to avoid UX degradation.
+.. The {prod-prev-short} data have been backed up. See {prod-docs-url-backup-recovery}.
 
 .Procedure
 
-. Optionally push back changes to Git repositories in all workspaces, stop all workspaces to avoid user experience degradation, and backup {prod-short} data. See {prod-docs-url-backup-recovery}.
+. Download xref:attachment$migration/1-prepare.sh[1-prepare.sh].
++
+`1-prepare.sh` is a migration script that shuts down {prod-prev-short} and {identity-provider}, fetches users data, and dumps the {prod-short} database.
+
+. Download xref:attachment$migration/2-migrate.sh[2-migrate.sh].
++
+`2-migrate.sh` is a migration script that fetches {prod-prev-short} {identity-provider} and database data, and repopulates the database with updated data.
+
+. Download xref:attachment$migration/3-subscribe.sh[3-subscribe.sh].
++
+`3-subscribe.sh` is a migration script that deletes {prod-prev-short} Operator and {identity-provider} resources, updates the CheCluster CR, and creates a new {prod-short} Operator subscription.
+
+. Download xref:attachment$migration/4-wait.sh[4-wait.sh].
++
+`4-wait.sh` is a migration script that waits until {prod-short} is ready, which can take more than 5 minutes.
 
 . Set the environment variables to use in the migration scripts:
 +
 [source,bash,subs="+attributes"]
 ----
-export INSTALLATION_NAMESPACE={prod-namespace} #<1>
+export INSTALLATION_NAMESPACE={prod-namespace} # <1>
 export PRODUCT_ID={prod-id}
 export PRODUCT_DEPLOYMENT_NAME={prod-deployment}
 export PRODUCT_OPERATOR_NAME={prod-operator}
 export PRODUCT_OLM_STABLE_CHANNEL={prod-stable-channel}
 export PRODUCT_OLM_CATALOG_SOURCE={prod-stable-channel-catalog-source}
 export PRODUCT_OLM_PACKAGE={prod-stable-channel-package}
-export PRE_MIGRATION_PRODUCT_OPERATOR_NAMESPACE={prod-namespace}  #<1>
+export PRE_MIGRATION_PRODUCT_OPERATOR_NAMESPACE={prod-namespace} # <2>
 export PRE_MIGRATION_PRODUCT_SHORT_ID={pre-migration-prod-id-short}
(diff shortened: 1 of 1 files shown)


tolusha (approved):

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,85 @@ The workspace engine and authentication system used up to and including {prod-pr
 
 .Prerequisites
 
-* An instance of {prod-prev-short} deployed on xref:supported-platforms.adoc[], using the default internal PostgreSQL database, and with OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
-* Command line tools `oc`, `curl`, and `jq` are available.
+* An instance of {prod-prev-short} deployed on one of the xref:supported-platforms.adoc[], using the default internal PostgreSQL database, and with OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
```
rolfedh:
```suggestion
* An instance of {prod-prev-short} running on one of the xref:supported-platforms.adoc[]. This instance uses the default internal PostgreSQL database and has OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}.
```

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,85 @@ The workspace engine and authentication system used up to and including {prod-pr
...
+* The following CLI tools are available:
```
rolfedh:
```suggestion
* The following command line tools are available:
```

max-cx: Thank you, Rolfe.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,85 @@ The workspace engine and authentication system used up to and including {prod-pr
...
+** `oc`
+** `curl`
+** `jq`
 * The host running the upgrade commands is running on Linux.
+* Optional:
+.. All changes from all workspaces have been commited and pushed to their Git remotes.
+.. All workspaces have been stopped to avoid UX degradation.
+.. The {prod-prev-short} data have been backed up. See {prod-docs-url-backup-recovery}.
```
rolfedh, line 23: - Would these be better as first steps in the procedure? 
- Instead of saying "optional," you can preface the step with a condition (precondition). That way, the user can decide if the condition applies to their situation.  
- Some steps, like backing up data, are so common they might not need the precondition.

For example:

```
. To avoid losing work, commit and push changes from all workspaces to their Git remotes.
. To avoid impacting users' experiences, stop all workspaces.
. Back up the {prod-prev-short} data. See {prod-docs-url-backup-recovery}.
```

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: @rolfedh, i applied it with a couple of minor tweaks:
```
* An instance of {prod-prev-short} deployed on one of the xref:supported-platforms.adoc[]. The instance uses the default internal PostgreSQL database and has OpenShift OAuth enabled. See {prod-docs-url-enable-oauth}. 
```

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,85 @@ The workspace engine and authentication system used up to and including {prod-pr
...
 
 .Procedure
 
-. Optionally push back changes to Git repositories in all workspaces, stop all workspaces to avoid user experience degradation, and backup {prod-short} data. See {prod-docs-url-backup-recovery}.
+. Download xref:attachment$migration/1-prepare.sh[1-prepare.sh].
```
rolfedh, line 27: Would it work better to batch these four steps into one. For example:
```suggestion
. Download the following migration scripts:
** xref:attachment$migration/1-prepare.sh[1-prepare.sh]: Shuts down {prod-prev-short} and {identity-provider}, fetches users data, and dumps the {prod-short} database.
** etc...
```

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,85 @@ The workspace engine and authentication system used up to and including {prod-pr
...
 export PRE_MIGRATION_PRODUCT_SHORT_ID={pre-migration-prod-id-short}
 export PRE_MIGRATION_PRODUCT_DEPLOYMENT_NAME={pre-migration-prod-deployment}
 export PRE_MIGRATION_PRODUCT_OPERATOR_NAME={pre-migration-prod-operator}
 export PRE_MIGRATION_PRODUCT_CHE_CLUSTER_CR_NAME={pre-migration-prod-checluster}
 export PRE_MIGRATION_PRODUCT_IDENTITY_PROVIDER_DEPLOYMENT_NAME={identity-provider-id}
 export PRE_MIGRATION_PRODUCT_SUBSCRIPTION_NAME={pre-migration-prod-subscription}
 ----
-<1> Adapt to the {orch-namespace} where {prod-short} is installed.
+<1> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
+<2> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
```
rolfedh, line 63: I'm not sure how these attributes render. Do they sound more like normal speech if you precede them with "The"? For example:
```suggestion
<1> The `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
<2> The `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
```

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,88 @@ The workspace engine and authentication system used up to and including {prod-pr
...
+. Download xref:attachment$migration/4-wait.sh[4-wait.sh].
++
+`4-wait.sh` is a migration script that waits until {prod-short} is ready, which can take more than 5 minutes.
 
 . Set the environment variables to use in the migration scripts:
 +
 [source,bash,subs="+attributes"]
 ----
-export INSTALLATION_NAMESPACE={prod-namespace} #<1>
+export INSTALLATION_NAMESPACE={prod-namespace} # <1>
```
rolfedh, line 47: Would it help to make the use or absence of command prompts consistent in this file? Or is this an exception because we expect the user to copy the whole block of commands? (For example, to edit the values in a text file, and then paste them to a terminal?)

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,88 @@ The workspace engine and authentication system used up to and including {prod-pr
...
+[source,terminal]
+----
+$ ./1-prepare.sh && ./2-migrate.sh && ./3-subscribe.sh && ./4-wait.sh
 ----
-<1> Shuts down {prod-prev-short} and {identity-provider}, fetch users data and dumps {prod-short} database.
-<2> Fetches {prod-prev-short} {identity-provider} and database data, and repopulates the database with updated data.
-<3> Deletes {prod-prev-short} Operator and {identity-provider} resources, updates CheCluster CR and creates a new {prod-short} Operator subscription.
-<4> Waits until {prod-short} is ready (it can take more than 5 minutes).
 
 .Verification
```
rolfedh, line 79: This seems to duplicate the `.Verification` on line 124.

rolfedh: Combine?

rolfedh: Or is that a verification of the rollback procedure?

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,88 @@ The workspace engine and authentication system used up to and including {prod-pr
...
 
 * Open {prod-short} dashboard and log in.
```
rolfedh, line 81: What should the user expect to see?

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -11,67 +11,88 @@ The workspace engine and authentication system used up to and including {prod-pr
...
 
 [IMPORTANT]
 ====
-If migration fails, use the following steps to roll back to the previous {prod-prev-short} installation.
+If migration fails, use the following steps to roll back to the pre-migration {prod-prev-short} state.
```
rolfedh, line 85: Please move this procedure to a separate module and link to it.
Admonitions should be short, concise, and not contain procedures.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: Rolfe, i like the wording in your suggestion. However, the third/last step would only apply to those admins who are already using the referenced functionality, which means it's definitively optional.

These steps were originally, as you can see in the diff, in the procedure as one first step, listed as a series in one sentence.
However, i'd like to avoid lumping all non-migration tasks and migration tasks into one procedure.
So I figured it'd make things clearer if i moved those into the prerequisites.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
rolfedh: In our Slack thread, we discussed this and decided it might be better to not use command prompts in this instance. Instead, we should tell the user to paste the block of commands into a text editor and insert the correct values before pasting the block of commands into a terminal.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: Also, i'm not worried about the fact that with this layout and these extra points the list of prerequisites takes more lines that readers are used to seeing in Prerequisites. This is a very important procedure, so it's something that has to be approached with extra care anyway.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
```diff
@@ -80,16 +101,26 @@ export PRE_MIGRATION_PRODUCT_OLM_PACKAGE={pre-migration-prod-package}
...
 
-. Download and execute the rollback script: xref:attachment$migration/rollback.sh[rollback.sh].
+. Run `chmod` on the `rollback.sh` script:
 +
-[source,bash,subs="+attributes"]
+[source,terminal]
 ----
-chmod +x ./rollback.sh
-./rollback.sh
+$ chmod +x ./rollback.sh
```
nickboldt, line 113: seems overkill to split this into 2 steps, but :shrug:

rolfedh (approved):
/ltgm

rolfedh:
Pre-approved this PR to expedite merging.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: Thanks, Rolfe, but this looks too crowded to me:

image

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
rolfedh: @max-cx Sure, keeping it in the prerequisites is fine. We still need to inform the user of the basis for their choice. Stating that it's optional isn't enough.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: This is how they're rendered:
image

I'd like to avoid `The` there, and I'm aware of the styleguide guidance to start a sentence with some capitalized word, but in this case adding `The` makes it more confusing in my opinion.

It's either (a) `eclipse-che` or (b) _another namespace_.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
rolfedh: It sounds like you should provide the user with a precondition for the third/last prerequisite.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: Yes, that is a verification of the rollback procedure.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: Thank you, Rolfe. I'll open an issue for this to be dealt with after the release.

modules/administration-guide/pages/upgrading-che-7-41-on-openshift.adoc
max-cx: Adding this question to the new issue.

PR #2368, opened 2022-06-10
fix: Update GitLab configuration page

vinokurig:
Signed-off-by: Igor Vinokur <ivinokur@redhat.com>

Update GitLab oAuth configuration page.
screenshot-0 0 0 0_4000-2022 06 10-09_27_42

fixes https://github.com/eclipse/che/issues/21435

## Pull request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/partials/proc_applying-the-gitlab-authorized-application-secret.adoc
+++ b/modules/administration-guide/partials/proc_applying-the-gitlab-authorized-application-secret.adoc
@@ -42,7 +42,7 @@ data:
   secret: __<Base64_GitLab_Client_Secret>__ <4>
 ----
 <1> The {prod-short} namespace. The default is `{prod-namespace}`.
-<2> The *GitLab server URL*.
+<2> The *GitLab server URL*. Use `https​://gitlab.com` for the `SAAS` version.
 <3> The Base64-encoded *GitLab Application ID*.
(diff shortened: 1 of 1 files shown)


themr0c:
Is the feature for Che 7.50, or does it work already on Che 7.46?

themr0c:
In the corresponding issue I see:

> We can configure only SAAS or on-premises version with the current oAuth provider.

I believe we should add this information in the docs as well. Something like:

> To enable users to work with a remote Git repository that is hosted using one GitLab instance:

vinokurig:
> Is the feature for Che 7.50, or does it work already on Che 7.46?

It already works since the gitlab oAuth is added

amisevsk (approved):

vinokurig:
@themr0c
> In the corresponding issue I see:
>
> >We can configure only SAAS or on-premises version with the current oAuth provider.
>
>I believe we should add this information in the docs as well. Something like:
>
> >To enable users to work with a remote Git repository that is hosted using one GitLab instance:

I think from this PR it is clear what to do if admin wants to use the SAAS **OR** onpremises version. I don't think it is a good idea to describe that we doesn't support using both instances as it would only confue the reader.

themr0c (approved):

PR #2369, opened 2022-06-10
fix: Add startingCSV to subscriptions

tolusha:
Signed-off-by: Anatolii Bazko <abazko@redhat.com>

Downstream values:
```
pre-migration-prod-starting-csv: crwoperator.v2.15.4 # (or v2.15.104 if running the Technoogy Preview version)
prod-stable-channel-starting-csv: devspacesoperator.v3.0.0
```

* Get CSV name from `status.installedCSV` field
* Add `startingCSV` field to subscription to avoid migration to the version with CheCluster API v2, in this case rollback won't be possible
* Set `installPlanApproval` to `Manual`
* Install plan approval

https://issues.redhat.com/browse/CRW-3028

## Pull request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -70,6 +70,7 @@ asciidoc:
     pre-migration-prod-package: eclipse-che
     pre-migration-prod-channel: stable
     pre-migration-prod-catalog-source: community-operators
+    pre-migration-prod-starting-csv: eclipse-che.v7.41.2
     prod-checluster: eclipse-che
     prod-cli: chectl
     prod-deployment: che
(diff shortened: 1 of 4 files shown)


nickboldt:
Note that if a new CVE fix rolls out via freshmaker, `pre-migration-prod-starting-csv: crwoperator.v2.15.4-0.1652874390.p` will no longer be the most recent release.

And for airgapped customers who haven't fetched these updates, that version won't exist in their clusters.

So you probably want just this:

`pre-migration-prod-starting-csv: crwoperator.v2.15.4`

But also... does the migration work for 2.15.104 (tech preview) as well as 2.15.4 (GA, no DWO) ?

nickboldt (approved):
+1, need this so we can enforce that migration is ONLY to 3.0, and after that we use OLM to handle updates through 3.x

themr0c (approved):

PR #2385, opened 2022-06-28
fix: take the rollback procedure out of the admonition into a separate module

max-cx:
1. Take the rollback procedure out of the admonition into a separate module to make it comply with the modular docs framework.
2. Improve the verification step in response to Rolfe's feedback.
3. Make minor improvements to the abstract sentences and some steps.

RHDEVDOCS-4135

This PR needs to be backported to `7.46.x` and `7.48.x`.
For `7.50.x` and onward, see https://github.com/eclipse-che/che-docs/pull/2389.

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


modules/administration-guide/partials/proc_rolling-the-upgrade-back-to-che-7-41-on-openshift.adoc
```diff
@@ -0,0 +1,51 @@
...
+export PRE_MIGRATION_PRODUCT_OLM_STARTING_CSV={pre-migration-prod-starting-csv}
+----
+<1> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
+<2> `{prod-namespace}` or another {orch-namespace} where {prod-prev-short} was previously installed.
+
+. Run `chmod` on the `rollback.sh` script:
++
+[source,terminal]
+----
+$ chmod +x ./rollback.sh
```
nickboldt: personally, I'd combine these two steps into a single command, because anyone running linux-based shell commands doesn't need to be told to run chmod (and the instructions don't explain what chmod is, only that you should run it ... so you're not educating anyone on WHY the file needs to be executable).

Thus:

```
. Run the `rollback.sh` script.
+
[source,terminal]
----
$ chmod +x ./rollback.sh; ./rollback.sh
----
```

nickboldt:
most of this PR is about grafana and prometheus... any chance you got an extra commit leaked into this? Or does the PR title need to be updated to reflect the wider scope of changes?

modules/administration-guide/partials/proc_rolling-the-upgrade-back-to-che-7-41-on-openshift.adoc
max-cx: Thank you, Nick. I've applied this suggestion to both procedures now in a new commit.

max-cx:
> most of this PR is about grafana and prometheus... any chance you got an extra commit leaked into this? Or does the PR title need to be updated to reflect the wider scope of changes?

Yes, I should have used `git push -f` after rebasing on `upstream master`.
Thanks for spotting it.
Fixed now.

max-cx:
Preview:
[Upgrading Che v7.41 on OpenShift Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/9010781/Upgrading.Che.v7.41.on.OpenShift.Eclipse.Che.Documentation.pdf)

nickboldt:
Why does the PDF have version `Che 7.47` ? 

but also 
```
 PRODUCT_OLM_STARTING_CSV=eclipse-che.v7.49.0
```

@tolusha are we supporting 7.41 -> 7.47? or 7.49? or any arbitrary 7.yy? 

Asking because in DS, we only support migration from 2.15.z (7.42) to 3.0 (7.46), with no other migration options. 

For upstream I would expect the migration would be from the same versions... 7.41 or 7.42 -> 7.46, then use OLM to do subsequent updates (not migrations).

max-cx:
`Che 7.47` is `{prod-short} {prod-ver}` according to https://github.com/eclipse-che/che-docs/blob/main/antora.yml.

BTW, I wanted to double-check: because of the upgrade path, we will discard this section for DS 3.1, correct?

nickboldt:
> Che 7.47 is {prod-short} {prod-ver} according to https://github.com/eclipse-che/che-docs/blob/main/antora.yml.

then why do you ALSO have 7.49 in the document? 

image

And no, we need a migration guide for downstream too, but it's for 2.15 -> 3.0 only (not for 2.15 -> 3.x).

https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.0/html-single/administration_guide/index#upgrading-crw-2.15-on-openshift

max-cx:
@nickboldt, i'll have to check with @themr0c if we need to update https://github.com/eclipse-che/che-docs/blob/main/antora.yml.
Currently, lines 101 and 102 read
```
    prod-ver-patch: "7.47.0"
    prod-ver: "7.47"
```

max-cx:
@nickboldt, the version issue is now fixed as part of this PR: 

```
    prod-ver-patch: "7.49.0"
    prod-ver: "7.49"
    prod-last-version-pre-dwo: 7.41
```
New preview: 
[Upgrading Che 7.41 on OpenShift Eclipse Che Documentation.pdf](https://github.com/eclipse-che/che-docs/files/9011488/Upgrading.Che.7.41.on.OpenShift.Eclipse.Che.Documentation.pdf)

nickboldt:
@max-cx ok but you've missed the second half of my comment in https://github.com/eclipse-che/che-docs/pull/2385#issuecomment-1169965054 ... I am surprised to see we're encouraging users to skip from 7.41 to 7.49 (latest release), since for Dev Spaces the decision was to NOT support 2.15 -> 3.latest; instead the migration is ONLY for 2.15 -> 3.0.0. (Then for 3.0 -> 3.1 -> 3.2 -> 3.3 -> ... we would use OLM.)

So you're proposing a different support model and upgrade path in upstream than in downstream. I think that's a mistake.

I would expect that the upgrade path would be from:

* 7.41 (or 7.42) to 7.46. 

After that, migration is NOT supported; instead upgrades will be handled by OLM.

max-cx:
@nickboldt, good question, and I need a decision by an SME on this.
Please keep in mind that we can only document one product version per docs version.
If the SMEs decide to keep it strictly 7.46, then we have to drop it from the docs starting with 7.47 because we can't tell users to do something with 7.46 in the docs for 7.47+.

themr0c:
Then we can remove the upgrade procedure from main. But as we don't keep older releases published, the upgrade procedure will disappear from publication. Or we should keep 7.46.x published.

jc-berger (approved):
/lgtm

max-cx:
@nickboldt, @rkratky, and @themr0c, I propose a solution to this `prod-ver` confusion:
I take this migration section out of the docs for Che 7.46 and DS 3.0 and incorporate it instead into the docs for Che 7.41 and CRW 2.15 where it belongs because it is about migrating the previous version, the version that customers have installed and need to migrate. There we will keep the target version as 7.46 and 3.0 respectively. WDYT?
And for Che 7.46 and DS 3.0, we create a RN with a link to that location in the 7.41/2.15 docs.
(An alternative to this would be to start versioning the upstream docs by version number, rather than `stable` and `next`.)

nickboldt:
> I take this migration section out of the docs for Che 7.46 and DS 3.0 and incorporate it instead into the docs for Che 7.41 and CRW 2.15

No, it's already published in the DS 3.0 docs and we've been linking to it in various places. Don't break the 3.0 docs; just fix the version target in the 7.yy docs to say "you can migrate from 7.41 to 7.46, then use OLM for all updates after that."

max-cx:
Here's my latest proposal:

1. We publish this procedure in the `next` (`main`) docs for `7.49.x` and keep it in the DS 3.0 docs.
2. We also backport this update into the `stable` (`7.48.x`) and the `7.46.x` branches.
3. We then remove this procedure from the `7.50.x` docs, which means this procedure will then be absent from the DS 3.1 docs.
(This procedure will remain available in the DS 3.0 docs because the DS 3.0 docs will remain online.)
4. Subject to discussion: We add a RN for Che 7.50, quoting Nick: _"you can migrate from 7.41 to 7.46, then use OLM for all updates after that."_
5. Subject to discussion: We add a RN for DS 3.1 regarding the correct upgrade path (CRW 2.15 > DS 3.0 > DS 3.1) with a link to this procedure in the DS 3.0 docs.

nickboldt:
> We then remove this procedure from the 7.50.x docs, which means this procedure will then be absent from the DS 3.1 docs.
> (This procedure will remain available in the DS 3.0 docs because the DS 3.0 docs will remain online.)

This is OK as long as there's a section in the 3.1 docs (admin guide or KI/RN) that explains that if you're updating from 2.15.z, you have to migrate first [with link to the 3.0 docs], then use OLM to update from 3.0 -> 3.1. 

> Subject to discussion: We add a RN for DS 3.1 regarding the correct upgrade path (CRW 2.15 > DS 3.0 > DS 3.1) with a link to this procedure in the DS 3.0 docs.

I think we need this in place for 3.1, 3.2, and 3.future, so that anyone arriving at 3.latest docs knows that they need to refer to the 3.0 docs for migration, then can use OLM for subsequent updates.

As to the equivalent changes for upstream Che... whatever you think is best, but mirroring the downstream approach is likely simpler.

max-cx:
Thank you, everyone, for reviewing this PR.
I have to close it as I'm replacing it with two new PRs:
https://github.com/eclipse-che/che-docs/pull/2391/
https://github.com/eclipse-che/che-docs/pull/2393
I believe I have incorporated all your feedback above into these two PRs and their downstream MR counterparts.

PR #2388, opened 2022-07-04
fix: pem 'trusted certificate' update

deerskindoll:
Update of the "Importing untrusted TLS certificates to Che doc"

https://issues.redhat.com/browse/RHDEVDOCS-3370

3.1

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -34,7 +34,7 @@ If the cluster is configured with an intermediate certificate, then the whole ch
 
 == Adding new CA certificates into {prod-short}
 
-Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.crt`, `.ca-bundle`. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
+Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle`and others. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
 
 .Prerequisites
 
@@ -50,7 +50,7 @@ Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.
 +
 [CAUTION]
 ====
-* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
+* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported. Convert it to the supported `CERTIFICATE` format with the following command:
 ** `openssl x509 -in cert.pem -out cert.cer`
 ====
 
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -34,7 +34,7 @@ If the cluster is configured with an intermediate certificate, then the whole ch
 
 == Adding new CA certificates into {prod-short}
 
-Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle` and others. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
+Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle`, and others. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded certificate. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
 
 .Prerequisites
 
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -48,7 +48,7 @@ The following procedure is applicable for already installed and running instance
 [CAUTION]
 ====
 * Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.crt`, `.ca-bundle`. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate.
-* A certificate with the introductory phrase `BEGIN TRUSTED CERTIFICATE` is likely in the PEM `TRUSTED CERTIFICATE` format, which is not not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
+* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
 ** `openssl x509 -in cert.pem -out cert.cer`
 * {prod-short} already uses some reserved file names to automatically inject certificates into the ConfigMap, so you should avoid using the following reserved file names to save your certificates:
   ** `ca-bundle.crt`
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -40,7 +40,7 @@ Certificate files are typically stored as Base64 files, with extensions such as 
 
 * An active `{orch-cli}` session with administrative permissions to the destination {orch-name} cluster. See {orch-cli-link}.
 * Namespace for {prod-short} exists.
-* {prod-short} already uses some reserved file names to automatically inject certificates into the ConfigMap, so you should avoid using the following reserved file names to save your certificates:
+* {prod-short} already uses some reserved file names to automatically inject certificates into the ConfigMap, so avoid using the following reserved file names to save your certificates:
   ** `ca-bundle.crt`
   ** `ca.crt`
 
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -34,7 +34,7 @@ If the cluster is configured with an intermediate certificate, then the whole ch
 
 == Adding new CA certificates into {prod-short}
 
-Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle` and others. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
+Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle` and others. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
 
 .Prerequisites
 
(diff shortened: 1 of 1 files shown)


modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -47,7 +47,9 @@ The following procedure is applicable for already installed and running instance
 +
 [CAUTION]
 ====
-* Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.crt`, `.ca-bundle`. But, they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate.
+* Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.crt`, `.ca-bundle`. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate.
```
Srivaralakshmi:
```suggestion
* Certificate files are typically stored as Base64 ASCII files, such as files of the `.pem`, `.crt`, and `.ca-bundle` formats. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate.
```
Question: 
- Are there more types of file formats other than the ones mentioned here?

Suggestion:
- Can we say: "All Secrets that hold certificate files must use the Base64 ASCII certificate."?

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -47,7 +47,9 @@ The following procedure is applicable for already installed and running instance
...
+* A certificate with the introductory phrase `BEGIN TRUSTED CERTIFICATE` is likely in the PEM `TRUSTED CERTIFICATE` format, which is not not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
```
Srivaralakshmi:
```suggestion
* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
```

Srivaralakshmi: Observation and suggestion:

- Admonitions should be short and concise. Refer to https://redhat-documentation.github.io/supplementary-style-guide/#admonitions and do the needful.

Srivaralakshmi (changes requested):
@deerskindoll Left some suggestions. PTAL. Thanks!

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
deerskindoll: @themr0c @svor can you please help me with answering the question above? thanks beforehand!

Srivaralakshmi: @deerskindoll If there are only 3 formats, then say: "`.pem`, `.crt`, and `.ca-bundle` formats."

If there are more than 3, then say: "in formats such as `.pem`, `.crt`, `.ca-bundle`, and so on."

themr0c: All that can be summarized in a simple minimalistic step:

* Save the CA certificates to the DER format in the <custom-ca.cer> file. 

Example: Convert a certificate from the PEM format to the DER format

```
openssl x509 -in cert.pem  -out custom-ca.cer -outform DER
```

Additional resources
* DER format in the Section 5.1 of RFC 7468: https://datatracker.ietf.org/doc/html/rfc7468#section-5.1
* OpenSSL certificate display and signing utility manual page: https://www.openssl.org/docs/man1.0.2/man1/x509.html


----

Side notes:

* The file names are irrelevant.
* "Removing the header and footer lines" exactly means converting from PEM to DER.

We don't need to paste details about TLS certificates formats in the docs, it belongs to the standard knowledge of our persona, the Administrator. It's good knowledge for the writers:

The DER format is the DER encoding of the certificate and PEM is the base64 encoding of the DER encoding with header and footer lines added. The NET option is an obscure Netscape server format that is now obsolete. See: https://www.openssl.org/docs/man1.0.2/man1/x509.html

themr0c: Now I am reading it again and I am not sure. This requires testing: PEM = DER + base64 encoded + header + footer. So we need something inbetween for Java? PEM without header and footer?

themr0c: Diving deeper:

https://wiki.openssl.org/index.php/DER -- DER is a binary format for data structures described by ASN.1. For example, x509 is described in ASN1 and encoded in DER. It exists other encoding formats for ASN.1 but DER is the one chosen for security since there is only one possible encoding given a ASN.1. encoding (which is not the case for BER used in ldap, for example). 

https://wiki.openssl.org/index.php/PEM -- Privacy Enhanced Mail (PEM) is a format with goal to embed binary content into a content typed 7bits ASCII format. it looks like a HEADER line a binary content encoded in [Base64](https://wiki.openssl.org/index.php/Base64) cut by lines of max 64 characters and a TAIL line. content itself is often [DER](https://wiki.openssl.org/index.php/DER) encoded. Then PEM is an additionnal layer to easy cut and paste in mails/texts of binary content.

I believe the actual content is misleading, and what we need is really the certificate in DER format.

tolusha: We don't support certificates in DER format
https://github.com/eclipse/che/issues/18352

themr0c: Let's clarify: these are common formats:
* DER format: binary format.
* PEM format: the base64 encoding of the DER encoding with header and footer lines added. 

The various inputs are conflicting. Can we clarify the required format for the certificate. Not DER seems acquired. But help choosing between:

* PEM. (See: https://github.com/eclipse/che/issues/18352)
* A custom format: PEM, with header and footer lines removed. Equivalent to: the base64 encoding of the DER encoding. (See: actual docs content). 

Then we have also to adapt the openssl command (add a pipe to remove the headers if required). For the custom format, we have to choose the most useful command: "DER | base64" or "PEM | remove headers". (And possibly, be consistent in all the docs, which would deserve a new issue and pull request).

tolusha: To be clear, we support certificates in PEM format only.

deerskindoll: Hi all, I updated the wording of the disputed line. Please check the updated contents and let me know if they're correct or need additional edits. Thanks!

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -34,23 +34,24 @@ If the cluster is configured with an intermediate certificate, then the whole ch
 
 == Adding new CA certificates into {prod-short}
 
-The following procedure is applicable for already installed and running instances and for instances that are to be installed.
+Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle`and others. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
```
tolusha:
```suggestion
Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle` and others. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
```

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -34,23 +34,24 @@ If the cluster is configured with an intermediate certificate, then the whole ch
...
 
 . Save the certificates you need to import, to a local file system.
 +
 [CAUTION]
 ====
-* Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.crt`, `.ca-bundle`. But, they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate.
-* {prod-short} already uses some reserved file names to automatically inject certificates into the ConfigMap, so you should avoid using the following reserved file names to save your certificates:
-  ** `ca-bundle.crt`
-  ** `ca.crt`
+* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported. Convert it to the supported `CERTIFICATE` format with the following command:
```
tolusha, line 53: I am confused. Does it contradictory to:

```
Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle`and others. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
```

deerskindoll: As I understand the line, it explains what formats certificates have and which format should be used.

tolusha: I mean, here we mention, that `.cer` extension is used for binary-encoded certificate format which are not supported:

```
 Certificate files are typically stored as Base64 ASCII files, such as `.pem`, `.crt`, `.ca-bundle`. But they can also be binary-encoded, for example, as `.cer` files. All Secrets that hold certificate files should use the Base64 ASCII certificate rather than the binary-encoded certificate.
```

and below, we suggest to convert certificate using file extension that is not supported accordingly to phrase above.

```
openssl x509 -in cert.pem -out cert.cer
```

It sounds inconsistent for me.

deerskindoll: Then let me get rid of this sentence: But they can also be binary-encoded, for example, as `.cer` files.

Would that help?

tolusha: Yes, that's will be better from my point of view.

deerskindoll: Done

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -34,23 +34,24 @@ If the cluster is configured with an intermediate certificate, then the whole ch
 
 == Adding new CA certificates into {prod-short}
 
-The following procedure is applicable for already installed and running instances and for instances that are to be installed.
+Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle` and others. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
```
Srivaralakshmi:
```suggestion
Certificate files are typically stored as Base64 files, with extensions such as `.pem`, `.crt`, `.ca-bundle`, and others. All Secrets that hold certificate files should use the Base64-encoded certificate rather than binary-encoded certificate. The following procedure is applicable for already installed and running instances and for instances that are to be installed.
```

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -34,23 +34,24 @@ If the cluster is configured with an intermediate certificate, then the whole ch
...
 
 .Prerequisites
 
 * An active `{orch-cli}` session with administrative permissions to the destination {orch-name} cluster. See {orch-cli-link}.
 * Namespace for {prod-short} exists.
+* {prod-short} already uses some reserved file names to automatically inject certificates into the ConfigMap, so you should avoid using the following reserved file names to save your certificates:
```
Srivaralakshmi:
```suggestion
* {prod-short} already uses some reserved file names to automatically inject certificates into the ConfigMap, so you should avoid using the following reserved file names to save your certificates:
```
```suggestion
* {prod-short} already uses some reserved file names to automatically inject certificates into the config map, so you should avoid using the following reserved file names to save your certificates:
```

Srivaralakshmi (changes requested):
@deerskindoll PTAL at the suggestions. Thanks!

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
deerskindoll: Kubernetes documentation uses "ConfigMap" spelling: https://kubernetes.io/docs/concepts/configuration/configmap/

Can I keep it?

Srivaralakshmi (approved):
@deerskindoll Thanks for the changes.  Approved!

rkratky (approved):

PR #2383, opened 2022-06-21
chore: Add instructions how to add a manuall PAT secret

vinokurig:
## What does this pull request change?
Add instructions how to add a manuall PAT secret to be able to create factory from private repo if oAuth is not configured.
## What issues does this pull request fix or reference?
https://github.com/eclipse/che/issues/21189
depends on https://github.com/eclipse-che/che-server/pull/313
## Specify the version of the product this pull request applies to
next
## Pull Request checklist

The author and the reviewers validate the content of this pull request with the following checklist, in addition to the [automated tests](code_review_checklist.adoc).

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:
  - [ ] The page contains a redirection for the previous URL.
  - Propagate the URL change in:
    - [ ] Dashboard [default branding data](https://github.com/eclipse-che/che-dashboard/blob/main/packages/dashboard-frontend/src/services/bootstrap/branding.constant.ts)
    - [ ] Chectl [constants.ts](https://github.com/che-incubator/chectl/blob/main/src/constants.ts)
- [ ] Builds on [Eclipse Che hosted by Red Hat](https://workspaces.openshift.com).
- [ ] the *`Validate language on files added or modified`* step reports no vale warnings.

PR #2401, opened 2022-07-15
fix: reviewed and edited https://github.com/eclipse-che/che-docs/pull/2383/

max-cx:
I reviewed and edited https://github.com/eclipse-che/che-docs/pull/2383/.

RHDEVDOCS-4222

7.50+

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


modules/end-user-guide/pages/git-credentials-store.adoc
```diff
@@ -75,9 +77,9 @@ data:
 You can create and apply multiple Git credentials Secrets in your user {orch-namespace}. All of them will be copied into one Secret that will be mounted to the `{devworkspace}` containers. For example, if you set the mount path to `/etc/secret`, then the one Secret with all of your Git credentials will be mounted at `/etc/secret/credentials`. You must set all Git credentials Secrets in your user {orch-namespace} to the same mount path. You can set the mount path to an arbitrary path because the mount path will be automatically set in the Git configuration file configured at `/etc/gitconfig`.
 ====
 
-. Apply the Secret:
+. Apply the Secret.
 +
-[subs="+quotes,+attributes,+macros"]
+[source,terminal,subs="+quotes,+attributes,+macros"]
```
rkratky: This was correct. It's not source code, and it shouldn't have any syntax highlighting. (Macros are not used, so not needed.)
```suggestion
[subs="+quotes,+attributes"]
```

modules/end-user-guide/pages/git-credentials-store.adoc
```diff
@@ -75,9 +77,9 @@ data:
 You can create and apply multiple Git credentials Secrets in your user {orch-namespace}. All of them will be copied into one Secret that will be mounted to the `{devworkspace}` containers. For example, if you set the mount path to `/etc/secret`, then the one Secret with all of your Git credentials will be mounted at `/etc/secret/credentials`. You must set all Git credentials Secrets in your user {orch-namespace} to the same mount path. You can set the mount path to an arbitrary path because the mount path will be automatically set in the Git configuration file configured at `/etc/gitconfig`.
 ====
 
-. Apply the Secret:
+. Apply the Secret.
```
rkratky: Why not a colon? It seems intuitive... Is that some style-guide rule?

modules/end-user-guide/pages/git-provider-access-token.adoc
```diff
@@ -0,0 +1,68 @@
...
+----
++
+<1> Your Che user ID. You can retrieve `__<che-endpoint>__/api/user` to get the Che user data.
+<2> The Git provider name: `github` or `gitlab` or `bitbucket-server`.
+<3> The Git provider URL.
+<4> Your Git provider user ID.
+
+. Apply the Secret.
++
+[source,terminal,subs="+quotes,+attributes,+macros"]
```
rkratky:
```suggestion
[subs="+quotes,+attributes"]
```

modules/end-user-guide/pages/authenticating-yourself-to-a-git-server-from-a-workspace.adoc
```diff
@@ -9,11 +9,13 @@
```
rkratky, line 9: (I can't comment on lines above, so adding here.)

"Authenticating *yourself* ..." is redundant. I suggest removing "yourself" from it.

modules/end-user-guide/pages/authenticating-yourself-to-a-git-server-from-a-workspace.adoc
```diff
@@ -9,11 +9,13 @@
 
 In a workspace, you can run Git commands that require user authentication like cloning a remote private Git repository or pushing to a remote public or private Git repository.
 
-To configure user authentication to a Git server from a workspace, {prod-short} provides two options:
+User authentication to a Git server from a workspace can be configured by the administrator or user:
```
rkratky, line 12: The orig. version was better (and didn't use any passive voice).

modules/end-user-guide/pages/git-provider-access-token.adoc
```diff
@@ -0,0 +1,68 @@
+:navtitle: Git provider access token
+:description: Git provider access token
```
rkratky: Why doesn't the file name and this description reflect the title?

modules/end-user-guide/pages/git-credentials-store.adoc
```diff
@@ -1,22 +1,24 @@
-:navtitle: Using a Git credentials store
-:description: Using a Git credentials store
-:keywords: user-guide, configuring, user, secrets
-:page-aliases: 
+:navtitle: Git credentials store
+:description: Git credentials store
```
rkratky: Why doesn't the file name and this description reflect the title?

modules/end-user-guide/pages/using-git-credentials.adoc
```diff
@@ -0,0 +1,9 @@
+:navtitle: Using Git credentials
```
rkratky, line 2: Why is this a separate file? Wouldn't it be simpler to have a single procedure file with instructions for both the store and the token? Or, if those should be separate procedures, then keep them separate, but make a regular assembly out of this file.

rkratky (dismissed):

modules/end-user-guide/pages/authenticating-yourself-to-a-git-server-from-a-workspace.adoc
max-cx: I like the original version too. However, we need to add the important bit about `by the administrator or user`. How can we do that differently?

modules/end-user-guide/pages/using-git-credentials.adoc
max-cx: If i make a regular assembly with two procedures, then i can't reference either procedure in any other page, as far as i'm aware.

max-cx: But i guess i can drop those references and just reference the assembly page.

modules/end-user-guide/pages/git-credentials-store.adoc
max-cx: The user has other means to apply the Secret, not just what's in the block.

modules/end-user-guide/pages/git-credentials-store.adoc
max-cx: I use `source,terminal` to make the label **TERMINAL** appear at the upper right corner of the block when the reader hovers the cursor over the block (same as for YAML).

modules/end-user-guide/pages/git-provider-access-token.adoc
max-cx: I've removed `+macros`, and as already mentioned, I use `source,terminal` to make the label **TERMINAL** appear at the upper right corner of the block when the reader hovers the cursor over the block (same as for YAML).

modules/end-user-guide/pages/using-git-credentials.adoc
max-cx: OK, so that's what i'll do.

max-cx: Done. Thank you again for this suggestion.

modules/end-user-guide/pages/authenticating-yourself-to-a-git-server-from-a-workspace.adoc
max-cx: Thank you. I've removed `yourself` from the file and file name and the xrefs to that file.

modules/end-user-guide/pages/git-provider-access-token.adoc
max-cx: My reasoning was to save nav menu space.
This is outdated now because i included these modules as partials, so these lines are gone now.

modules/end-user-guide/pages/git-credentials-store.adoc
max-cx: My reasoning was to save nav menu space.
This is outdated now because i included these modules as partials, so these lines are gone now.

vinokurig (approved):

PR #2406, opened 2022-07-21
fix: importing untrusted tls certificates language review

deerskindoll:
language review of importing-untrusted-tls-certificates.adoc

https://issues.redhat.com/browse/RHDEVDOCS-3500

3.x

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -12,7 +12,7 @@ By default, external communications between {prod-short} components are encrypte
 When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component treats the certificates as signed by a trusted CA. You have to do this in the following cases:
 
 * The underlying {orch-name} cluster uses TLS certificates signed by an untrusted CA.
-* {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
+{prod-short} server or workspace components connect to external OIDC providers or a Git server that use TLS certificates signed by an untrusted CA.
 
 {prod-short} uses labeled ConfigMaps in {orch-namespace} as sources for TLS certificates. The ConfigMaps can have an arbitrary number of keys with a random number of certificates each.
 
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -9,7 +9,7 @@
 
 By default, external communications between {prod-short} components are encrypted with TLS. Communications of {prod-short} components with external services such as proxies, source code repositories, and identity provider might also require TLS. All communications encrypted with TLS require the use of TLS certificates signed by trusted Certificate Authorities (CA).
 
-When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component considers the certificates as signed by a trusted CA. Following cases typically require this addition:
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component treats the certificates as signed by a trusted CA. You have to do this in the following cases:
 
 * The underlying {orch-name} cluster uses TLS certificates signed by an untrusted CA.
 * {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -10,6 +10,7 @@
 By default, external communications between {prod-short} components are encrypted with TLS. Communications of {prod-short} components with external services such as proxies, source code repositories, and identity provider might also require TLS. All communications encrypted with TLS require the use of TLS certificates signed by trusted Certificate Authorities (CA).
 
 When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component considers the certificates as signed by a trusted CA. Following cases typically require this addition:
+
 * The underlying {orch-name} cluster uses TLS certificates signed by an untrusted CA.
 * {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
 
--- a/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
+++ b/modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
@@ -7,29 +7,23 @@
 [id="importing-untrusted-tls-certificates_{context}"]
 = Importing untrusted TLS certificates to {prod-short}
 
-External communications between {prod-short} components are, by default, encrypted with TLS. Communications of {prod-short} components with external services such as proxies, source code repositories, and identity provider may require using TLS. Those communications require the use of TLS certificates signed by trusted Certificate Authorities.
+By default, external communications between {prod-short} components are encrypted with TLS. Communications of {prod-short} components with external services such as proxies, source code repositories, and identity provider might also require TLS. All communications encrypted with TLS require the use of TLS certificates signed by trusted Certificate Authorities (CA).
 
-NOTE: When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it can be necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component will consider them as signed by a trusted CA.
-
-Typical cases that may require this addition are:
-
-* When the underlying {orch-name} cluster that uses TLS certificates signed by a CA that is not trusted.
-* When {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it's necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component considers them as signed by a trusted CA. Following cases typically require this addition:
+* The underlying {orch-name} cluster uses TLS certificates signed by a CA that isn't trusted.
+* {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
 
 {prod-short} uses labeled ConfigMaps in {orch-namespace} as sources for TLS certificates. The ConfigMaps can have an arbitrary number of keys with a random number of certificates each.
 
 [NOTE]
 ====
-When the OpenShift cluster contains cluster-wide trusted CA certificates added through the link:https://docs.openshift.com/container-platform/4.4/networking/configuring-a-custom-pki.html#nw-proxy-configure-object_configuring-a-custom-pki[cluster-wide-proxy configuration], {prod-short} Operator detects them and automatically injects them into this ConfigMap:
-
-- {prod-short} automatically labels the ConfigMap with the `config.openshift.io/inject-trusted-cabundle="true"` label.
-- Based on this annotation, OpenShift automatically injects the cluster-wide trusted CA certificates inside the `ca-bundle.crt` key of ConfigMap
+When an OpenShift cluster contains cluster-wide trusted CA certificates added through the link:https://docs.openshift.com/container-platform/4.10/networking/configuring-a-custom-pki.html#nw-proxy-configure-object_configuring-a-custom-pki[cluster-wide-proxy configuration], {prod-short} Operator detects them and automatically injects them into a ConfigMap. {prod-short} automatically labels the ConfigMap with the `config.openshift.io/inject-trusted-cabundle="true"` label. Based on this annotation, OpenShift automatically injects the cluster-wide trusted CA certificates inside the `ca-bundle.crt` key of ConfigMap.
 ====
 
 [IMPORTANT]
 ====
-Some {prod-short} components require to have a full certificate chain to trust the endpoint.
-If the cluster is configured with an intermediate certificate, then the whole chain (including self-signed root) should be added to {prod-short}.
+Some {prod-short} components require a full certificate chain to trust the endpoint.
+If the cluster is configured with an intermediate certificate, add the whole chain (including self-signed root) to {prod-short}.
 ====
 
 == Adding new CA certificates into {prod-short}
@@ -46,11 +40,11 @@ Certificate files are typically stored as Base64 files, with extensions such as 
 
 .Procedure
 
-. Save the certificates you need to import, to a local file system.
+. Save the certificates you need to import to a local file system.
 +
 [CAUTION]
 ====
-* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported. Convert it to the supported `CERTIFICATE` format with the following command:
+* A certificate with the introductory phrase `BEGIN TRUSTED CERTIFICATE` is likely in the PEM `TRUSTED CERTIFICATE` format, which is not not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
 ** `openssl x509 -in cert.pem -out cert.cer`
 ====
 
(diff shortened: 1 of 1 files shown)


modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
 [id="importing-untrusted-tls-certificates_{context}"]
 = Importing untrusted TLS certificates to {prod-short}
```
jc-berger, line 8: FYI, I can't find this content in the preview build that the PR prompts me to select for previews. https://62d9769557e45f00589af6f8--eclipse-che-docs-pr.netlify.app/docs/next/hosted-che/hosted-che/

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
...
 
-NOTE: When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it can be necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component will consider them as signed by a trusted CA.
-
-Typical cases that may require this addition are:
-
-* When the underlying {orch-name} cluster that uses TLS certificates signed by a CA that is not trusted.
-* When {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it's necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component considers them as signed by a trusted CA. Following cases typically require this addition:
```
jc-berger: User focus. 

Be user and action focused instead of using passive voice. What do you think of this rewrite? 

`When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate in the {prod-short} installation so that every {prod-short} component considers the certificates as signed by a trusted CA. The following cases require trusted CA authorization:` 

The last sentence kind of confused me, what do you think of my suggested rewrite? 

Also, `them` was a little confusing since it could conceivably refer to multiple things. So even though it's wordier, I think for clarity, it's good to say, `considers the certificates as signed by a trusted CA`.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
...
+* The underlying {orch-name} cluster uses TLS certificates signed by a CA that isn't trusted.
```
jc-berger: Am I understanding this bullet list right? I'm reading it as a list of cases where you NEED to make sure it's signed by a trusted CA. Therefore, if it is NOT, you must import the trusted CA signature (like you stated above)? 

For minimalism, I think this sentence could be re-phrased, what do you think of this?: 

`The following cases require trusted CA authorization:

- The underlying {orch-name} cluster that uses TLS certificates
- {prod-short} server or workspace components that connect to external services such as {identity-provider} or a Git server that use TLS certificates`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
...
-
-- {prod-short} automatically labels the ConfigMap with the `config.openshift.io/inject-trusted-cabundle="true"` label.
-- Based on this annotation, OpenShift automatically injects the cluster-wide trusted CA certificates inside the `ca-bundle.crt` key of ConfigMap
+When an OpenShift cluster contains cluster-wide trusted CA certificates added through the link:https://docs.openshift.com/container-platform/4.10/networking/configuring-a-custom-pki.html#nw-proxy-configure-object_configuring-a-custom-pki[cluster-wide-proxy configuration], {prod-short} Operator detects them and automatically injects them into a ConfigMap. {prod-short} automatically labels the ConfigMap with the `config.openshift.io/inject-trusted-cabundle="true"` label. Based on this annotation, OpenShift automatically injects the cluster-wide trusted CA certificates inside the `ca-bundle.crt` key of ConfigMap.
```
jc-berger, line 21: Usability 

I would remove the link. It distracts the user from the task at hand and prompts the user to leave your documentation before even finishing the sentence. 

I think links are best used in prereqs or additional resources. They're too distracting in a procedure or informative paragraph, etc 

`When an OpenShift cluster contains cluster-wide trusted CA certificates added through the cluster-wide-proxy configuration, {prod-short} Operator detects the certificates and automatically injects them into a ConfigMap.`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
...
 ====
 
 [IMPORTANT]
 ====
-Some {prod-short} components require to have a full certificate chain to trust the endpoint.
-If the cluster is configured with an intermediate certificate, then the whole chain (including self-signed root) should be added to {prod-short}.
+Some {prod-short} components require a full certificate chain to trust the endpoint.
+If the cluster is configured with an intermediate certificate, add the whole chain (including self-signed root) to {prod-short}.
```
jc-berger: IBM SG, avoid parenthesis except for acronyms. 

`If the cluster is configured with an intermediate certificate, add the whole chain, including self-signed root, to {prod-short}.`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -46,11 +40,11 @@ Certificate files are typically stored as Base64 files, with extensions such as
 
 .Procedure
 
-. Save the certificates you need to import, to a local file system.
+. Save the certificates you need to import to a local file system.
 +
 [CAUTION]
 ====
-* A certificate with the `BEGIN TRUSTED CERTIFICATE` introductory phrase is likely in the PEM `TRUSTED CERTIFICATE` format, which is not supported. Convert it to the supported `CERTIFICATE` format with the following command:
+* A certificate with the introductory phrase `BEGIN TRUSTED CERTIFICATE` is likely in the PEM `TRUSTED CERTIFICATE` format, which is not not supported by Java. Convert it to the supported `CERTIFICATE` format with the following command:
```
jc-berger: Omit the extra `not`

`...which is not supported by Java.`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
 $ {orch-cli} create configmap custom-certs --from-file=__<bundle-file-path>__ -n={prod-namespace}
 ----
 +
-Add another `-from-file=_<bundle-file-path>_` flag to apply more than one bundle. Otherwise, create another ConfigMap.
+To apply more than one bundle, add another `-from-file=_<bundle-file-path>_`. Otherwise, create another ConfigMap. 
+// Wouldn't "alternatively" make more sense than "otherwise"?
```
jc-berger: You need to delete this peer review note/question: `// Wouldn't "alternatively" make more sense than "otherwise"?`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
-. Label created ConfigMaps with both `app.kubernetes.io/part-of=che.eclipse.org` and `app.kubernetes.io/component=ca-bundle` labels:
+. Label created ConfigMaps with the `app.kubernetes.io/part-of=che.eclipse.org` and `app.kubernetes.io/component=ca-bundle` labels:
 +
 [subs="+attributes,+quotes"]
 ----
 $ {orch-cli} label configmap custom-certs app.kubernetes.io/part-of=che.eclipse.org app.kubernetes.io/component=ca-bundle -n <{project-context}-namespace-name>
 ----
 
-. Deploy {prod-short} if it has not been deployed before. Otherwise wait until the rollout of {prod-short} components finishes. If there are running workspaces, they should be restarted for the changes to take effect.
+. Deploy {prod-short} if it hasn't been deployed before. Otherwise wait until the rollout of {prod-short} components finishes. Restart running workspaces for the changes to take effect.
```
jc-berger: Avoid contractions and this look like it'd be two commands, not one. 

1. Deploy {prod-short} if it has not been deployed. Otherwise wait until the rollout of {prod-short} components finishes.
2. Restart running workspaces for the changes to take effect.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
+
+== Troubleshooting imported certificate issues
+
+If issues occur after adding the certificates, verify the below specified values at a {prod-short} installation and workspace level.
```
jc-berger: Accessibility and usability. 

Avoid directional words like `above` and `below`

`If issues occur after adding the certificates, verify the specified values at a {prod-short} installation and workspace level.`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
 
-== Verification at the {prod-short} installation level
 
-When something does not work as expected after adding the certificates, here is a list of things to verify:
+.Verifying imported certificates at the {prod-short} installation level
 
-- In case of a {prod-short} link:https://docs.openshift.com/container-platform/latest/operators/understanding/olm-what-operators-are.html[Operator] deployment, the namespace where the `CheCluster` is located contains labeled ConfigMaps with the right content:
+* In case of a {prod-short} link:https://docs.openshift.com/container-platform/latest/operators/understanding/olm-what-operators-are.html[Operator] deployment, the namespace where the `CheCluster` is located contains labeled ConfigMaps with the right content:
```
jc-berger: Again, omit links and the peer review note/question. 

``In case of a {prod-short} Operator deployment, the namespace where the `CheCluster` is located contains labeled ConfigMaps with the correct content:`

Note, I think for word choice `correct` is better since `right` could mean direction, not accuracy.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
+// Correct content? Relevant content?
```
jc-berger: omit this content.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
 +
 [subs="+attributes,+quotes",options="nowrap",role=white-space-pre]
 ----
 $ {orch-cli} get cm --selector=app.kubernetes.io/component=ca-bundle,app.kubernetes.io/part-of=che.eclipse.org -n {prod-namespace}
 ----
 +
-Check the content of ConfigMap by running:
+Check the content of ConfigMap by entering this command:
```
jc-berger: `Check the content of ConfigMap by entering:`

OR

`Check the content of ConfigMap by entering the following command:` 

I think the first versionI suggested reads better, but you could rephrase to either one of these :)

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
+* {prod-short} Pod Volumes list contains a volume that uses `ca-certs-merged` ConfigMap as data-source.
+To get the list of Volumes of the {prod-short} Pod, run:
 +
 [subs="+attributes,+quotes",options="nowrap",role=white-space-pre]
 ----
 $ {orch-cli} get pod -o json __<{prod-id-short}-pod-name>__ -n {prod-namespace} | jq .spec.volumes
 ----
 +
-- {prod-short} mounts certificates in folder `/public-certs/` of the {prod-short} server container. This command returns the list of files in that folder:
+* {prod-short} mounts certificates in the `/public-certs/` folder of the {prod-short} server container. Enter the following command to view the list of files in this folder:
```
jc-berger: Reword to be consistent with your previous commands: 

`To view the list of files in this folder, enter:`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,63 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
 +
 [subs="+attributes,+quotes",options="nowrap",role=white-space-pre]
 ----
 $ {orch-cli} exec -t __<{prod-id-short}-pod-name>__ -n {prod-namespace} -- ls /public-certs/
 ----
 +
-- In the {prod-short} server logs, there is a line for every certificate added to the Java truststore, including configured {prod-short} certificates.
+* In the {prod-short} server logs, there is a line for every certificate added to the Java truststore, including configured {prod-short} certificates. View them with following command:
```
jc-berger: For all these bullet points, you precede the action with context. I'm not used to seeing this. 

Can you simply give the users a series of actions and make this a procedure module?

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -122,41 +121,42 @@ Your keystore contains 141 entries:
...
 ----
 $ openssl x509 -in __<certificate-file-path>__ -fingerprint -noout
 SHA1 Fingerprint=3F:DA:BF:E7:A7:A7:90:62:CA:CF:C7:55:0E:1D:7D:05:16:7D:45:60
 ----
 
-== Verification at the workspace level
+.Verifying imported certificates at the workspace level
 
-- Start a workspace, obtain the {orch-namespace} name in which it has been created, and wait for the workspace to be started.
+* Start a workspace, obtain the {orch-namespace} name in which it has been created, and wait for the workspace to be started.
```
jc-berger: No need for the comma before `and` 

`...it has been created and wait for the workspace to be started.`

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -122,41 +121,42 @@ Your keystore contains 141 entries:
...
+// What do I need the name for?
```
jc-berger: remove this content

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -122,41 +121,42 @@ Your keystore contains 141 entries:
...
 
-- Get the name of the workspace Pod with the following command:
+* Get the name of the workspace Pod with the following command:
```
jc-berger: minimalism 

Avoid the whole `with this command` it fluff and unnecessary.

jc-berger (changes requested):

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
deerskindoll: Here's the link: https://62d9769557e45f00589af6f8--eclipse-che-docs-pr.netlify.app/docs/next/administration-guide/importing-untrusted-tls-certificates/

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
rkratky: @jc-berger Thank you for going the extra mile on this. I agree that adding the info to Additional resources could be helpful, too. But I don't agree with removing it from the text. Links are an inherent and useful part of the web. It would be a different matter if, let's say, every second term in the paragraph was a link, or if the link was overly long. But that's not the case. This is well placed and useful. Users are well aware of hypertext's capability to include links, and they can use them when they need. It's not confusing.

Btw, see pg. 270 of the IBM SG.

max-cx:
1. I recommend breaking this file down into multiple files because this looks like an assembly that needs to include at least two files. We don't normally include multiple procedures in one file in modular docs.
2. I recommend rethinking the formatting of the second section (_Troubleshooting imported certificate issues_). I'll think of some specific suggestions on this point.
I've opened RHDEVDOCS-4246 for these changes.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,62 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
+. Restart running workspaces for the changes to take effect.
+
+== Troubleshooting imported certificate issues
+
+If issues occur after adding the certificates, verify the specified values at a {prod-short} installation and workspace level.
 
-== Verification at the {prod-short} installation level
 
-When something does not work as expected after adding the certificates, here is a list of things to verify:
+.Verifying imported certificates at the {prod-short} installation level
```
max-cx: Installation to me sounds like the process of installing something. This reads ambiguously here.

```suggestion
.Verifying imported certificates at the {prod-short} instance level
```

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -61,58 +55,62 @@ Certificate files are typically stored as Base64 files, with extensions such as
...
 $ {orch-cli} label configmap custom-certs app.kubernetes.io/part-of=che.eclipse.org app.kubernetes.io/component=ca-bundle -n <{project-context}-namespace-name>
 ----
 
-. Deploy {prod-short} if it has not been deployed before. Otherwise wait until the rollout of {prod-short} components finishes. If there are running workspaces, they should be restarted for the changes to take effect.
+. Deploy {prod-short} if it hasn't been deployed before. Otherwise wait until the rollout of {prod-short} components finishes. 
+. Restart running workspaces for the changes to take effect.
+
+== Troubleshooting imported certificate issues
+
+If issues occur after adding the certificates, verify the specified values at a {prod-short} installation and workspace level.
```
max-cx: Installation to me sounds like the process of installing something. This reads ambiguously here.
```suggestion
If issues occur after adding the certificates, verify the specified values at the {prod-short} instance level and workspace level.
```

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
...
 
-NOTE: When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it can be necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component will consider them as signed by a trusted CA.
-
-Typical cases that may require this addition are:
-
-* When the underlying {orch-name} cluster that uses TLS certificates signed by a CA that is not trusted.
-* When {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate in the {prod-short} installation so that every {prod-short} component considers the certificates as signed by a trusted CA. Following cases typically require this addition:
```
max-cx: Installation to me sounds like the process of installing something. This reads ambiguously here.

```suggestion
When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component considers the certificates as signed by a trusted CA. Following cases typically require this addition:
```

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
jc-berger: Thanks, Robert :) I always default to the writer knowing their content better than I as the reviewer. I didn't mean to imply links can never be used. As you pointed out, It's a good choice to use here.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,23 @@
...
 
-NOTE: When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it can be necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component will consider them as signed by a trusted CA.
-
-Typical cases that may require this addition are:
-
-* When the underlying {orch-name} cluster that uses TLS certificates signed by a CA that is not trusted.
-* When {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component considers the certificates as signed by a trusted CA. Following cases typically require this addition:
```
jc-berger: This content is good, but it doesn't render properly, at least not in the preview:
Screen Shot 2022-07-22 at 1 59 03 PM

jc-berger (approved):
/lgtm

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
deerskindoll: this will be addressed in https://issues.redhat.com/browse/RHDEVDOCS-4246

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
deerskindoll: fixed

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
deerskindoll: I think it's OK to keep it as it is because the current phrasing takes into account that the components might be using trusted certificates from the get-go.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,24 @@
...
+By default, external communications between {prod-short} components are encrypted with TLS. Communications of {prod-short} components with external services such as proxies, source code repositories, and identity provider might also require TLS. All communications encrypted with TLS require the use of TLS certificates signed by trusted Certificate Authorities (CA).
 
-NOTE: When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it can be necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component will consider them as signed by a trusted CA.
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component considers the certificates as signed by a trusted CA. Following cases typically require this addition:
```
max-cx: Two suggestions:
* `considers` (sounds like anthropomorphism or AI) > `treats`
* `Following cases typically require this addition` > `You have to do this in the following cases:` (makes it read more user-centric)
```suggestion
When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component treats the certificates as signed by a trusted CA. You have to do this in the following cases:
```

tolusha (approved):

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -7,29 +7,24 @@
...
 
-NOTE: When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, it can be necessary to import the CA certificate in the {prod-short} installation so that every {prod-short} component will consider them as signed by a trusted CA.
+When the certificates used by {prod-short} components or by an external service are signed by an untrusted CA, you must import the CA certificate into the {prod-short} instance so that every {prod-short} component treats the certificates as signed by a trusted CA. You have to do this in the following cases:
 
-Typical cases that may require this addition are:
-
-* When the underlying {orch-name} cluster that uses TLS certificates signed by a CA that is not trusted.
-* When {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
+* The underlying {orch-name} cluster uses TLS certificates signed by an untrusted CA.
+* {prod-short} server or workspace components connect to external services such as {identity-provider} or a Git server that use TLS certificates signed by an untrusted CA.
```
l0rd: Today Keycloak is not the default OIDC provider anymore. Hence we may need to change that in https://github.com/eclipse-che/che-docs/blob/main/antora.yml#L29.

deerskindoll: Is there something I can do? Who's the current default provider?

max-cx: @l0rd, what should be the new value for these attributes/variables?
```
    identity-provider-id: keycloak
    identity-provider: Keycloak
```
@deerskindoll, btw, looks like we might need to update the downstream docs as well; currently, this is downstream:
```
:identity-provider-id: keycloak
:identity-provider: RH-SSO
```

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
```diff
@@ -122,41 +121,41 @@ Your keystore contains 141 entries:
...
-- Get the name of the workspace Pod with the following command:
+* Get the name of the workspace Pod:
 +
 [subs="+attributes,+quotes",options="nowrap",role=white-space-pre]
 ----
 $ {orch-cli} get pods -o=jsonpath='{.items[0].metadata.name}' -n __<workspace namespace>__ | grep '^workspace.*'
 ----
 
-- Get the name of the Che-Theia IDE container in the workspace Pod with the following command:
+* Get the name of the Che-Theia IDE container in the workspace Pod:
```
l0rd, line 143: Che-theia is just one of the IDEs that we support so we may want to change that, probably in another PR.

l0rd (approved):
LGTM. This doc required some updates related to the switch to devworkspace but it can be done in a separate issue/PR.

modules/administration-guide/pages/importing-untrusted-tls-certificates.adoc
max-cx: I propose the following two. They are the same both upstream and downstream, correct? 
```
    identity-provider-id: openshift-oauth
    identity-provider: OpenShift OAuth
```

max-cx: Feedback from Anatolii:
image

max-cx: I'm opening a RHDEVDOCS Jira issue for this: https://issues.redhat.com/browse/RHDEVDOCS-4296.

max-cx: Thank you, @tolusha, for drafting this change:
```suggestion
{prod-short} server or workspace components connect to external OIDC providers or a Git server that use TLS certificates signed by an untrusted CA.
```

max-cx (approved):

PR #2409, opened 2022-07-27
chore: Add instructions how to add a manuall PAT secret (#2383 -> 7.50.x)

vinokurig:
Add instructions how to add a manuall PAT secret to be able to create factory from private repo if oAuth is not configured.

Chery pick from the main branch, see https://github.com/eclipse-che/che-docs/pull/2383#issuecomment-1195241618
## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


l0rd:
@vinokurig for my understanding: why we need 2 distinct procedures? I would have rather replaced the existing one (using-a-git-credentials-store_che) with the new one (using-a-git-provider-access-token_che) since the git credential store will be created automatically with the new one hiding that complexity from the user. But more importantly the `using-a-git-credentials-store_che` has a big problem with repositories that have a devfile in it and I would discourage it at all. If what I am saying makes sense I can create a new PR to remove the old procedure.

vinokurig:
@l0rd I don't think that the procedure`using-a-git-credentials-store_che` is redundant because even if it could be created form the PAT secret, user might want to create it separately. However I am OK to have just one approach, so you are welcome with a PR.

PR #2414, opened 2022-08-02
Adding a list of supported languages

deerskindoll:
Adding a list of supported languages to the user guide.

https://issues.redhat.com/browse/RHDEVDOCS-4164

Dev Spaces 3.0, 3.1

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


modules/end-user-guide/partials/ref-supported-languages.adoc
```diff
@@ -0,0 +1,8 @@
+[id="supported-languages_{context}"]
+= Supported languages
+
+Following table lists languages supported by {prod-short}.
```
rkratky:
```suggestion
```

modules/end-user-guide/nav.adoc
```diff
@@ -4,6 +4,7 @@
 ** xref:developer-workspaces.adoc[]
 ** xref:first-time-contributors.adoc[]
 ** xref:benefits-of-pull-requests-review-in-che.adoc[]
+** xref:supported-languages.adoc[]
```
rkratky, line 7: Such file (`supported-languages.adoc`) doesn't exist (the PR doesn't add it), so the link cannot work. In order for it to work, add a `supported-languages.adoc` file into the `pages/` directory.

modules/end-user-guide/partials/ref-supported-languages.adoc
```diff
@@ -0,0 +1,8 @@
+[id="supported-languages_{context}"]
```
rkratky: This file needs to be in the `pages/` directory, and it must be called `supported-languages.adoc`.

rkratky (changes requested):
@deerskindoll, comments added.

modules/end-user-guide/partials/ref-supported-languages.adoc
deerskindoll: I see. I tried moving it to the pages folder but because I didn't rename it, it didn't work for obvious reasons.

rkratky (approved):

l0rd:
@rkratky @themr0c Eclipse Che is not directly supporting languages anymore. We are supporting IDEs (VS Code, Theia, IntelliJ, PyCharm) and each IDE has a list of languages that it supports. That means that Eclipse Che supports thousands of languages and what is listed here is just a list of source code examples that we include in Che. How can we update that languages list so that it makes more sense for Che?

l0rd:
cc @Kasturi1820

PR #2441, opened 2022-09-12
fix: use underscores in user-replaced values

themr0c:
Use underscores in user-replaced values, rather than hyphens, to comply with https://redhat-documentation.github.io/supplementary-style-guide/#user-replaced-values.

fixes: https://issues.redhat.com/browse/RHDEVDOCS-3498


nickboldt (approved):
seems reasonable

themr0c:
Error on "CSVs" is a false positive. Opening  https://github.com/redhat-documentation/vale-at-red-hat/issues/299 and merging this.

PR #2452, opened 2022-09-21
chore: add missing templates referenced in CONTRIBUTING.adoc

themr0c:
chore: add missing templates referenced in CONTRIBUTING.adoc


templates/partial/snippet.adoc
```diff
@@ -0,0 +1,3 @@
+// A text snippet is a section of text that is stored in an AsciiDoc file.
```
max-cx, line 1: `snip` or `snippet`
BTW, we've been using `snip`.

themr0c: ???

max-cx: I meant the `snip_...adoc` file-name convention in our docs:
https://github.com/eclipse-che/che-docs/tree/main/modules/administration-guide/examples

Anyway, I guess it shouldn't matter in this file.

max-cx (approved):

PR #2454, opened 2022-09-22
chore: next version is 7.54

themr0c:
Fix version for /next/.


tolusha (approved):

themr0c:
That's probably not good. We `7.54` being not released, we can't single-source from that branch. We probably need to have `main` there.

themr0c:
```
Failure: Cannot read che.properties from URL https://raw.githubusercontent.com/eclipse-che/che-server/7.54.x/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```

themr0c:
But `7.50` is wrong ...

PR #2464, opened 2022-10-07
fix: add isDefault to the start-preview-server task

themr0c:
Update the devfile commands according to the latest devfile docs in https://devfile.io/docs/2.1.0/adding-commands

* Add `isDefault: true` to the start-preview-server task, in the hope that the default "Run Build Task" (Ctrl+Shift+B) would work. (It doesn't).
* Remove `label` as is it not used by VS Code, and doesn't appear in devfile docs.
* Remove the obsolete `newtopic` command (script is unmaintained on our side).


l0rd (approved):

max-cx:
@themr0c, which branches to merge/chery-pick this into and what's the milestone?

PR #2468, opened 2022-10-12
fix: maven procedure update

max-cx:
Update the Maven-related procedure in response to a review of a customer support case and Nick's proposal.

https://issues.redhat.com/browse/RHDEVDOCS-4547
https://issues.redhat.com/browse/CRW-3418

`main` and cherry-pick to `7.52.x` and `7.53.x`

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
+++ b/modules/end-user-guide/pages/enabling-maven-artifact-repositories.adoc
@@ -12,10 +12,11 @@ You can enable a Maven artifact repository in Maven workspaces that run in a res
 .Prerequisites
 
 * You are not running any Maven workspace.
+* You know your user namespace, which is `__<username>__-{prod-id-short}` where `__<username>__` is your {prod-short} username.
 
 .Procedure
 
-. Apply the Secret for the TLS certificate:
+. In the `__<username>__-{prod-id-short}` namespace, apply the Secret for the TLS certificate:
 +
 [source,yaml,subs="+quotes,+attributes,+macros"]
 ----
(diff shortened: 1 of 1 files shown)


max-cx:
@nickboldt, fyi:
 
1.

Upstream: `prod-id-short: che`

Downstream: `:prod-id-short: devspaces`

2.

I initially tried incorporating the following info from your proposal into the Prerequisites:

```
* You are logged into an Openshift cluster.
* ... NOT the `openshift-operators` or `openshift-devspaces` namespaces.
```

But then I thought we can drop it:
- `You are logged into an Openshift cluster` is something that is self-explanatory, as well as inapplicable to the Che instances installed on Kubernetes. I could indeed split it into upstream/downstream examples, but as I just mentioned, this is self-explanatory, so we can avoid unnecessary complexity and effort.
- `NOT the openshift-operators or openshift-devspaces namespaces` can be skipped because I already added the user namespace to both the Prerequisites and every step in the procedure, so it'd be a distraction and an overkill.

WDYT?

nickboldt (approved):
While we generally avoid using second-person statements in procedures like "you are not" or "you know", this is at least consistent with where we HAVE broken the style guide to use second-person statements.

+1, LGTM

nickboldt:
Given 7.55.x is in progress today, you can probably skip cherrypicking this to 7.53.x and instead pick it to 7.55.x

svor (approved):

deerskindoll (approved):
lgtm

PR #2470, opened 2022-10-17
fix: use 'restricted environment' rather than 'air-gap'

themr0c:
fixes https://issues.redhat.com/browse/RHDEVDOCS-4545


modules/end-user-guide/pages/troubleshooting-slow-workspaces.adoc
```diff
@@ -32,7 +32,7 @@ Installing offline::
 +
 _Role: Administrator_
 +
-Components of {prod-short} are OCI images. Set up {prod} in offline mode (air-gap scenario) to reduce any extra download at runtime because everything needs to be available from the beginning. See xref:administration-guide:installing-che-in-a-restricted-environment.adoc[].
+Components of {prod-short} are OCI images. Set up {prod} in offline mode to reduce any extra download at runtime because everything needs to be available from the beginning. See xref:administration-guide:installing-che-in-a-restricted-environment.adoc[].
```
max-cx, line 35: I'm fully aware that this phrase is in the original text and out of the scope of this issue, but since it's related to the removed phrase in the sentence, my first question is "What is the offline mode?" and it's worth asking: is there an `offline mode` that a Che admin can switch on/off? If not, then I suggest editing it to something like this:
```suggestion
Components of {prod-short} are OCI images. Set up {prod} to run offline to reduce any extra download at runtime because everything needs to be available from the beginning. See xref:administration-guide:installing-che-in-a-restricted-environment.adoc[].
```

max-cx (approved):
I added one suggestion.

PR #2477, opened 2022-11-02
fix: supported languages page update

deerskindoll:
supported languages page is updated to a stack samples page instead

https://issues.redhat.com/browse/RHDEVDOCS-4518

3.2, 3.3

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:
- [ x] the *`Validate language on files added or modified`* step reports no vale warnings.

### DIFF
--- a/modules/end-user-guide/pages/supported-languages.adoc
+++ b/modules/end-user-guide/pages/supported-languages.adoc
@@ -1,10 +1,18 @@
 :_content-type: REFERENCE
-:description: Supported languages
+:description: Stack samples
 :keywords: user-guide, languages
-:navtitle: Supported languages
+:navtitle: Stack samples
 :page-aliases:
 
 [id="supported-languages"]
-= Supported languages
+= Stack samples
 
 include::example$snip_{project-context}-supported-languages.adoc[leveloffset=+1]
+
+[NOTE]
+====
+The amount of supported languages may vary depending on the selected IDE. For a full list of languages supported by your IDE, refer to its documentation.
+====
+
+.Additional resources
+. link:https://github.com/eclipse-che/che-devfile-registry/tree/main/devfiles[Git repository] of stack samples using various programing languages.


modules/end-user-guide/pages/supported-languages.adoc
```diff
@@ -1,10 +1,18 @@
 :_content-type: REFERENCE
-:description: Supported languages
+:description: Stack samples
 :keywords: user-guide, languages
-:navtitle: Supported languages
+:navtitle: Stack samples
 :page-aliases:
 
 [id="supported-languages"]
```
max-cx, line 7: I suggest changing the file name from `supported-languages.adoc` to `stack-samples.adoc` and updating `nav.adoc`.
Also the `id` needs to be updated so that everything is aligned:
```suggestion
[id="stack-samples"]
```

modules/end-user-guide/pages/supported-languages.adoc
```diff
@@ -1,10 +1,18 @@
...
-= Supported languages
+= Stack samples
```
max-cx, line 9: Here's my suggested text for a rewrite of the NOTE from the bottom of the file.
IMO, this is a better place for this bit, and the reader better read it before reading the list of stacks.

```suggestion
TIP: Do not confuse stack samples and programming languages that are supported by the in-browser IDE that you select for a workspace. For programming language support, consult the documentation of the IDE and its applicable extensions.

```

max-cx: @l0rd, does the above suggestion address your concern here?

modules/end-user-guide/pages/supported-languages.adoc
```diff
@@ -1,10 +1,18 @@
 :_content-type: REFERENCE
-:description: Supported languages
+:description: Stack samples
 :keywords: user-guide, languages
-:navtitle: Supported languages
+:navtitle: Stack samples
 :page-aliases:
```
max-cx, line 5: If you change the file name as I suggested below:
```suggestion
:page-aliases: supported-languages.adoc
```

modules/end-user-guide/pages/supported-languages.adoc
```diff
@@ -1,10 +1,18 @@
...
 
 [id="supported-languages"]
-= Supported languages
+= Stack samples
 
 include::example$snip_{project-context}-supported-languages.adoc[leveloffset=+1]
```
max-cx, line 10: I'm not sure if we need to use `leveloffset=+1` with `include` statements. I never use it here. See if it works without it:
```suggestion
include::example$snip_{project-context}-supported-languages.adoc[]
```

modules/end-user-guide/pages/supported-languages.adoc
```diff
@@ -1,10 +1,18 @@
...
+
+[NOTE]
+====
+The amount of supported languages may vary depending on the selected IDE. For a full list of languages supported by your IDE, refer to its documentation.
+====
+
+.Additional resources
+. link:https://github.com/eclipse-che/che-devfile-registry/tree/main/devfiles[Git repository] of stack samples using various programing languages.
```
max-cx, line 18: I suggest you do another `include` statement here:
```suggestion
include::example$snip_{project-context}-link-to-devfile-registry.adoc[]
```
and move this bullet point to a snip file in the examples folder with just a line in it like this:
```
* link:https://github.com/eclipse-che/che-devfile-registry/tree/main/devfiles[Devfiles of {prod-short} stack samples on GitHub]
```
and then add a downstream snip file with a downstream link as well.

max-cx (changes requested):
See my suggestions.

modules/end-user-guide/pages/supported-languages.adoc
```diff
@@ -1,10 +1,18 @@
...
 
 [id="supported-languages"]
-= Supported languages
+= Stack samples
 
 include::example$snip_{project-context}-supported-languages.adoc[leveloffset=+1]
+
+[NOTE]
+====
+The amount of supported languages may vary depending on the selected IDE. For a full list of languages supported by your IDE, refer to its documentation.
```
jc-berger, line 14: IBM SG wants us to avoid `may` 

I think it reads perfectly fine to simply say, 
`The amount of supported languages vary depending on the selected IDE.`

modules/end-user-guide/pages/supported-languages.adoc
jc-berger, line 18: Not a necessary change, but for findability and usability, I think it reads better and is easier to click links at the end of a sentence. 

What do you think of this restructure? 

`To see stack samples using various programming languages, go to link:https://github.com/eclipse-che/che-devfile-registry/tree/main/devfiles[Git repository]`

jc-berger (changes requested):

deerskindoll:
content needs to be redone

PR #2479, opened 2022-11-07
chore: version 7.56.0 in main branch

themr0c:
## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -98,8 +98,8 @@ asciidoc:
     prod-upstream: Eclipse&#160;Che
     prod-url: "https://__&lt;che_fqdn&gt;__"
     prod-ver-major: "7"
-    prod-ver-patch: "7.55.0"
-    prod-ver: "7.55"
+    prod-ver-patch: "7.56.0"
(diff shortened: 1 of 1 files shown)


max-cx:
@themr0c, I wonder if we also need to check and update the following variables: 
```
prod-stable-channel-starting-csv: eclipse-che.v7.49.0
prod-operator-index: registry.access.redhat.com/redhat/community-operator-index:v4.10
ocp4-ver: "4.10"
devworkspace-operator-version-patch: "0.15.2"
devworkspace-operator-index: registry.redhat.io/redhat/redhat-operator-index:v4.10
```
If yes on any of them, then I have two suggestions:
1. Group into one separate section of this file all the variables that we need to double-check each time we update the release number
2. Add a comment with the link to the page where we can quickly check the latest value of each of those variables.

max-cx (approved):
Approving, but please first read my comment https://github.com/eclipse-che/che-docs/pull/2479#issuecomment-1305398374.

PR #2489, opened 2022-11-08
feat: Antora assembler

themr0c:
Use Antora Assembler to produce a monolithic AsciiDoc file for each module.

This technique is used to produce the downstream documentation. 

Applying the same technique upstream will help prevent committing incompatible content.

See: https://gitlab.com/antora/antora-assembler/

As a side effect, the Antora Assembler is producing PDF output. We have to be cautious with HTML specific content (use `ifndef`). I fixed legit vale warnings. I removed obsolete statements that were disabling vale, but are not required anymore.  One is left, and I found a workaround to avoid an error during PDF generation. Discussion in Antora community about this workaround, see: https://antora.zulipchat.com/#narrow/stream/282400-users/topic/Assembler.20specific.20attributes/near/308605574

In downstream, we also have different steps that are not reproduced here:
* Define different AsciiDoc attributes in the Antora playbook
* Define different examples in an [Antora Distributed Component Version](https://docs.antora.org/antora/latest/distributed-component-version/)
* Remove some pages from the navigation by using an Antora extension that injects Antora Collector configuration.

### DIFF
--- a/modules/administration-guide/pages/creating-a-telemetry-plugin.adoc
+++ b/modules/administration-guide/pages/creating-a-telemetry-plugin.adoc
@@ -397,7 +397,7 @@ spec:
       plugins:                           <2>
       - 'http://apache-che.apps-crc.testing/plugin.yaml'
 ----
-<1> The editor Id to set default plug-ins for.
+<1> The editor identification to set the default plug-ins for.
 <2> List of URLs to devfile v2 plug-ins.
 
 
--- a/modules/administration-guide/pages/creating-a-telemetry-plugin.adoc
+++ b/modules/administration-guide/pages/creating-a-telemetry-plugin.adoc
@@ -355,11 +355,8 @@ $ oc apply -f manifest.yaml
 ----
 
 .Verification steps
-
-pass:[<!-- vale RedHat.TermsErrors = NO -->]
-
-After the deployment has started, confirm that `plugin.yaml` is available in the web server:
-
+* After the deployment has started, confirm that `plugin.yaml` is available in the web server:
++
 ----
 $ curl apache-che.apps-crc.testing/plugin.yaml
 ----
@@ -367,7 +364,6 @@ $ curl apache-che.apps-crc.testing/plugin.yaml
 [id="specifying-the-telemetry-plug-in-in-a-devworkspace"]
 == Specifying the telemetry plug-in in a {devworkspace}
 . Add the following to the `components` field of an existing {devworkspace}:
-
 +
 ----
 components:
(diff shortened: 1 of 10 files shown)


modules/administration-guide/pages/creating-a-telemetry-plugin.adoc
```diff
@@ -405,7 +397,7 @@ spec:
       plugins:                           <2>
       - 'http://apache-che.apps-crc.testing/plugin.yaml'
 ----
-<1> The editorId to set default plug-ins for.
+<1> The editor Id to set default plug-ins for.
```
max-cx:
```suggestion
<1> The link:https://www.eclipse.org/che/docs/stable/end-user-guide/selecting-a-workspace-ide/[IDE `id`] to set the default plug-ins for.
```

modules/end-user-guide/examples/snip_che-supported-languages.adoc
```diff
@@ -1,24 +1,20 @@
...
 Go:: Stack with Go 1.14
 Java Lombok:: Java Stack with Lombok 1.18.18, OpenJDK 11 and Maven 3.6.0
 Java with Spring Boot and MongoDB:: Java stack with OpenJDK 8, MongoDB and Spring Boot Guestbook demo application
-Java Spring Boot:: Java stack with OpenJDK 11 and Spring Boot Petclinic demo application
-NodeJS Angular Web Application:: Stack for developing NodeJS Angular Web Application
-NodeJS Express Web Application:: Stack with NodeJS 10
-NodeJS MongoDB Web Application:: Stack with NodeJS 10 and MongoDB 3.4
-NodeJS React Web Application:: Stack for developing NodeJS React Web Application
-NodeJS Web Application based on Yarn:: Stack for developing NodeJS Web Application based on Yarn 
+Java Spring Boot:: Java stack with OpenJDK 11 and Spring Boot Pet clinic demo application
```
max-cx, line 9: In line with https://spring-petclinic.github.io/ ?
```suggestion
Java Spring Boot:: Java stack with OpenJDK 11 and Spring Boot PetClinic demo application
```

max-cx (approved):
@themr0c, please review my suggestions before merging this.

modules/administration-guide/pages/creating-a-telemetry-plugin.adoc
themr0c: According to https://www.merriam-webster.com/dictionary/id:

id is an abbreviation for identification.

So either:

```suggestion
<1> The editor id to set the default plug-ins for.
```

or:

```suggestion
<1> The editor identification to set the default plug-ins for.
```

but it's not code.

modules/end-user-guide/examples/snip_che-supported-languages.adoc
themr0c: PetClinic will produce a legit vale error. CamelCase is a programmatic way to write Camel Case. Therefore Pet clinic is legit.

themr0c:
Backported to 7.56.x as it fixes build errors.

PR #2491, opened 2022-11-11
procedures: uninstalling che process update (rhdevdocs-4450)

deerskindoll:
update of uninstallation steps

https://issues.redhat.com/browse/RHDEVDOCS-4450

3.1, 3.2, 3.3

## Pull Request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/uninstalling-che.adoc
+++ b/modules/administration-guide/pages/uninstalling-che.adoc
@@ -20,26 +20,16 @@ To uninstall an instance of {prod} {prod-ver}:
 
 .Procedure
 
-. Obtain the name of the {prod-short} {orch-namespace}. The default is `{prod-namespace}`.
+* Remove the {prod-short} instance from the `{prod-namespace}` {orch-namespace}:
 +
 [subs="+attributes,quotes"]
 ----
-$ {orch-cli} get checluster --all-namespaces \
-  -o=jsonpath="{.items[*].metadata.namespace}"
+$ {prod-cli} server:delete
 ----
 
-. Remove the {prod-short} instance from the `{prod-namespace}` {orch-namespace}:
-+
-[subs="+attributes,quotes"]
-----
-$ {prod-cli} server:delete -n {prod-namespace} # <1>
-----
-<1> The {orch-namespace} that you got in step 1.
+[TIP]
 
-include::example$snip_{project-context}-note-uninstalling-in-minikube-cluster.adoc[]
-
-[NOTE] 
-====
-If {prod-short} was installed from the OpenShift web console, `{prod-cli}` will not uninstall the {devworkspace} operator.
-To uninstall the {devworkspace} Operator, see link:https://docs.openshift.com/container-platform/{ocp4-ver}/web_console/odc-about-web-terminal.html#deleting-the-devworkspace-operator-dependency[Deleting the {devworkspace} Operator dependency].
-====
+----
+Use the `--delete-namespace` flag to remove the {prod} namespace in addition to the instance.
+Use the `--delete:all` flag to remove the {devworkspace} operator in addition to the {prod} namespace and instance.
+----
--- a/modules/administration-guide/pages/uninstalling-che.adoc
+++ b/modules/administration-guide/pages/uninstalling-che.adoc
@@ -27,6 +27,6 @@ $ {prod-cli} server:delete
 
 [TIP]
 ====
-Use the `--delete-namespace` flag to remove the {prod} namespace in addition to the instance.
-Use the `--delete:all` flag to remove the {devworkspace} operator in addition to the {prod} namespace and instance.
+The `--delete-namespace` option also removes the {prod-short} namespace.
+The `--delete-all` option also removes the {devworkspace} Operator, the related resources, and the {prod-short} namespace.
 ====
--- a/modules/administration-guide/pages/uninstalling-che.adoc
+++ b/modules/administration-guide/pages/uninstalling-che.adoc
@@ -28,5 +28,6 @@ $ {prod-cli} server:delete
 [TIP]
 ====
 The `--delete-namespace` option also removes the {prod-short} namespace.
+
 The `--delete-all` option also removes the {devworkspace} Operator, the related resources, and the {prod-short} namespace.
 ====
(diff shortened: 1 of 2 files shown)


modules/administration-guide/pages/uninstalling-che.adoc
```diff
@@ -20,26 +20,16 @@ To uninstall an instance of {prod} {prod-ver}:
...
-$ {orch-cli} get checluster --all-namespaces \
-  -o=jsonpath="{.items[*].metadata.namespace}"
+$ {prod-cli} server:delete
 ----
 
-. Remove the {prod-short} instance from the `{prod-namespace}` {orch-namespace}:
-+
-[subs="+attributes,quotes"]
+[TIP]
+
```
max-cx:
```suggestion
[TIP]
```

modules/administration-guide/pages/uninstalling-che.adoc
```diff
@@ -20,26 +20,16 @@ To uninstall an instance of {prod} {prod-ver}:
...
 ----
-$ {prod-cli} server:delete -n {prod-namespace} # <1>
+Use the `--delete-namespace` flag to remove the {prod} namespace in addition to the instance.
+Use the `--delete:all` flag to remove the {devworkspace} operator in addition to the {prod} namespace and instance.
 ----
```
max-cx: We need to use `====` for admonitions. BTW, `TIP` wording doesn't have to be in the imperative mood like the procedure steps, which might help with making it shorter.
```suggestion
====
The `--delete-namespace` option also removes the {prod-short} namespace.

The `--delete-all` option also removes the {devworkspace} Operator, the related resources, and the {prod-short} namespace.
====

File: modules/administration-guide/pages/uninstalling-che.adoc [hunk 3]
```diff
@@ -20,26 +20,16 @@ To uninstall an instance of {prod} {prod-ver}:
 
 .Procedure
 
-. Obtain the name of the {prod-short} {orch-namespace}. The default is `{prod-namespace}`.
+* Remove the {prod-short} instance from the `{prod-namespace}` {orch-namespace}:
```
max-cx: I suggest getting rid of `{prod-namespace}` here because it might not be necessarily the case (some customers might be using a different namespace), plus it's not important to mention here. (The namespace mattered in the previous version of this procedure because back then it had to be added explicitly, which is no longer the case.)
```suggestion
* Remove the {prod-short} instance:
```

max-cx:
@deerskindoll, I am aware that line 13 was already in the docs and it wasn't part of your changes, but it's worth mentioning here  (1) that we don't normally reference `{prod-ver}` in the docs because it's normally redundant info (the docs are always consistent about one specific version, so it's always `{prod-ver}`) and (2) that `{prod-short}` is more comfortable to read downstream where `{prod}` is quite long (*Red Hat OpenShift Dev Spaces*). Also, senior tech writers have been asking for full-sentence intros.

All in all, I suggest replacing
```
To uninstall an instance of {prod} {prod-ver}:
```
with something like
```
Use `{orch-cli}` to uninstall the {prod-short} instance.
```

max-cx:
Also, I suggest deleting the first prerequisite because it was a prerequisite for a removed step:
```
* An active `{orch-cli}` session with administrative permissions to the destination {orch-name} cluster. See {orch-cli-link}.
```

deerskindoll:
@max-cx feedback incorporated

azatsarynnyy (approved):

modules/administration-guide/pages/uninstalling-che.adoc
```diff
@@ -10,36 +10,23 @@
...
-
-include::example$snip_{project-context}-note-uninstalling-in-minikube-cluster.adoc[]
-
-[NOTE] 
+[TIP]
 ====
-If {prod-short} was installed from the OpenShift web console, `{prod-cli}` will not uninstall the {devworkspace} operator.
-To uninstall the {devworkspace} Operator, see link:https://docs.openshift.com/container-platform/{ocp4-ver}/web_console/odc-about-web-terminal.html#deleting-the-devworkspace-operator-dependency[Deleting the {devworkspace} Operator dependency].
+The `--delete-namespace` option also removes the {prod-short} namespace.
+The `--delete-all` option also removes the {devworkspace} Operator, the related resources, and the {prod-short} namespace.
```
max-cx: :) Perhaps try it with a blank line in between (that's why I had it in my suggestion) to break it into two paragraphs?

```suggestion
The `--delete-namespace` option also removes the {prod-short} namespace.

The `--delete-all` option also removes the {devworkspace} Operator, the related resources, and the {prod-short} namespace.
```

modules/administration-guide/pages/uninstalling-che.adoc
```diff
@@ -10,36 +10,24 @@
...
-include::example$snip_{project-context}-note-uninstalling-in-minikube-cluster.adoc[]
-
-[NOTE] 
+[TIP]
 ====
-If {prod-short} was installed from the OpenShift web console, `{prod-cli}` will not uninstall the {devworkspace} operator.
-To uninstall the {devworkspace} Operator, see link:https://docs.openshift.com/container-platform/{ocp4-ver}/web_console/odc-about-web-terminal.html#deleting-the-devworkspace-operator-dependency[Deleting the {devworkspace} Operator dependency].
+The `--delete-namespace` option removes the {prod-short} namespace.
+
+The `--delete-all` option removes the {devworkspace} Operator and the related resources. The {prod-short} namespace isn't removed.
```
max-cx: `The {prod-short} namespace isn't removed.`
Because the subject is missing due to the passive voice, this sentence might be initially misread (in terms of readability) as an expectation or instructions.
Prob the best would be to drop this sentence because the first sentence is clear about what gets removed and namespace removal is specified by the other option in the same admonition. If you still feel strongly about making it explicit, you can rephrase it to state that this option does not remove the namespace (but as I said, imo, it's as redundant here as stating for the other option that it does not remove the {devworkspace} Operator and the related resources).

max-cx (approved):
LGTM with a comment in https://github.com/eclipse-che/che-docs/pull/2491/files#r1023826112.

PR #2495, opened 2022-11-15
fix: Replace Eclipse Theia in examples

max-cx:
This PR updates the page https://www.eclipse.org/che/docs/stable/end-user-guide/selecting-an-in-browser-ide-for-all-workspaces-that-clone-the-same-git-repository/#parameters-for-che-editor-yaml, where Eclipse Theia is replaced by another editor.

https://issues.redhat.com/browse/RHDEVDOCS-4578

`main` and `7.56.x`

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
+++ b/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
@@ -28,23 +28,34 @@ reference: https://__<hostname_and_path_to_a_remote_file>__/che-editor.yaml
 inline:
   schemaVersion: 2.1.0
   metadata:
-    name: theia-ide
-  commands:
-    ...
-  events:
-    ...
+    name: IntelliJ IDEA Community Edition
   components:
-    - name: che-machine-exec
+    - name: intellij
       container:
-        image: 'quay.io/eclipse/che-machine-exec:next'
-        command:
-          ...
-        memoryLimit: 128Mi
+        image: 'quay.io/che-incubator/che-idea@sha256:b02663b2aea521bcbcafdbec7eba27b6d804e3885dc54bb7d5c771ed77d4a97f'
+        volumeMounts:
+          - name: projector-user
+            path: /home/projector-user
+        mountSources: true
+        memoryLimit: 2048M
         memoryRequest: 32Mi
-        cpuLimit: 500m
-        cpuRequest: 30m
-      attributes:
-...
+        cpuLimit: 1500m
+        cpuRequest: 100m
+        endpoints:
+          - name: intellij
+            attributes:
+              type: main
+              cookiesAuthEnabled: true
+              urlRewriteSupported: true
+              discoverable: false
+              path: /?backgroundColor=434343&wss
+            targetPort: 8887
+            exposure: public
+            secure: false
+            protocol: https
+      attributes: {}
+    - name: projector-user
+      volume: {}
 ----
 ====
 
--- a/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
+++ b/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
@@ -28,7 +28,7 @@ reference: https://__<hostname_and_path_to_a_remote_file>__/che-editor.yaml
 inline:
   schemaVersion: 2.1.0
   metadata:
-    name: IntelliJ IDEA Community Edition
+    name: JetBrains IntelliJ IDEA Community IDE
   components:
     - name: intellij
       container:
--- a/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
+++ b/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
@@ -32,7 +32,8 @@ inline:
   components:
     - name: intellij
       container:
-        image: 'quay.io/che-incubator/che-idea@sha256:b02663b2aea521bcbcafdbec7eba27b6d804e3885dc54bb7d5c771ed77d4a97f'
+```suggestion
+        image: 'quay.io/che-incubator/che-idea:next'
         volumeMounts:
           - name: projector-user
             path: /home/projector-user
--- a/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
+++ b/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
@@ -32,7 +32,6 @@ inline:
   components:
     - name: intellij
       container:
-```suggestion
         image: 'quay.io/che-incubator/che-idea:next'
         volumeMounts:
           - name: projector-user
--- a/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
+++ b/modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
@@ -78,7 +78,7 @@ registryUrl: __<url_of_custom_plug-in_registry>__
 ... # <1>
 override:
   containers:
-    - name: intellij-ide
+    - name: che-idea
       memoryLimit: 1280Mi
       cpuLimit: 1510m
       cpuRequest: 102m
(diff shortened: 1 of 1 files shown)


nickboldt:
Seems ok but it would be easier to downstream this if you use a che-code example, as that's the new default (not idea) in both Che and DS. 

Also since you are using che-code on line 10 already... https://github.com/eclipse-che/che-docs/pull/2495/files#diff-54d179fd6ca1bbc0d938388bb0adbac6a9ced7bb9d4e23d368654269f815e6c3R10 ... this would make the full example more internally consistent.

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
```diff
@@ -7,7 +7,7 @@ The simplest way to select an IDE in the `che-editor.yaml` is to specify the `id
 ====
 [source,yaml]
 ----
-id: che-incubator/che-code/insiders
+id: che-incubator/che-idea/latest
```
themr0c, line 10: I believe we should keep `che-code/insiders`. It stays interesting for users to switch to the latest Visual Studio Code version.

max-cx: I'm afraid that `che-incubator/che-code/insiders` is the only VS Code version that we offer both upstream and downstream. If that's the case, then there's no other VS Code version to switch to. And if we could provide an upstream-only VS Code version, that wouldn't help us much here for the downstream docs.

themr0c: Yes, so keep che-code/insiders rather than replacing by che-idea?

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
```diff
@@ -7,7 +7,7 @@ The simplest way to select an IDE in the `che-editor.yaml` is to specify the `id
...
 ----
 ====
```
themr0c, line 12: Example block, then code block

```suggestion
====
----
```

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
max-cx: :) Well, I don't want to confuse readers by showing them here how to specify the *default* editor which, in this particular example, they explicitly don't need to specify.

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
max-cx: :) Thank you, but that's the closing syntax, not the opening syntax.

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
themr0c: I believe there is a misunderstanding:

Che (and Dev Spaces) (hopefully) runs by default on a released version of the `che-code` container, such as `che-incubator/che-code/7.56.0`.

The user can switch to the development version of the `che-doc` container: `che-incubator/che-code/insiders`.

See the tag history in quay: https://quay.io/repository/che-incubator/che-code?tab=tags

Therefore, the "Switch to the development version of the Visual Studio Code editor" example still makes sense.

Moreover, the new first class citizen replacing Che-Theia is Visual Studio Code, not IntelliJ, therefore examples should focus on Visual Studio Code as much as possible.

themr0c: After internal discussion:  `che-incubator/che-idea/latest` is the right choice.

themr0c (approved):

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
```diff
@@ -28,23 +28,34 @@ reference: https://__<hostname_and_path_to_a_remote_file>__/che-editor.yaml
...
+    name: IntelliJ IDEA Community Edition
   components:
-    - name: che-machine-exec
+    - name: intellij
       container:
-        image: 'quay.io/eclipse/che-machine-exec:next'
-        command:
-          ...
-        memoryLimit: 128Mi
+        image: 'quay.io/che-incubator/che-idea@sha256:b02663b2aea521bcbcafdbec7eba27b6d804e3885dc54bb7d5c771ed77d4a97f'
```
nickboldt: why pick a specific sha? you could use `:next` or `:2021.3-next` here...

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
```diff
@@ -28,23 +28,34 @@ reference: https://__<hostname_and_path_to_a_remote_file>__/che-editor.yaml
 inline:
   schemaVersion: 2.1.0
   metadata:
-    name: theia-ide
-  commands:
-    ...
-  events:
-    ...
+    name: IntelliJ IDEA Community Edition
```
nickboldt: This is not the officially sanctioned product name as seen in https://issues.redhat.com/browse/CRW-3098?focusedCommentId=21141809&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21141809

See https://github.com/redhat-developer/devspaces/pull/829/files#diff-14148f60b7e8d725c2f196edeb04c6b4750e411561536ecccf55fad177152222R165-R166
and
https://github.com/eclipse-che/che-plugin-registry/pull/1441/files#diff-4526f5cc968bddb01b7aaa7b95bbf3be602a7cfe9b7a69230ecbef43751732b4R640

Thanks!

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
nickboldt, line 10: upstream this could be `/next` but downstream we only have `/latest`.

nickboldt (changes requested):
some changes required; some suggestions to improve too

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
max-cx: Great. Thank you for sharing the link.
IMO, two options here:
`IntelliJ IDEA Community` (taken from the display name)
or
`JetBrains IntelliJ IDEA Community IDE for {prod}` (taken from the description)
Which one do you prefer?

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
max-cx: Apologies for that. I just plagiarized a file from Artem's test repo.
Can I rewrite it like this?
```Suggestion
```suggestion
        image: 'quay.io/che-incubator/che-idea:next'
```

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
nickboldt: If you use `JetBrains IntelliJ IDEA Community IDE` you'll make the brand/trademark people happier. 

And by not using the `for {prod}' suffix (upstream) or `{prod} with` prefix (downstream), transformation of doc from up to down will be simpler.

nickboldt:
```suggestion
    name: JetBrains IntelliJ IDEA Community IDE
```

nickboldt (approved):
+1, with one last suggestion to apply

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
max-cx: Thank you, Nick!

modules/end-user-guide/partials/ref_parameters-for-che-editor-yaml.adoc
```diff
@@ -28,23 +28,35 @@ reference: https://__<hostname_and_path_to_a_remote_file>__/che-editor.yaml
...
+    name: JetBrains IntelliJ IDEA Community IDE
   components:
-    - name: che-machine-exec
+    - name: intellij
       container:
-        image: 'quay.io/eclipse/che-machine-exec:next'
-        command:
-          ...
-        memoryLimit: 128Mi
+```suggestion
```
max-cx:
```suggestion
```

PR #2531, opened 2022-12-21
docs: URL parameter for configuring remotes

dkwon17:
Signed-off-by: dkwon17 <dakwon@redhat.com>

Adds a `URL parameter for configuring project remotes` section in the user guide.

image

This PR is to document the feature added by this PR: https://github.com/eclipse-che/che-dashboard/pull/691

Che 7.59

(This feature was [Backported to Che 7.58](https://github.com/eclipse-che/che-dashboard/pull/698), therefore can be available for DS 3.4)

`main` and cherry-pick to `7.58`

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


ibuziuk (approved):

ibuziuk:
@max-cx could you please review?

modules/end-user-guide/pages/url-parameter-for-configuring-project-remotes.adoc
```diff
@@ -0,0 +1,49 @@
...
+:page-aliases:
+
+[id="url-parameter-for-configuring-project-remotes"]
+= URL parameter for configuring project remotes
+
+When you visit a URL for starting a new workspace, one Git remote will be configured by default. Git remotes can be configured manually in the command line after workspace startup, or manually by editing the workspace's {devworkspace} {orch-name} resource after workspace creation. 
+
+To have {prod-short} automatically configure Git remotes on workspace creation, the `remotes=<remotes>` URL parameter can be used.
+
+The first remote specified in the `remotes` parameter will have it's default branch checked out on workspace startup.
```
nickboldt:
```suggestion
The first remote specified in the `remotes` parameter will have its default branch checked out on workspace startup.
```

modules/end-user-guide/pages/url-parameter-for-configuring-project-remotes.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+The value of the `remotes` parameter can be provided in two formats: by providing both remote names and URLs, and by providing only URLs.
```
nickboldt:
```suggestion
The value of the `remotes` parameter can be provided in two formats: by providing both remote names and URLs, or by providing only URLs.
```

modules/end-user-guide/pages/url-parameter-for-configuring-project-remotes.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+Remote URLs can be specified by providing a list of URLs in the following format:
+
+[source,subs="+quotes,+attributes,+macros"]
+----
+{https://git-url/origin/project.git,https://git-url/upstream/project.git, ... }
+----
+
+When remotes are configured in this format, the remote names are implicit. 
+The first remote is `origin`, the second remote is `upstream`, and subsequent remotes are `fork1`, `fork2`, ... ,  `fork(n-2)`.
```
nickboldt: so it doesn't use the user-or-organization path segments for the n-2 and n-1 items, but instead forces names to be `origin`,`upstream` and `forkN` ? 

Hmm. that's an odd choice but I guess if `origin` and `upstream` is a standard convention... then this makes sense.

How hard would it be to pick the organization path segment and use THAT as the remote name? Not sure we need to, just wondering...

modules/end-user-guide/pages/url-parameter-for-configuring-project-remotes.adoc
```diff
@@ -0,0 +1,49 @@
+:_content-type: CONCEPT
+:description: URL parameter for configuring project remotes
+:keywords: configuring-remotes, configuring-remotes-parameter, remotes-parameter, project-remotes-parameter
+:navtitle: URL parameter for configuring project remotes
+:page-aliases:
+
+[id="url-parameter-for-configuring-project-remotes"]
+= URL parameter for configuring project remotes
+
+When you visit a URL for starting a new workspace, one Git remote will be configured by default. Git remotes can be configured manually in the command line after workspace startup, or manually by editing the workspace's {devworkspace} {orch-name} resource after workspace creation.
```
nickboldt: lots of "manually" in this sentence, which is redundant.

```suggestion
When you open a URL to start a new workspace, one Git remote will be configured automatically. Additional Git remotes can then be configured on the command line after workspace startup, or by editing the {devworkspace} {orch-name} resource for the new workspace after its creation. 
```

modules/end-user-guide/pages/url-parameter-for-configuring-project-remotes.adoc
dkwon17: I think this is a nice enhancement that is straight forward to implement, do you mind if I create an issue @nickboldt ?

max-cx: @dkwon17 and @nickboldt, I'm not a developer, but I find it confusing for a few reasons:
1. I have a number of remotes in my git, and most of them are named according to their default names when they get added. For example, one of my remotes is named `dkwon17`. So as a user, I'd like to operate with whatever names those remotes get by default (that's already sufficient for me to be able to identify them in the editor of the workspace).
2. Whatever work I am doing in a local clone is normally to be pulled from or pushed to `origin`, which in this case is the main repo: `pass:c,a,q[{prod-url}]#__<git_repository_url>__`. so I'd find it confusing to have an additional remote named `origin`.
3. When I read that the `origin` remote's default branch is checked out, the question that popped up in my head is: and what about the other additional remote, which branch is checked out there?

max-cx: I started editing a new branch to open a PR against the branch of this PR, but now it's not clear to me whether we should wait with this update of the docs if there is to be any iteration of this feature. Should this PR be **converted to draft** in the meantime until the documented feature is updated?

max-cx: Regarding examples, I'd like to see (but am not competent to suggest here) some real-life repos, for example:
```
pass:c,a,q[{prod-url}]#https://github.com/golang/example.git?remotes={{golang,https://github.com/golang/tools.git},{godoctor,https://github.com/godoctor/godoctor.git}}
```

dkwon17: 1. I've created an issue for the enhancement [here](https://github.com/eclipse/che/issues/21943), but that is still possible by providing explicit remote names to the `remotes` parameter (ie: `?remotes={{dkwon17,https://github.com/dkwon17/tools.git}}`)

2. The origin branch in the `remotes` parameter will actually replace the original one, so at the end, there is still one `origin` remote.

3. I'm not sure if I understand the question, but if another remote is to be checked out and no branch is specified, it would be the default branch for that remote too.

I can update this PR for more clarity

max-cx: @dkwon17, no need to update your PR yet. I can complete my review without problems and you can see what you want to keep from my proposed changes.

To clarify regarding

> 2. The origin branch in the remotes parameter will actually replace the original one, so at the end, there is still one origin remote.

Does `https://<che_fqdn>#<git_repository_url>` from https://www.eclipse.org/che/docs/stable/end-user-guide/starting-a-new-workspace-with-a-clone-of-a-git-repository/ work together with this parameter? If yes, then I assume that repository will also be cloned, correct? So it's only about which of these repos gets named `origin`, correct? If that's the case, maybe we should not document the URL-only syntax option for now, and only document {name,URL} syntax? WDYT, the new page gets published just without the URL-only syntax option (until the new enhancement issue is resolved)?

dkwon17: Yes, `https://<che_fqdn>#<git_repository_url>` from https://www.eclipse.org/che/docs/stable/end-user-guide/starting-a-new-workspace-with-a-clone-of-a-git-repository/ work together with this parameter.

>  If yes, then I assume that repository will also be cloned, correct? 

Yes, but only if `origin` remote has not been provided in in the `remotes` parameter

When accessing this URL:
```
https://<che_fqdn>#https://github.com/eclipse/che?remotes={{origin,https://github.com/other-user/che}}
```
We get the project devfile from `https://github.com/eclipse/che` and initially set the `origin` remote to `https://github.com/eclipse/che`. But since the `remotes` parameter provides a new `origin` remote, it will replace the old `origin` remote (`https://github.com/eclipse/che`) with the new one (`https://github.com/other-user/che`). This happens before any project is cloned.

In other words, in this case the devfile comes from `https://github.com/eclipse/che`, but the `origin` remote (which is also the project that we clone) is `https://github.com/other-user/che`.

This is something I should have documented to make it more clear, I'd like to do so after your review. There are also a slight change introduced by https://github.com/eclipse-che/che-dashboard/pull/704 that I would like to add to the docs.

In the interest of making the docs minimal, I'm ok with only documenting the `{name,URL}` syntax until we prioritize https://github.com/eclipse/che/issues/21943 next sprint (Jan 25 - Feb 14). After prioritization (or after the enhancement is finished), I can provide a new PR to document it again

dkwon17: For some concrete examples:

#### 1. `origin` remote not provided in `remotes` url param
```
{prod-url}#https://github.com/eclipse-che/che-docs.git?remotes={{max-cx,https://github.com/max-cx/che-docs.git},{dkwon17,https://github.com/dkwon17/example.git}}
```

Creates a workspace with these remotes configured:
```
origin  -> https://github.com/eclipse-che/che-docs.git
max-cx  -> https://github.com/max-cx/che-docs.git
dkwon17 -> https://github.com/dkwon17/che-docs.git
```

#### 2. `origin` remote provided in `remotes` url param
```
{prod-url}#https://github.com/eclipse-che/che-docs.git?remotes={{origin,https://github.com/max-cx/che-docs.git},{dkwon17,https://github.com/dkwon17/che-docs.git}}
```

Creates a workspace with these remotes configured:
```
origin  -> https://github.com/max-cx/che-docs.git
dkwon17 -> https://github.com/dkwon17/che-docs.git
```

In both cases, the origin/main branch will be checked out on workspace startup

ibuziuk:
@max-cx can we merge the PR now?

modules/end-user-guide/pages/url-parameter-for-configuring-project-remotes.adoc
max-cx: @dkwon17, regarding your two examples above, does example 2 mean that `#https://github.com/eclipse-che/che-docs.git?` will not be cloned for the workspace?

dkwon17: Yes, for example 2,  `https://github.com/eclipse-che/che-docs.git` will not be cloned for the workspace, and ` https://github.com/max-cx/che-docs.git` will be cloned instead

max-cx (changes requested):
I've reviewed this PR and propose a rewrite in https://github.com/dkwon17/che-docs/pull/15.

max-cx:
Also, I'd like to suggest documenting the other way of doing this (without specifying the remote names) as an admonition like TIP. Please correct me if I am wrong, but my current understanding is that explicitly specifying the names of the additional remotes manually by the user offers more flexibility in terms of which remote becomes the origin and is checked out, which might cover more use cases. If that is true, then that would be the number one choice in the docs. And I wouldn't recommend offering readers multiple ways of doing the same thing as separate and parallel sections in the docs.

dkwon17:
@max-cx yes, documenting the other way of doing this as an admonition/TIP sounds good to me

max-cx (approved):

ibuziuk (approved):

modules/end-user-guide/pages/url-parameter-for-additional-remotes.adoc
```diff
@@ -0,0 +1,25 @@
...
+[source,subs="+quotes,+attributes,+macros"]
+----
+pass:c,a,q[{prod-url}]#__<git_repository_url>__?remotes={{__<name_1>__,__<url_1>__},{__<name_2>__,__<url_2>__},{__<name_3>__,__<url_3>__},...}
+----
+
+[IMPORTANT]
+====
+If you do not enter the name `origin` for any of the additional remotes, the remote from __<git_repository_url>__ will be cloned and named `origin` by default, and its expected branch will be checked out automatically.
+
+If you enter the name `origin` for one of the additional remotes, its default branch will be checked out automatically, but the remote from __<git_repository_url>__ will NOT be cloned for the workspace.
```
max-cx:
```suggestion

* If you do not enter the name `origin` for any of the additional remotes, the remote from __<git_repository_url>__ will be cloned and named `origin` by default, and its expected branch will be checked out automatically.

* If you enter the name `origin` for one of the additional remotes, its default branch will be checked out automatically, but the remote from __<git_repository_url>__ will NOT be cloned for the workspace.

```

PR #2534, opened 2023-01-05
fix: remove reference to "supported browsers"

osslate:

Removes references to specific browsers as "supported" (Chrome and Firefox).

* https://issues.redhat.com/browse/RHDEVDOCS-4754

7.58

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/pages/benefits-of-pull-requests-review-in-che.adoc
+++ b/modules/end-user-guide/pages/benefits-of-pull-requests-review-in-che.adoc
@@ -12,7 +12,6 @@
 
 .Prerequisites
 * You have access to the repository hosted by your Git provider.
-* You use a {prod}-supported browser: Google Chrome or Mozilla Firefox.
 * You have access to a {prod-short} instance.
 
 .Procedure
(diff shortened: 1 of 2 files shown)


modules/end-user-guide/pages/benefits-of-pull-requests-review-in-che.adoc
```diff
@@ -12,7 +12,6 @@
 
 .Prerequisites
 * You have access to the repository hosted by your Git provider.
-* You use a {prod}-supported browser: Google Chrome or Mozilla Firefox.
```
osslate, line 15: @max-cx is this line something we want to remove entirely? I was torn between removing it, or replacing it with something like

> You have access to a web browser.

max-cx: The way I see it, having a browser is the number one prerequisite for using this product. So if we keep it as an explicit prerequisite in this procedure or any other, it might as well be added to every single procedure in the user guide. Also, I remember that the minimalism and modular docs advice is against listing obvious requirements in the prerequisites list. So you don't have to worry about keeping it.

max-cx (approved):

max-cx:
Merging this without an SME review because this was already discussed with and requested by the SMEs.

max-cx:
Backported to `7.58.x`

PR #2562, opened 2023-02-23
fix: removing che theia from troubleshooting slow workspaces (rhdevdocs-4580)

deerskindoll:
removing che theia from troubleshooting slow workspaces

https://issues.redhat.com/browse/RHDEVDOCS-4572

devspaces 3.6, eclipse che 7.63

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


ibuziuk (commented):
LGTM @azatsarynnyy @RomanNikitenko folks, not really sure though if the same instructions are valid for che-code

azatsarynnyy:
> LGTM @azatsarynnyy @RomanNikitenko folks, not really sure though if the same instructions are valid for che-code

@ibuziuk What is related to the `Reducing the number of public endpoints` section is valid for Che-Code as well
https://github.com/eclipse-che/che-plugin-registry/blob/aea518ad4b8773534569a6bc8c640cf207e15063/che-editors.yaml#L212-L232

But `CDN configuration` is related to Che-Theia only. So, it can be removed.
@deerskindoll ^^

Also, I think that the `Improving workspace runtime performance` section should be revised as it mentions `meta.yaml` plugin descriptors that are not relevant for plugin-registry anymore.

ibuziuk (commented):
@deerskindoll the following section should be removed:

```
Optimizing workspace plugins::
+
_Role: User_
+
When selecting various plugins, each plugin can bring its own sidecar container, which is an OCI image. {orch-name} pulls the images of these sidecar containers.
+
Reduce the number of plugins, or disable them to see if start time is faster. See also xref:administration-guide:caching-images-for-faster-workspace-start.adoc[].
Reducing the number of public endpoints::
```

ibuziuk:
@deerskindoll CDN configuration should be removed

```
CDN configuration::
+
The IDE editor uses a CDN (Content Delivery Network) to serve content. Check that the content uses a CDN to the client (or a local route for offline setup).
+
To check that, open Developer Tools in the browser and check for `vendors` in the *Network* tab. `vendors.<random_id>.js` or `editor.main.*` should come from CDN URLs.
```

azatsarynnyy (approved):
LGTM

svor (approved):
LGTM

modules/end-user-guide/pages/troubleshooting-slow-workspaces.adoc
```diff
@@ -70,52 +55,31 @@ Ensure the CPU settings in the devfile definition, `devfile.yaml`, are correct:
...
-      memoryLimit: "512Mi" <1>
-  extensions:
-    - https://link.to/vsix
 ----
-<1> Specifies the memory limit for the plugin.
-+
-In the devfile definition (`devfile.yaml`):
-+
 [source,yaml]
 ----
```
aireilly, line 75: This code block is not closed.

deerskindoll: it got deleted and won't appear in the new version of the document

aireilly: Are you sure? I see an unclosed codeblock here.

image

deerskindoll: you're right. that one stays.

PR #2564, opened 2023-02-27
fix: update RELEASE.adoc

max-cx:
Update the steps in `RELEASE.adoc`

https://issues.redhat.com/browse/RHDEVDOCS-4975

Merge to `main`

## Pull Request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:

### DIFF
--- a/RELEASE.adoc
+++ b/RELEASE.adoc
@@ -13,23 +13,38 @@ The publication is decoupled. See link:https://github.com/eclipse-che/che-docs/t
 
 .Prerequisites
 
-* Define the Eclipse Che version to release: `__<major>__.__<minor>__.0`.
+* The new release is published on https://github.com/eclipse/che/releases and in https://mattermost.eclipse.org/eclipse/channels/eclipse-che-releases.
 
 .Procedure
 
-. Create the `__<major>__.__<minor>__.x` branch from the prerelease branch.
+. From the `main` branch, create a release branch with a branch name in the format `__<major>__.__<minor>__.x`, and then checkout the created release branch.
++
+NOTE: The name of the release branch differs from the number of the release in the trailing `x`.
 
-. Edit the `antora.yml` file in the `__<major>__.__<minor>__.x` branch to define:
+. Update the following product-version-related values in `antora.yml`:
 +
 [source,yaml,subs="+attributes,+quotes"]
 ----
 prerelease: false
-version: __<major>__.__<minor>__.x
+version: __<major>__.__<minor>__.x # name of release branch
 asciidoc:
   attributes:
     prod-ver-patch: "_<major>_._<minor>_.0"
     prod-ver: "_<major>_._<minor>_"
     prod-ver-major: "_<major>_"
+----
+
+. Look up the latest DWO release in https://github.com/devfile/devworkspace-operator/tags, double-check it with the SMEs, and update the DWO version patch value in `antora.yml`:
++
+[source,yaml,subs="+attributes,+quotes"]
+----
+asciidoc:
+  attributes:
+    ...
+    # for the project
+    ...
+    devworkspace-operator-version-patch: "x.y.z"
+    ...
 ----
 
 . Commit and push the changes to the release branch.
--- a/RELEASE.adoc
+++ b/RELEASE.adoc
@@ -21,7 +21,7 @@ The publication is decoupled. See link:https://github.com/eclipse-che/che-docs/t
 +
 NOTE: The name of the release branch differs from the number of the release in the trailing `x`.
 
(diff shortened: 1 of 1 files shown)


RELEASE.adoc
```diff
@@ -13,25 +13,40 @@ The publication is decoupled. See link:https://github.com/eclipse-che/che-docs/t
 
 .Prerequisites
 
-* Define the Eclipse Che version to release: `__<major>__.__<minor>__.0`.
+* The new release is published on https://github.com/eclipse/che/releases and in https://mattermost.eclipse.org/eclipse/channels/eclipse-che-releases.
```
amisevsk: I'm not sure this should be a prerequisite -- the release published on https://github.com/eclipse/che/releases is tagged from the `<major>.<minor>.x` branch, but this doc outlines when that branch is created.

max-cx: @amisevsk, as far as I understand, this file documents tasks toward publishing a new version of the docs, which is for a new release of the product, Che. The word `release` might be confusing here. This is all purely about the docs repo, https://github.com/eclipse-che/che-docs/.

amisevsk: Ah sorry, I misinterpreted the context -- thanks.

amisevsk (approved):

RELEASE.adoc
max-cx:
```suggestion
* The new release is published on https://github.com/eclipse/che/releases and https://mattermost.eclipse.org/eclipse/channels/eclipse-che-releases.
```

RELEASE.adoc
```diff
@@ -13,25 +13,40 @@ The publication is decoupled. See link:https://github.com/eclipse-che/che-docs/t
...
 
 .Procedure
 
-. Create the `__<major>__.__<minor>__.x` branch from the prerelease branch.
+. From the `main` branch, create a release branch with a branch name in the format `__<major>__.__<minor>__.x`, and then checkout the created release branch.
++
+NOTE: The name of the release branch differs from the number of the release in the trailing `x`.
 
-. Edit the `antora.yml` file in the `__<major>__.__<minor>__.x` branch to define:
+. Update the following product-version-related values in `antora.yml`:
```
max-cx:
```suggestion
. In the `__<major>__.__<minor>__.x` branch, update the following product-version-related values in `antora.yml`:
```

RELEASE.adoc
```diff
@@ -13,25 +13,40 @@ The publication is decoupled. See link:https://github.com/eclipse-che/che-docs/t
...
-version: __<major>__.__<minor>__.x
+version: __<major>__.__<minor>__.x # name of release branch
 asciidoc:
   attributes:
     prod-ver-patch: "_<major>_._<minor>_.0"
     prod-ver: "_<major>_._<minor>_"
     prod-ver-major: "_<major>_"
 ----
 
+. Look up the latest DWO release in https://github.com/devfile/devworkspace-operator/tags, double-check it with the SMEs, and update the DWO version patch value in `antora.yml`:
```
max-cx: This step is likely to be dropped soon because the engineering discovered that this attribute is not needed in the docs and should be removed from the docs altogether (requires a docs update). I opened a Jira for that.

RELEASE.adoc
```diff
@@ -13,25 +13,40 @@ The publication is decoupled. See link:https://github.com/eclipse-che/che-docs/t
...
 
 .Procedure
 
-. Create the `__<major>__.__<minor>__.x` branch from the prerelease branch.
+. From the `main` branch, create a release branch with a branch name in the format `__<major>__.__<minor>__.x`, and then checkout the created release branch.
++
+NOTE: The name of the release branch differs from the number of the release in the trailing `x`.
 
-. Edit the `antora.yml` file in the `__<major>__.__<minor>__.x` branch to define:
+. In the `__<major>__.__<minor>__.x` branch, update the following product-version-related values in `antora.yml`:
```
max-cx, line 47: I think this should be done while still in `main` before creating the release branch because the `main` has to be continuously updated too. So I think I should depart from the original wording of this step.

amisevsk (approved):
:+1: 

Agree on DWO patch version being dropped if possible -- even with a correct major.minor.bugfix, freshmaker builds will use a different version so it's not fully accurate.

RELEASE.adoc
max-cx: I tried doing this while in `main`, but that requires a PR because `main` appears to be protected.
I'm adding a separate step for updating the same values for `main`.

RELEASE.adoc
max-cx: Actually, let me remove it from this doc now.

PR #2567, opened 2023-03-03
fix: improve git PAT configuration documentation

svor:
Updates [Deploying Che with support for Git repositories with self-signed certificates](https://www.eclipse.org/che/docs/stable/administration-guide/deploying-che-with-support-for-git-repositories-with-self-signed-certificates/) and Using a [Git-provider access token](https://www.eclipse.org/che/docs/stable/end-user-guide/using-a-git-provider-access-token/) articles

These PR provides a few updates related to (_taken from the related issue_):
-  Retrieving BB user id (for the annotation che.eclipse.org/scm-userid) is not straightforward. That's because using the curl command that is linked in the documentation fails with non-authorized. A link to the [authentication for BB API](https://developer.atlassian.com/server/bitbucket/how-tos/example-basic-authentication/) should be added too.

-  I didn't know exactly what URL to use: `https://{bb-host}/users/{user-slug}/repos/{repo-name}/` didn't work but `https://{bb-host}/scm/~{user-slug}/{repo-name}.git`. **We should support both BB URL schemas**.

-  I initially set che.eclipse.org/scm-userid: 2 because BB user id is an integer. But the secret failed to be parsed because a string is expected. We should document that the scm-userid should be wrapped with "" or ''.

-  I didn't know exactly what URL to use: https://{bb-host}/users/{user-slug}/repos/{repo-name}/ didn't work but https://{bb-host}/scm/~{user-slug}/{repo-name}.git. We should support both BB URL schemas.

-  After successfully setting up the PAT secret, git clone continued to fail because BB was using an untrusted SSL certificate. The article "[Deploying Che with support for Git repositories with self-signed certificates](https://www.eclipse.org/che/docs/stable/administration-guide/deploying-che-with-support-for-git-repositories-with-self-signed-certificates/)" should be referenced as an "additional resource".

-  As some point the configmap is referenced as a secret ("All Secrets that hold certificate files...")

-  It's not clear if certs in binary format are supported or not (they are not supported but this sentence is confusing "Also, they can be encoded as binary data, for example, .cer...")

-   It doesn't mention that the certificate chain is required

-   It doesn't mention that in the githost the protocol (i.e. https) is required

https://github.com/eclipse/che/issues/21989
https://issues.redhat.com/browse/RHDEVDOCS-4976

Che 7.62.0
DS 3.6.0
Merge to `main`
Cherry-pick to `7.62.x`, `7.63.x`, `7.64.x`

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:


modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
```diff
@@ -59,7 +59,7 @@ If you have the link:https://www.gnu.org/software/coreutils/base64[base64] comma
 +
 * GitHub: link:https://docs.github.com/en/rest/users/users#get-a-user[Get a user]. See the `id` value in the response.
 * GitLab: link:https://docs.gitlab.com/ee/api/users.html#for-normal-users[List users: For normal users], use the `username` filter: `/users?username=:username`. See the `id` value in the response.
-* Bitbucket Server: link:https://developer.atlassian.com/server/bitbucket/rest/v802/api-group-api/#api-api-latest-users-userslug-get[Get users]. See the `id` value in the response.
+* Bitbucket Server: use the `curl` command: `curl -D- -u username:password -X GET -H "Content-Type: application/json" https://{baseurl}/rest/api/latest/users/{userSlug}` and see the `id` value in the response. More information could be found in the link:https://developer.atlassian.com/server/bitbucket/rest/v802/api-group-api/#api-api-latest-users-userslug-get[Get users BitBucket documentation] and the link:https://developer.atlassian.com/server/bitbucket/how-tos/example-basic-authentication/[authentication for BitBucket API].
```
max-cx: @svor,  in `https://{baseurl}/rest/api/latest/users/{userSlug}`,
`{baseurl}` and `{userSlug}` are some environment variables or are these placeholders for some values that the reader has to enter here manually?

svor: @max-cx these placeholders for the user's input
It was taken from https://developer.atlassian.com/server/bitbucket/rest/v802/api-group-api/#api-api-latest-users-userslug-get

modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
```diff
@@ -87,7 +87,7 @@ type: Opaque
 <1> Your {prod-short} user ID.
 <2> The Git provider name: `github` or `gitlab` or `bitbucket-server` or `azure-devops`.
 <3> The Git provider URL.
-<4> Your Git provider user ID.
+<4> Your Git provider user ID, the value should be wrapped with ' ' or " ".
```
max-cx:
```suggestion
<4> Your Git provider user ID wrapped with `' '` or `" "`.
```

modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
max-cx: @svor, thank you.

1. I haven't seen a response, so I'm assuming that here I use my user slug value for the API call, and then I locate the `id` field in the response, correct?

2. Looks like our repo tools are failing on Atlassian's markup.
So I wonder if we can switch to our habitual markup here, and also some suggested changes to the wording:

```suggestion
* Bitbucket Server: Get the `id` by running `curl -D- -u username:password -X GET -H "Content-Type: application/json" https://__<baseurl>__/rest/api/latest/users/__<userSlug>__`. See the Bitbucket API documentation about link:https://developer.atlassian.com/server/bitbucket/rest/v802/api-group-api/#api-api-latest-users-userslug-get[using the user slug] and link:https://developer.atlassian.com/server/bitbucket/how-tos/example-basic-authentication/[authentication].
```

svor: 1. Correct
2. Looks good, applying your suggestion, thanks!

modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
```diff
@@ -87,7 +87,7 @@ type: Opaque
 <1> Your {prod-short} user ID.
 <2> The Git provider name: `github` or `gitlab` or `bitbucket-server` or `azure-devops`.
 <3> The Git provider URL.
-<4> Your Git provider user ID.
+<4> Your Git provider user ID wrapped with `' '` or `" "`.
```
max-cx:
```suggestion
<4> Your Git provider user ID wrapped in `' '` or `" "`.
```

modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
max-cx: Thank you.

max-cx (approved):

max-cx:
Cherry-picked to `7.62.x`.
No need to cherry-pick to `7.63.x` and `7.64.x` because those branches are not created yet, so the merged changes will get there from main.

PR #2568, opened 2023-03-07
docs: troubleshooting webview loading error (rhdevdocs-4803)

deerskindoll:
new troubleshooting page about webview loading error in incognito/private mode

https://issues.redhat.com/browse/RHDEVDOCS-4803

7.60

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
+++ b/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
@@ -6,7 +6,7 @@
 
 [id="troubleshooting-webview-loading-error"]
 
-= Troubleshooting Webview loading error
+= Troubleshooting webview loading error
 
 pass:[<!-- vale RedHat.Spelling = NO -->]
 
@@ -15,7 +15,7 @@ If you use Microsoft Visual Studio Code - Open Source in a private browsing wind
 
 [WARNING]
 ====
-Mozilla Firefox Private Browsing mode link:https://bugzilla.mozilla.org/show_bug.cgi?id=1320796[does not support ServiceWorkers] necessary for Microsoft Visual Studio Code - Open Source to load Webviews. There is no workaround available.
+Mozilla Firefox Private Browsing mode link:https://bugzilla.mozilla.org/show_bug.cgi?id=1320796[does not support ServiceWorkers] necessary for Microsoft Visual Studio Code - Open Source to load webviews. There is no workaround available.
 ====
 
 .Fixing Webview loading error in Google Chrome Incognito mode
--- a/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
+++ b/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
@@ -1,7 +1,7 @@
 :_content-type: PROCEDURE
-:description: Troubleshooting Webview loading error
+:description: Troubleshooting webview loading error
 :keywords: user-guide, troubleshooting-webview-loading-error
-:navtitle: Troubleshooting Webview loading error
+:navtitle: Troubleshooting webview loading error
 :page-aliases: .:troubleshooting-webview-loading-error.adoc
 
 [id="troubleshooting-webview-loading-error"]
--- a/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
+++ b/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
@@ -1,25 +1,25 @@
 :_content-type: PROCEDURE
-:description: Troubleshooting webview loading error
+:description: Troubleshooting Webview loading error
 :keywords: user-guide, troubleshooting-webview-loading-error
-:navtitle: Troubleshooting webview loading error
+:navtitle: Troubleshooting Webview loading error
 :page-aliases: .:troubleshooting-webview-loading-error.adoc
 
 [id="troubleshooting-webview-loading-error"]
-= Troubleshooting webview loading error
+= Troubleshooting Webview loading error
 
-If you use Microsoft Visual Studio Code - Open Source in an incognito or private browser window, you can encounter an issue with webviews. The webview is not loaded and an error message appears: *Error loading webview: Error: Could not register service workers*.
+If you use Microsoft Visual Studio Code - Open Source in an incognito or private browser window, you can encounter an issue with Webviews. The Webview is not loaded and an error message appears: *Error loading webview: Error: Could not register service workers*.
 
-The issue affects the following browers:
+The issue affects the following browsers:
 
 * Google Chrome in Incognito mode
 * Mozilla Firefox in Private Browsing mode
 
 [WARNING]
 ====
-Mozilla Firefox Private Browsing mode link:https://bugzilla.mozilla.org/show_bug.cgi?id=1320796[does not support ServiceWorkers] necessary for Microsoft Visual Studio Code - Open Source to load webviews. There is no workaround available.
+Mozilla Firefox Private Browsing mode link:https://bugzilla.mozilla.org/show_bug.cgi?id=1320796[does not support ServiceWorkers] necessary for Microsoft Visual Studio Code - Open Source to load Webviews. There is no workaround available.
 ====
 
-.Fixing webview loading error in Google Chrome Incognito mode
+.Fixing Webview loading error in Google Chrome Incognito mode
 
 Change the cookie settings in Google Chrome to fix the issue.
 
--- a/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
+++ b/modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
@@ -5,6 +5,9 @@
 :page-aliases: .:troubleshooting-webview-loading-error.adoc
 
 [id="troubleshooting-webview-loading-error"]
+
+pass:[<!-- vale RedHat.Spelling = NO -->]
+
 = Troubleshooting Webview loading error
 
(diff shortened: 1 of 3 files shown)


azatsarynnyy (approved):
LGTM
Thanks, @deerskindoll for updating the `Troubleshooting` chapter.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
```diff
@@ -0,0 +1,42 @@
+:_content-type: PROCEDURE
+:description: Troubleshooting Webview loading error
+:keywords: user-guide, troubleshooting-webview-loading-error
+:navtitle: Troubleshooting Webview loading error
+:page-aliases: .:troubleshooting-webview-loading-error.adoc
+
+[id="troubleshooting-webview-loading-error"]
+
+= Troubleshooting Webview loading error
```
max-cx, line 9: Looks like `webview` is lower-case in sentences and sentence capitalization:
https://code.visualstudio.com/api/ux-guidelines/webviews
https://code.visualstudio.com/api/extension-guides/webview

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+pass:[<!-- vale RedHat.Spelling = NO -->]
+
+If you use Microsoft Visual Studio Code - Open Source in an incognito or private browser window, you can encounter an issue with Webviews. The Webview is not loaded and an error message appears: *Error loading webview: Error: Could not register service workers*.
```
max-cx: If you want, you can shorten it even further and not have to repeat the same info in text and in the quoted error message (which seems to be good enough and self-explanatory):
```suggestion
If you use Microsoft Visual Studio Code - Open Source in a private browsing window, you might encounter the following error message: *Error loading webview: Error: Could not register service workers*.
```

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+The issue affects the following browsers:
+
+* Google Chrome in Incognito mode
+* Mozilla Firefox in Private Browsing mode
```
max-cx: BTW, there are private browsing modes in other browsers:
* https://blogs.opera.com/news/2014/10/how-to-open-private-window-opera-for-computers/
* https://support.microsoft.com/en-us/microsoft-edge/browse-inprivate-in-microsoft-edge-e6f47704-340c-7d4f-b00d-d0cf35aa1fcc

If the error might appear in any other browser, then it might be worth to consider removing this list if the SME agrees. What does the SME think?

```suggestion
```

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+[WARNING]
```
max-cx: It find it a bit confusing: the warning is about Mozilla and the procedure is about Chrome.
Let me think if I can suggest some alternative.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
```diff
@@ -0,0 +1,42 @@
...
+.Fixing Webview loading error in Google Chrome Incognito mode
+
+Change the cookie settings in Google Chrome to fix the issue.
+
+.Procedure
+
+. Open *Settings*.
+. Select *Privacy and security*.
+. Click *Cookies and other site data*.
+. Select *Allow all cookies*.
```
max-cx: I assume the reader still has the option of not using a private window, so it'd be worth letting the reader know that not using a private window solves this problem too. Maybe something to confirm with the SME.

Another way of presenting all of this information is to turn it into a table like this, but this obviously calls for an SME's review:

```suggestion
.Dealing with the webview error in a private browsing window

[cols="25h,~"]
|===
| Browser | Workarounds

| Google Chrome
|
* Go to *Settings* -> *Privacy and security* -> *Cookies and other site data* -> *Allow all cookies*.

* Do not use an Incognito window.

| Mozilla Firefox
|
* Do not use a private window.

* See link:https://bugzilla.mozilla.org/show_bug.cgi?id=1320796[Bug 1320796: Support ServiceWorkers in Private Browsing Mode].

|===

```

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
azatsarynnyy: We just faced the problem in these two browsers. We didn't check it for other browsers.
That's the reason why the specific browsers are mentioned.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
azatsarynnyy: just to clarify: the idea is that there's no workaround for Firefox, but it exists for Chrome.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
azatsarynnyy: > I assume the reader still has the option of not using a private window, so it'd be worth letting the reader know that not using a private window solves this problem too. Maybe something to confirm with the SME.

That's correct. As it's mentioned at the beginning of this doc, the issue is relevant for private/incognito modes only.

azatsarynnyy: I like the idea of presenting workarounds as a table :+1:

azatsarynnyy:
Thanks, @max-cx for reviewing the content. I addressed some of your comments.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
max-cx: I guess stating specific browsers can be interpreted that other browsers that are not mentioned (have been tested and) are not affected. That's why I suggested removing this list.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
max-cx: Yes, yes. I got that. Thank you.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
azatsarynnyy: I see your point @max-cx. Thanks for the clarification.
I agree - let's remove the browsers' names.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
azatsarynnyy: +1

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
azatsarynnyy: agree, it sounds better with your suggestion @max-cx 
Thank you!

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
deerskindoll: it is but Vale says otherwise atm. I'll have to update the error rules first.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
deerskindoll: I see your point but I think not using a private window in Firefox isn't really a workaround. That's why I used a warning instead. 
I implemented your suggestion to get rid of the list but I think I should keep it, actually. Maybe rephrase it a little bit to emphasize we don't know if other browsers have the same issue. Because a table with workarounds implies the same as a list, doesn't it? That we have tested the browsers and we have solutions for the error for each browser.

deerskindoll: Update: the incognito mode does more than just block cookies so even if you enable cookies, other things such as site data still won't be saved.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
deerskindoll: I updated the wording, please let me know if it's ok with you now.

max-cx (approved):
Pre-approving if I am not available when you need this approval.

modules/end-user-guide/pages/troubleshooting-webview-loading-error.adoc
```diff
@@ -0,0 +1,38 @@
+:_content-type: PROCEDURE
+:description: Troubleshooting webview loading error
+:keywords: user-guide, troubleshooting-webview-loading-error
+:navtitle: Troubleshooting webview loading error
+:page-aliases: .:troubleshooting-webview-loading-error.adoc
+
+[id="troubleshooting-webview-loading-error"]
+
+pass:[<!-- vale RedHat.Spelling = NO -->]
```
max-cx, line 11: These two lines, 8 and 9, are probably causing the rendering issue.

max-cx: Can you move `pass:[<!-- vale RedHat.Spelling = NO -->]` to a line after 11?

max-cx: And maybe try to have no empty lines between the id line and the title line?

PR #2620, opened 2023-07-05
procedures: Add caution against installing operators in separate namespaces

amisevsk:
Add a warning to the installation docs to warn against installing the {prod} Operator in a different namespace than the {devworkspace} Operator (operator and any operators it depends on must be in the same namespace).

In addition, update the {devworkspace} variable to reflect accurate name for DevWorkspace (no space between Dev Workspace).

[CRW-4522](https://issues.redhat.com/browse/CRW-4522)

next

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/installing-che-on-openshift-using-the-web-console.adoc
+++ b/modules/administration-guide/pages/installing-che-on-openshift-using-the-web-console.adoc
@@ -24,6 +24,8 @@ If you have trouble xref:installing-che-on-openshift-using-cli.adoc[installing {
 . Install the {prod} Operator.
 +
 TIP: See link:https://docs.openshift.com/container-platform/{ocp4-ver}/operators/admin/olm-adding-operators-to-cluster.html#olm-installing-from-operatorhub-using-web-console_olm-adding-operators-to-a-cluster[Installing from OperatorHub using the web console].
++
+CAUTION: The {prod} Operator depends on the {devworkspace} Operator. If you install the {prod} Operator manually to a non-default namespace, ensure that the {devworkspace} Operator is also installed in the same namespace. This is required as the Operator Lifecycle Manager will attempt to install the {devworkspace} Operator as a dependency within the {prod} Operator namespace, potentially resulting in two conflicting installations of the {devworkspace} Operator if the latter is installed in a different namespace.
 
 . Create the `{prod-namespace}` project in OpenShift as follows:
 +
--- a/antora.yml
+++ b/antora.yml
@@ -26,7 +26,7 @@ asciidoc:
     che-plugin-registry-directory: che-plugin-registry
     devworkspace-operator-index-disconnected-install: quay.io/devfile/devworkspace-operator-index:release-digest
     devworkspace-operator-version-patch: "0.21.1"
-    devworkspace: Dev Workspace
+    devworkspace: DevWorkspace
     devworkspace-id: devworkspace
     docker-cli: docker
     hosted-che-docs: xref:hosted-che:hosted-che.adoc[]


l0rd:
@amisevsk would it be possible to briefly explain why this restriction is necessary? And this has an impact on the OpenShift Web Terminal too right?

amisevsk:
@l0rd I've added some detail here while trying to keep the caution succinct. You are right that the Web Terminal Operator also has this requirement (resulting in the requirement that all three operators (Che, DevWorkspace, Web Terminal) are installed in the same namespace. However, I'm leaving Web Terminal out of these docs to avoid cross referencing potentially irrelevant operators. The Web Terminal documentation does mention that the DevWorkspace Operator is a dependency, but does not contain a caution about conflicting namespaces (though it only documents the UI flow, where it is hard to install to another namespace).

l0rd (approved):

l0rd:
> However, I'm leaving Web Terminal out of these docs to avoid cross referencing potentially irrelevant operators.

I think this is out of scope of this PR. But I am wondering: when an admin installs Che/Dev Spaces on a cluster where WTO has been previously installed, shouldn't we warn him that the installation namespace needs to match WTO namespace?

modules/administration-guide/pages/installing-che-on-openshift-using-the-web-console.adoc
```diff
@@ -24,6 +24,8 @@ If you have trouble xref:installing-che-on-openshift-using-cli.adoc[installing {
 . Install the {prod} Operator.
 +
 TIP: See link:https://docs.openshift.com/container-platform/{ocp4-ver}/operators/admin/olm-adding-operators-to-cluster.html#olm-installing-from-operatorhub-using-web-console_olm-adding-operators-to-a-cluster[Installing from OperatorHub using the web console].
++
+CAUTION: The {prod} Operator depends on the {devworkspace} Operator. If installing the {prod} Operator manually to a non-default namespace, ensure that the {devworkspace} Operator is also installed in the same namespace. This is required as the Operator Lifecycle Manager will attempt to install the {devworkspace} Operator as a dependency within the {prod} Operator namespace, potentially resulting in two conflicting installations of the {devworkspace} Operator if the latter is installed in a different namespace.
```
deerskindoll:
```suggestion
CAUTION: The {prod} Operator depends on the {devworkspace} Operator. If you install the {prod} Operator manually to a non-default namespace, ensure that the {devworkspace} Operator is also installed in the same namespace. This is required as the Operator Lifecycle Manager will attempt to install the {devworkspace} Operator as a dependency within the {prod} Operator namespace, potentially resulting in two conflicting installations of the {devworkspace} Operator if the latter is installed in a different namespace.
```

deerskindoll (commented):
a small suggestion

amisevsk:
> when an admin installs Che/Dev Spaces on a cluster where WTO has been previously installed, shouldn't we warn him that the installation namespace needs to match WTO namespace?

This would be ideal, though it's hard to figure out the best way to document it -- technically the requirement applies to any operators which have dependencies and should be part of the OpenShift documentation (i.e. there should be no articles suggesting separate namespaces without mentioning this caveat). Documenting it for the case of WTO in a way that isn't confusing or requiring a long digression into how OLM works is the sticking point.

modules/administration-guide/pages/installing-che-on-openshift-using-the-web-console.adoc
amisevsk: Thanks!

deerskindoll (approved):
lgtm

PR #2624, opened 2023-07-12
Docs: Adding a step to verify Authorized Applications

satyamburhade:
Docs: Adding a step to verify Authorized Applications

Verify that the Application is Authorized, Click your avatar and go to Edit Profile → Applications → Authorized Applications

It has been observed that users skip checking whether the Application is authorized or not, Which causes OAuth failure

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/partials/proc_setting-up-the-gitlab-authorized-application.adoc
+++ b/modules/administration-guide/partials/proc_setting-up-the-gitlab-authorized-application.adoc
@@ -42,6 +42,8 @@ $ echo -n '__<gitlab_application_id>__' | base64
 $ echo -n '__<gitlab_client_secret>__' | base64
 ----
 
+. Verify that the Application is Authorized, Click your avatar and go to *Edit Profile* → *Applications* → *Authorized Applications*
+
 .Additional resources
 
 * link:https://docs.gitlab.com/ee/integration/oauth_provider.html#authorized-applications[GitLab Docs: Authorized applications]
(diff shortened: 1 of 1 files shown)


amisevsk (approved):

modules/administration-guide/partials/proc_setting-up-the-gitlab-authorized-application.adoc
```diff
@@ -42,6 +42,8 @@ $ echo -n '__<gitlab_application_id>__' | base64
 $ echo -n '__<gitlab_client_secret>__' | base64
 ----
 
+. Verify that the Application is Authorized, Click your avatar and go to *Edit Profile* → *Applications* → *Authorized Applications*
```
ibuziuk:
```suggestion
. Verify that the application is authorized by clicking on your avatar and selecting *Edit Profile* → *Applications* → *Authorized Applications*
```

vinokurig (changes requested):
This is not relevant for the Admin which setups the application. The `Authorized applications` list shows applications authorised for the specific user, but not for the Admin.

satyamburhade:
Hello @vinokurig, 
Yes, I agree but even when a person who is configuring and testing an OAuth connection will have an idea to check and verify the step.

vinokurig:
> but even when a person who is configuring and testing an OAuth connection will have an idea to check and verify the step.

This should be taken out of the scope. May be add `verification steps` like here https://www.eclipse.org/che/docs/stable/administration-guide/importing-untrusted-tls-certificates/

satyamburhade:
@vinokurig I think it's more of a config step because if it's not done then the user will not be able to use OAuth. So if I change it like below will it be okay? WDYT?

`Make sure that the Application is Authorized, Click your avatar and go to Edit Profile → Applications → Authorized Applications`

vinokurig:
@satyamburhade 
> I think it's more of a config step because if it's not done then the user will not be able to use OAuth.

Why the user will not be able to use the OAuth? What configuration step are we missing?

PR #2645, opened 2023-08-30
docs: adding the token verification note to the 'Using a Git-provider access token' section

ibuziuk:
Adding the token verification note to the 'Using a Git-provider access token' section - https://64ef2edfa59a110c459b58af--eclipse-che-docs-pr.netlify.app/docs/next/end-user-guide/using-a-git-provider-access-token/

image

https://github.com/eclipse/che/issues/22359

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
+++ b/modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
@@ -44,6 +44,22 @@ On OpenShift, you can use the `oc` command-line tool to log in to the cluster:
 +
 [IMPORTANT]
 ====
+Personal access tokens are sensitive information and should be kept confidential. Treat them like passwords. If you are having trouble with authentication, ensure you are using the correct token and have the appropriate permissions for cloning repositories:
+
+. Open a terminal locally on your computer
+. Use the `git` command to clone the repository using your personal access token. The format of the `git` command vary based on the Git Provider. As an example, GitHub personal access token verification can be done using the following command:
+
+----
+git clone https://<PAT>@github.com/username/repo.git
+----
+
+Replace `<PAT>` with your personal access token, and `username/repo` with the appropriate repository path.
+If the token is valid and has the necessary permissions, the cloning process should be successful. Otherwise, this is an indicator of incorrect personal access token, insufficient permissions, or other issues.
+====
+
++
+[IMPORTANT]
+====
 For GitHub Enterprise Cloud, verify that the token is link:https://docs.github.com/en/enterprise-cloud@latest/authentication/authenticating-with-saml-single-sign-on/authorizing-a-personal-access-token-for-use-with-saml-single-sign-on[authorized for use within the organization].
 ====
 


modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
```diff
@@ -41,6 +41,22 @@ On OpenShift, you can use the `oc` command-line tool to log in to the cluster:
...
++
+[IMPORTANT]
+====
+Personal access tokens are sensitive information and should be kept confidential. Treat them like passwords. If you are having trouble with authentication, ensure you are using the correct token and have the appropriate permissions for cloning repositories:
+
+. Open a terminal locally on your computer
+. Use the `git` command to clone the repository using your personal access token. The format of the `git` command vary based on the Git Provider. As an example for GitHub, personal access token verification can be done using the following command:
+
+----
+git clone https://<PAT>@github.com/username/repo.git
```
vinokurig, line 53: This pattern doesn't fit to the BitBucket (`https://<host>/scm/~<user>/<repo>.git`) and DevAzure (`https://vinokurig@dev.azure.com/vinokurig/public/_git/public`) URL formats

ibuziuk: right, but I'm not sure if we should include all the patterns in the docs. That is why it is stated that `The format of the `git` command vary based on the Git Provider. As an example for GitHub ...`

vinokurig:
I could not find any info about git clone an URL with a token in the git documentation. Looks like this should be supported by a Git Provider itself. Are you sure that such clone works with all providers we support?

ibuziuk:
> Looks like this should be supported by a Git Provider itself

right, so the patterns for different provides are the following:

- GitHub `git clone https://<PAT>@github.com/username/repo.git`
- GitLab `git clone https://<PAT>@gitlab.com/username/repo.git`
- Azure `git clone https://<PAT>@dev.azure.com/organization/project/_git/repo`

We can include this information in the section as well, or just use GitHub as an example

modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
```diff
@@ -41,6 +41,22 @@ On OpenShift, you can use the `oc` command-line tool to log in to the cluster:
 
 . Generate your access token on your Git provider's website.
 
++
+[IMPORTANT]
+====
+Personal access tokens are sensitive information and should be kept confidential. Treat them like passwords. If you are having trouble with authentication, ensure you are using the correct token and have the appropriate permissions for cloning repositories:
+
+. Open a terminal locally on your computer
+. Use the `git` command to clone the repository using your personal access token. The format of the `git` command vary based on the Git Provider. As an example for GitHub, personal access token verification can be done using the following command:
```
deerskindoll:
```suggestion
. Use the `git` command to clone the repository using your personal access token. The format of the `git` command vary based on the Git Provider. As an example, GitHub personal access token verification can be done using the following command:
```

deerskindoll (approved):
one small nitpick, otherwise lgtm

l0rd (approved):

amisevsk (approved):
LGTM; I agree that we should be careful about documenting how to use a PAT to clone -- I seem to recall us having issues in dashboard with different URL schemes for enterprise versions of VCMs

modules/end-user-guide/pages/using-a-git-provider-access-token.adoc
```diff
@@ -41,6 +41,22 @@ On OpenShift, you can use the `oc` command-line tool to log in to the cluster:
...
+
+. Open a terminal locally on your computer
+. Use the `git` command to clone the repository using your personal access token. The format of the `git` command vary based on the Git Provider. As an example, GitHub personal access token verification can be done using the following command:
+
+----
+git clone https://<PAT>@github.com/username/repo.git
+----
+
+Replace `<PAT>` with your personal access token, and `username/repo` with the appropriate repository path.
+If the token is valid and has the necessary permissions, the cloning process should be successful. Otherwise, this is an indicator of incorrect personal access token, insufficient permissions, or other issues.
```
vinokurig, line 57: I am not sure that the token scopes to clone a repo are full enough to use with CHE. We require read user permissions for retrieving username and email, but it might be not necessary for the clone.

amisevsk: The project clone container only sees the token, and relies on username/email to be set via gitconfig, so while the full flow might not work the same without OAuth (prompting the user for name/email) it should be fine for cloning at least.

vinokurig: This should be handled in the notification then

ibuziuk: well, I think that user / email scope is separate is not relevant for this verification procedure. Based on the customer issue triage by far the most common issue was related to the project cloning. @vinokurig please, me know what should be updated in this PR (GH suggestions are most welcome)

vinokurig (approved):
LGTM as a project clone verification step

PR #2705, opened 2024-03-20
chore: add missing @antora/assembler dependency

themr0c:
Fixes build error: 

```
 [08:50:29.285] FATAL (antora): Cannot find module '@antora/assembler'
```


deerskindoll (approved):
lgtm

PR #2740, opened 2024-05-30
docs: added pvc resize info

deerskindoll:

added pvc resize info

https://issues.redhat.com/browse/CRW-6544

main

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -63,6 +63,5 @@ pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
 ====
 Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
 
-If you want to resize a PVC,
-modify the default workspace storage size in the `CheCluster` custom resource and restart the workspace.
+If you want to resize a PVC that is in use by a workspace, you must restart the workspace must before the PVC change can occur.
 ====
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -58,3 +58,13 @@ spec:
 <3> The claim size must be specified as a link:https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/[{kubernetes} resource quantity]. The available quantity units include: `Ei`, `Pi`, `Ti`, `Gi`,  `Mi` and `Ki`.
 
 pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
+
+[IMPORTANT]
+====
+Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may have unexpected consequences.
+
+If you want to resize a PVC, you have two options:
+
+* Modify the devfile's volume sizes accordingly and restart the workspace.
+* Modify the default workspace storage size in the `CheCluster` custom resource.
+====
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -65,6 +65,5 @@ Manually modifying a PVC on the cluster that was provisioned by {prod-short} is 
 
 If you want to resize a PVC, you have two options:
 
-* Modify the devfile's volume sizes accordingly and restart the workspace.
 * Modify the default workspace storage size in the `CheCluster` custom resource.
 ====
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -63,7 +63,5 @@ pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
 ====
 Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
 
-If you want to resize a PVC, you have two options:
-
-* Modify the default workspace storage size in the `CheCluster` custom resource.
+If you want to resize a PVC, modify the default workspace storage size in the `CheCluster` custom resource.
 ====
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -61,7 +61,7 @@ pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
 
 [IMPORTANT]
 ====
-Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may have unexpected consequences.
+Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
 
 If you want to resize a PVC, you have two options:
 
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -63,5 +63,5 @@ pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
 ====
 Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
 
-If you want to resize a PVC that is in use by a workspace, you must restart the workspace before the PVC change can occur.
+If you want to resize a PVC that is in use by a workspace, you must restart the workspace for the PVC change to occur.
 ====
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -63,5 +63,6 @@ pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
 ====
 Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
 
-If you want to resize a PVC, modify the default workspace storage size in the `CheCluster` custom resource.
+If you want to resize a PVC,
+modify the default workspace storage size in the `CheCluster` custom resource and restart the workspace.
 ====
--- a/modules/administration-guide/pages/configuring-storage-sizes.adoc
+++ b/modules/administration-guide/pages/configuring-storage-sizes.adoc
@@ -63,5 +63,5 @@ pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
 ====
 Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
 
-If you want to resize a PVC that is in use by a workspace, you must restart the workspace must before the PVC change can occur.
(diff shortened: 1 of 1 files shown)


deerskindoll:
@AObuchow is this wording ok?

modules/administration-guide/pages/configuring-storage-sizes.adoc
```diff
@@ -57,4 +57,14 @@ spec:
 <2> Specify a claim size on the next line or omit the next line to set the default claim size value. The specified claim size is only used when you select this storage strategy.
 <3> The claim size must be specified as a link:https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/[{kubernetes} resource quantity]. The available quantity units include: `Ei`, `Pi`, `Ti`, `Gi`,  `Mi` and `Ki`.
 
-pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
\ No newline at end of file
+pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
+
+[IMPORTANT]
+====
+Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may have unexpected consequences.
```
AObuchow: Personal preference: I'd reword end of the sentence as "and may result in unexpected consequences."

modules/administration-guide/pages/configuring-storage-sizes.adoc
```diff
@@ -57,4 +57,14 @@ spec:
...
+
+If you want to resize a PVC, you have two options:
+
+* Modify the devfile's volume sizes accordingly and restart the workspace.
```
AObuchow: I actually think the devworkspace might have to be deleted and recreated (not just restarted) :( I'll verify this in a bit and get back to you.

modules/administration-guide/pages/configuring-storage-sizes.adoc
```diff
@@ -57,4 +57,14 @@ spec:
...
+* Modify the default workspace storage size in the `CheCluster` custom resource.
```
AObuchow: It might be worth linking the documentation on configuring the storage size in Che/DevSpaces. I'll find the link.

AObuchow (commented):
Some small notes. Will get back with further clarification. Great work @deerskindoll :)

modules/administration-guide/pages/configuring-storage-sizes.adoc
AObuchow: Here's the link: https://github.com/eclipse-che/che-docs/blob/main/modules/administration-guide/pages/configuring-storage-sizes.adoc

modules/administration-guide/pages/configuring-storage-sizes.adoc
deerskindoll: you're the native speaker, I'll follow your guidance :D

modules/administration-guide/pages/configuring-storage-sizes.adoc
AObuchow: Oops.. I just realized the link I gave _is_ the page you're adding this note too :sweat_smile: Ignore this hahah.

modules/administration-guide/pages/configuring-storage-sizes.adoc
AObuchow: I just verified this: you have to modify the devfile's volume sizes accordingly (and backup these changes to git or wherever your devfile is stored) and **delete the workspace, and re-create the workspace** from your modified devfile. 

Restarting the workspace is not sufficient to change the PVC size. This is a feature I'd like to implement though, I hope it can become priortized :')

deerskindoll: since you have to delete the workspace and re-create it, should we even mention it as an option? creating a brand new workspace with a different PVC size seems a little bit faster.

AObuchow: Agreed -- it doesn't make sense to suggest deleting and re-creating the workspace. 

>  creating a brand new workspace with a different PVC size seems a little bit faster.

Agreed :+1:

modules/administration-guide/pages/configuring-storage-sizes.adoc
rafaeltuelho, line 1: The current note says two options exist, but the text only describes one. What is the second option? Also, it is worth mentioning that changing the volume size in the `devfile` and restarting the workspace will not expand the PV automatically.

deerskindoll: good catch! apologies, I'll fix the wording right away. Andrew and I decided to remove the second option from the note because it's more trouble than it's worth.

Check the updated wording and let me know if the message that modifying PVC is risky (and there's only one way to do it) is explicit enough. If yes, I think mentioning that changing the volume + restarting the workspace won't work is unnecessary. I can add it if you think it's important, though.

rafaeltuelho: Okay. Looks good. Concerning the change in the `devfile` I think it is worth adding a note, maybe not in the Admin guide, but in the User guide. Without a note a developer user may think it is just a matter of changing the volume size in its devfile and it will be automatically resized.

modules/administration-guide/pages/configuring-storage-sizes.adoc
```diff
@@ -57,4 +57,11 @@ spec:
...
 
-pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
\ No newline at end of file
+pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
+
+[IMPORTANT]
+====
+Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
+
+If you want to resize a PVC, modify the default workspace storage size in the `CheCluster` custom resource.
```
AObuchow: Sorry, I should have mentioned this earlier, but this is also unsupported :melting_face: Changing the default workspace storage size in the Che Cluster custom resource won't result in the existing PVCs being resized.

You'd have to delete and re-create your workspace (or delete the PVC and have it be automatically re-created, but this too is not _officially_ supported and only works for the per-workspace storage strategy as it causes a [bug](https://github.com/devfile/devworkspace-operator/issues/1258) with the per-user storage strategy).

So, it seems like the documentation we're adding here is either only a warning (that modifying the PVC on the cluster is not supported). We could also remark that you have to delete the workspace and re-create it, but that seems redundant, as we've discussed.

rafaeltuelho: I see it. So, yes, let's add it as a warning/important note. We definitely need to have this note added to the docs. I've seen users asking for this, and we need to state this even as a warning to make it clear the right/supported procedure for this.

rafaeltuelho: Whether in the `devfile` or `CheCluster` CR, users need to know the proper way to have a workspace PV resized.

deerskindoll: Since both options that we planned to feature in the note are actually not supported, is there a proper way to resize PVC outside of setting everything up from scratch?

rafaeltuelho: Well, in my experiments I noticed that changing the PVC manually and restarting the workspace works. But I'm not sure if the user itself would have permission to do that in its own namespace or if it would have to ask a cluster-admin to perform this change.

deerskindoll: @AObuchow @rafaeltuelho your call, guys. decide what should be included in the notification and I'll put it in the admin + end user guides.

AObuchow: After discussion, we've agreed upon informing the user that manually expanding the PVC is possible, but not officially supported. The user has to restart their workspace after requesting the expansion of their PVC (by modifiying the PVC on the cluster) in order for the PVC to actually get resized (as the workspace pod needs to detach from the PVC to allow expansion to occur).

deerskindoll: done! I hope I explained it correctly, please have a look.

modules/administration-guide/pages/configuring-storage-sizes.adoc
```diff
@@ -57,4 +57,12 @@ spec:
...
-pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
\ No newline at end of file
+pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
+
+[IMPORTANT]
+====
+Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
+
+If you want to resize a PVC,
+modify the default workspace storage size in the `CheCluster` custom resource and restart the workspace.
```
AObuchow: There's no actual need to modify the default workspace storage size in the Che-Cluster CR after modifying the PVC. 

I think this sentence could just be "If you choose to expand a PVC that is in use by a workspace, the workspace must be restarted before PVC expansion can occur"

AObuchow: I think after this change we should be good to merge. Sorry for being slow to follow-up on this @deerskindoll

AObuchow (changes requested):

modules/administration-guide/pages/configuring-storage-sizes.adoc
deerskindoll: no worries! I'll edit the wording.

deerskindoll: @AObuchow I reworded your suggestion a bit. please have a look

modules/administration-guide/pages/configuring-storage-sizes.adoc
```diff
@@ -57,4 +57,11 @@ spec:
...
 
-pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
\ No newline at end of file
+pass:[<!-- vale RedHat.CaseSensitiveTerms = YES -->]
+
+[IMPORTANT]
+====
+Manually modifying a PVC on the cluster that was provisioned by {prod-short} is not officially supported and may result in unexpected consequences.
+
+If you want to resize a PVC that is in use by a workspace, you must restart the workspace must before the PVC change can occur.
```
AObuchow: Small typo, must is repeated twice "you must restart the workspace must before"

deerskindoll: ooops, fixing immediately

AObuchow (approved):
LGTM :) 

@rafaeltuelho @dkwon17 @ibuziuk WDYT?

AObuchow (approved):

ibuziuk:
@deerskindoll @rafaeltuelho PR got stale for quite a while :/ can we merge now?

deerskindoll:
> @deerskindoll @rafaeltuelho PR got stale for quite a while :/ can we merge now?

if @rafaeltuelho is happy with the content, I'll be happy happy to publish it asap. there's also a code owner review needed, though.

rafaeltuelho:
It seems reasonable to me!

ibuziuk (approved):

PR #2745, opened 2024-06-11
chore: fix docs generating odd newline symbols

mkuznyetsov:

With release of Che 7.87.0, the content of [checlusters.yaml](https://github.com/eclipse-che/che-operator/blob/7.87.x/config/crd/bases/org.eclipse.che_checlusters.yaml) has changed due to upgrade of `controller-gen` software, that is used to create it.

Now, multi-line blocks for property descriptions are prefixed with `| -`, which results in creating extra `\n` sequences, when reading this block with yq. Then these sequences will trigger the HTML link check failure, which could be seen in the failure of [this workflow](https://github.com/eclipse-che/che-docs/actions/runs/9448342722/job/26032463596?pr=2743)

To fix this, in the generator script one extra instruction is added to replace any of the `\n` sequences with a space

This fix is needed to make the docs buildable (and pass the PR check for the release pull request - https://github.com/eclipse-che/che-docs/pull/2743 )

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/tools/checluster_docs_gen.sh
+++ b/tools/checluster_docs_gen.sh
@@ -159,9 +159,10 @@ parse_section() {
     else
       DESCR_BUFF=$(echo "$section" | yq -M '.properties.'"$PROP"'.description')
     fi
-    DESCR_BUFF="${DESCR_BUFF:1:-1}"       # Removes first and last quotes "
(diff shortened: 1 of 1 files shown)


deerskindoll (approved):
lgtm

PR #2983, opened 2025-10-24
chore: Add networkpolicies to security best practices

tolusha:
Add networkpolicies to security best practices

https://issues.redhat.com/browse/CRW-8328

main

## Pull Request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/partials/proc_creating-workspaces.adoc
+++ b/modules/end-user-guide/partials/proc_creating-workspaces.adoc
@@ -83,6 +83,9 @@ spec:
       - name: tooling-container
         container:
           image: quay.io/devfile/universal-developer-image:ubi9-latest
+          env:
+            - name: CHE_DASHBOARD_URL
+              value: pass:c,a,q[{prod-url}]/dashboard/ <7>
 ----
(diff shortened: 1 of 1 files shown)


modules/administration-guide/pages/security-best-practices.adoc
```diff
@@ -114,6 +114,90 @@ Every developer has a project in which the following objects are created and man
 Access to the resources in a namespace must be limited to the developer owning it. Granting read access to another developer is equivalent to sharing the developer credentials and should be avoided.
 ====
 
+.Network policies
+
+Network policies provide network-level isolation between {namespace} and pods in {kubernetes}.
```
ibuziuk: @tolusha I believe we should consider adding an `IMPORTANT:` block here + maybe cross-link it with https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/networking_overview/index#about-openshift-sdn docs to clarify that by default there is a "flat" pod network where every pod can communicate with every other pod and service on the cluster, and this default setup is not appropriate for the multitenant environment.

PR #3005, opened 2026-01-12
fix: change htmltest and antora version to solve build failures

mkuznyetsov:
This PR fixes a number of issues, that caused latest runs of CI builds to fail:
#### 1) build failed when htmltest found "broken links"
htmltest will now be NOT failing on errors, due to high amount of false positives. Any broken links for now will only be logged, but the job will be marked as success
#### 2) build failed when antora assembler produced warnings (as we configured to fail on warnings)
Warnings pointed to [recent changes to antora assembler](https://gitlab.com/antora/antora-assembler/-/issues/121) , where now a warning is produced if no site url is set (it marks those resources as unresolved). 

While the full fix will be handled in a separate PR, for now we will use the previous version of antora assembler

https://github.com/eclipse-che/che/issues/23695

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -16,6 +16,7 @@ DirectoryPath: build/site # Not build/site to avoid false positives on 404.html
 OutputDir: .cache/htmltest
 CacheExpires: "12h" # Default is 2 weeks.
 ExternalTimeout: 60 # (seconds) default is 15.
+IgnoreExternalBrokenLinks: true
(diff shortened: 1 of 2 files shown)


deerskindoll (approved):
lgtm

svor (approved):

PR #3012, opened 2026-01-16
procedures: Manage extension installation in Code - OSS with ConfigMap

RomanNikitenko:
Adds instructions how to manage extension installation in Code - OSS
<details>
<summary> Click here to see screenshots</summary>
image
image
image
image
image
image
image

#### Apply editor configurations changes:

image

</details>

 
- https://issues.redhat.com/browse/CRW-8866
- https://issues.redhat.com/browse/CRW-9626
- https://issues.redhat.com/browse/CRW-9627

Che-Code PRs:
- https://github.com/che-incubator/che-code/pull/605
- https://github.com/che-incubator/che-code/pull/617

## Pull Request checklist

- Any procedure:
  - [x] Successfully tested.
- Any page or link rename:

Assisted-by: Cursor AI

### DIFF
--- a/modules/administration-guide/pages/manage-extension-installation.adoc
+++ b/modules/administration-guide/pages/manage-extension-installation.adoc
@@ -1,0 +1,88 @@
+:_content-type: PROCEDURE
+:description: Manage extension installation with ConfigMap
+:keywords: extensions, workspace
+:navtitle: Manage extension installation with ConfigMap
+// :page-aliases:
+
+[id="visual-studio-code-manage-extensions-installation"]
+= Manage extension installation with ConfigMap
+
+This page describes how Code - OSS manages extension installation using a ConfigMap. 
+With these controls, you can enforce a fine-grained allow/deny list using the `AllowedExtensions` policy and block installs via the CLI, default extensions, and the `workbench.extensions.command.installFromVSIX` API command. 
+The sections below show how to enable and enforce these controls in Code - OSS.
+
+
+The following items are currently supported:
+
+* BlockCliExtensionsInstallation property - when enabled, blocks installation of extensions via CLI
+* BlockDefaultExtensionsInstallation property - when enabled, blocks installation of default extensions, see xref:default-extensions-for-microsoft-visual-studio-code.adoc[] 
+* BlockInstallFromVSIXCommandExtensionsInstallation property - when enabled, blocks installation of extensions via the workbench.extensions.command.installFromVSIX API command
+* AllowedExtensions section - provides fine-grained control over Code - OSS extension installation. For conceptual background, see link:https://code.visualstudio.com/docs/setup/enterprise#_configure-allowed-extensions/[Configure allowed extensions].
+
+.Procedure
+
+* Add a new ConfigMap to the {prod-namespace} namespace and specify the properties you want to add.
++
+====
+[source,yaml]
+----
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: vscode-editor-configurations
+  namespace: eclipse-che
+  
+  labels:
+    app.kubernetes.io/component: workspaces-config
+    app.kubernetes.io/part-of: che.eclipse.org
+  
+data:
+  policy.json: |
+    {
+      "BlockCliExtensionsInstallation": true,
+      "BlockDefaultExtensionsInstallation": true,
+      "BlockInstallFromVSIXCommandExtensionsInstallation": true,
+      "AllowedExtensions": { 
+          "*": true,
+          "dbaeumer.vscode-eslint": false,
+          "ms-python.python": false,
+          "redhat": false
+       }
+    }
+----
+====
+
+* Start or restart your workspace 
+
+[NOTE]
+====
+Make sure that the Configmap contains data in a valid JSON format.
+====
+
+.Verification
+. Verify that the `BlockCliExtensionsInstallation` property is applied:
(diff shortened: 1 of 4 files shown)


modules/administration-guide/pages/manage-extension-installation.adoc
```diff
@@ -0,0 +1,88 @@
...
+With these controls, you can enforce a fine-grained allow/deny list using the `AllowedExtensions` policy and block installs via the CLI, default extensions, and the `workbench.extensions.command.installFromVSIX` API command. 
+The sections below show how to enable and enforce these controls in Code - OSS.
+
+
+The following items are currently supported:
+
+* BlockCliExtensionsInstallation property - when enabled, blocks installation of extensions via CLI
+* BlockDefaultExtensionsInstallation property - when enabled, blocks installation of default extensions, see xref:default-extensions-for-microsoft-visual-studio-code.adoc[] 
+* BlockInstallFromVSIXCommandExtensionsInstallation property - when enabled, blocks installation of extensions via the workbench.extensions.command.installFromVSIX API command
+* AllowedExtensions section - provides fine-grained control over Code - OSS extension installation. For conceptual background, see link:https://code.visualstudio.com/docs/setup/enterprise#_configure-allowed-extensions/[Configure allowed extensions].
```
rgrunber: It might be useful to clarify that it seems this setting also serves as a restriction on extension activation. For example, in the case we discussed where someone could just insert an extension at `/checode/remote/extensions/` and update the `extensions.json` file there, the extension does get installed, but this property blocks the activation :

Screenshot From 2026-01-20 12-18-13

rgrunber (commented):
Overall, looks good. I would maybe consider giving an example that demonstrates how to completely disable extensions from being installed as this may be fairly common. Specifically, it should include :

```
"AllowedExtensions": {
  "*": false
}
```

If not, then it should be clear that the above block is just as important as the other 3 booleans in blocking extension installation.

One last thing to consider is it seems I needed to add some annotations to get things working if I didn't wish to apply the configmap to the `eclipse-che` namespace (ie. only apply it to my user's namespace). Might be good to document this.

ibuziuk (approved):

RomanNikitenko:
@rgrunber 
thank you very much for your review!

I tried to apply all your comments, please take a look at the last commit to check if everything looks good!
thanks!

---
update: I also updated screenshots in the PR description.

modules/administration-guide/pages/manage-extension-installation.adoc
```diff
@@ -0,0 +1,145 @@
...
+  
+  labels:
+    app.kubernetes.io/component: workspaces-config
+    app.kubernetes.io/part-of: che.eclipse.org
+  
+data:
+  policy.json: |
+    {
+      "AllowedExtensions": {
+        "*": false
```
rgrunber, line 89: This is very minor, but this only disables extension activation. They can still be installed, right ? Would you add the `Block*` booleans to also disable installation from all the relevant areas ? I would be fine with saying that prevents installation.

rgrunber (approved):
Overall, I think we are fine to merge. ~~Just one minor point in the comments.~~

modules/administration-guide/pages/manage-extension-installation.adoc
rgrunber: Ok, looks like `AllowedExtensions` is enough to block the install. Sounds fine.

modules/administration-guide/pages/manage-extension-installation.adoc
```diff
@@ -0,0 +1,153 @@
...
+  namespace: eclipse-che
+  
+  labels:
+    app.kubernetes.io/component: workspaces-config
+    app.kubernetes.io/part-of: che.eclipse.org
+  annotations:
+    controller.devfile.io/mount-as: file
+    controller.devfile.io/mount-path: /checode-config
+    controller.devfile.io/read-only: 'true'
+
```
RomanNikitenko, line 42: I've added these annotations to get `policy.json` section as a file on file system
@tolusha is it OK from your point of view?

tbh, I thought these annotations are added automatically to the corresponding ConfigMap in user namespace
but it is not true, many thanks to @rgrunber for thorough testing!

modules/administration-guide/pages/manage-extension-installation.adoc
```diff
@@ -0,0 +1,153 @@
...
+* Add a new ConfigMap to the {prod-namespace} namespace and specify the properties you want to add.
++
+====
+[source,yaml]
+----
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: vscode-editor-configurations
+  namespace: eclipse-che
```
RomanNikitenko, line 33: it's not possible to use `{prod-namespace}` here instead of `eclipse-che`
is there any other way to do it?
otherwise we need to replace the value manually for the downstream

modules/administration-guide/pages/manage-extension-installation.adoc
tolusha: Hello.

Only the following labels added automatically, but not annotations:
* controller.devfile.io/mount-to-devworkspace
* controller.devfile.io/watch-configmap

It is better to use `controller.devfile.io/mount-as: subpath` otherwise you will override the `/checode-config` directory completely.

RomanNikitenko: thank you!
updated https://github.com/eclipse-che/che-docs/pull/3012/commits/5940b0d01b862acc3b5b4ea3067f51dcb1a349a5

PR #3059, opened 2026-03-31
docs: mention DWO_NAMESPACE env var in backup docs

dkwon17:
Adds a notice about the DWO_NAMESPACE env var which is needed for the dashboard to detect backup status. This is needed only when the DWO install namespace is not openshift-operators.

DS 3.27.

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/administration-guide/pages/devworkspace-backup-regular-oci-registry.adoc
+++ b/modules/administration-guide/pages/devworkspace-backup-regular-oci-registry.adoc
@@ -33,6 +33,8 @@ config:
 ----
 <1> For Red Hat OpenShift, the default installation namespace for the {devworkspace} operator is `openshift-operators`. See the xref:devworkspace-operator.adoc[{devworkspace} operator overview].
 
+include::partial$snip_defining-dwo-namespace-for-backups.adoc[]
+
 The `authSecret` must be named `devworkspace-backup-registry-auth`. It must reference a {kubernetes} Secret of type `kubernetes.io/dockerconfigjson` that contains credentials to access the registry.
 The secret should be created in the installation {namespace} for the {devworkspace} operator.
 
--- a/modules/administration-guide/pages/devworkspace-backup-regular-oci-registry.adoc
+++ b/modules/administration-guide/pages/devworkspace-backup-regular-oci-registry.adoc
@@ -33,8 +33,6 @@ config:
 ----
 <1> For Red Hat OpenShift, the default installation namespace for the {devworkspace} operator is `openshift-operators`. See the xref:devworkspace-operator.adoc[{devworkspace} operator overview].
 
-include::partial$snip_defining-dwo-namespace-for-backups.adoc[]
-
 The `authSecret` must be named `devworkspace-backup-registry-auth`. It must reference a {kubernetes} Secret of type `kubernetes.io/dockerconfigjson` that contains credentials to access the registry.
 The secret should be created in the installation {namespace} for the {devworkspace} operator.
 
@@ -56,3 +54,5 @@ kubectl label secret devworkspace-backup-registry-auth controller.devfile.io/wat
 ====
 The {devworkspace} Operator copies the `devworkspace-backup-registry-auth` secret to each {devworkspace} {namespace} so that backups from user workspaces can be pushed to the registry. If you do not want that secret copied automatically, create a `devworkspace-backup-registry-auth` secret with user-specific credentials in each {devworkspace} {namespace} instead.
 ====
+
+include::partial$snip_defining-dwo-namespace-for-backups.adoc[]
--- a/modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
+++ b/modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
@@ -1,6 +1,8 @@
 :_content-type: SNIPPET
 
-**Note:** If the installation {namespace} for the {devworkspace} operator is not `openshift-operators`, you must define it as an environment variable for the {prod} dashboard in the `CheCluster` Custom Resource.
+[NOTE]
+====
+If the installation {namespace} for the {devworkspace} operator is not `openshift-operators`, you must define it as an environment variable for the {prod} dashboard in the `CheCluster` Custom Resource.
 
 [source,yaml,subs="+attributes,+quotes"]
 ----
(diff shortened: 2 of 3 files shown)


modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
```diff
@@ -0,0 +1,18 @@
...
+kind: CheCluster
+spec:
+  components:
+    dashboard:
+      deployment:
+        containers:
+          - env:
+              - name: DWO_NAMESPACE
+                value: $OPERATOR_INSTALL_NAMESPACE
+            name: che-dashboard
```
dkwon17, line 19: This value is the same for both upstream and downstream.

rohanKanojia (approved):

modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
```diff
@@ -0,0 +1,18 @@
+:_content-type: SNIPPET
+
+**Note:** If the installation {namespace} for the {devworkspace} operator is not `openshift-operators`, you must define it as an environment variable for the {prod} dashboard in the `CheCluster` Custom Resource.
```
gtrivedi88: Replace:
```asciidoc
**Note:** If the installation {namespace}...
```

With:
```asciidoc
[NOTE]
====
If the installation {namespace} for the {devworkspace} operator is not `openshift-operators`, you must define it as an environment variable for the {prod} dashboard in the `CheCluster` Custom Resource.

[source,yaml,subs="+attributes,+quotes"]
----
...
----
====
```

This wraps both the explanatory text and the YAML example inside a single admonition block, which is the correct AsciiDoc pattern for a note with a code example.

modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: org.eclipse.che/v2
+kind: CheCluster
+spec:
+  components:
+    dashboard:
+      deployment:
+        containers:
+          - env:
+              - name: DWO_NAMESPACE
+                value: $OPERATOR_INSTALL_NAMESPACE
```
gtrivedi88: Two options:

**Option A** — Use the formatted replaceable-text convention and add a brief explanation:
```asciidoc
                value: __<operator_install_namespace>__
```
Then after the source block, add:
```
where `__<operator_install_namespace>__` is the namespace where the {devworkspace} operator is installed.
```

**Option B** — Add a callout marker `<1>` (matching the pattern already used in the parent files) with an explanation after the source block.

Note: the parent procedure files already use `$OPERATOR_INSTALL_NAMESPACE <1>` with a callout explaining it's `openshift-operators` by default. For consistency, this snippet should provide the same guidance since it's a standalone reusable fragment that may be read without the surrounding callout context.

modules/administration-guide/pages/devworkspace-backup-regular-oci-registry.adoc
```diff
@@ -33,6 +33,8 @@ config:
 ----
 <1> For Red Hat OpenShift, the default installation namespace for the {devworkspace} operator is `openshift-operators`. See the xref:devworkspace-operator.adoc[{devworkspace} operator overview].
 
+include::partial$snip_defining-dwo-namespace-for-backups.adoc[]
```
gtrivedi88: The snippet include is inserted between the callout explanation for `<1>` (line 34) and the `authSecret` explanation paragraph (line 38). This breaks the reading flow — the reader goes from a callout about `$OPERATOR_INSTALL_NAMESPACE` to a note about `DWO_NAMESPACE`, then back to `authSecret` details.

Consider moving the `include::` to the end of the file (after the WARNING block about secret copying), matching the placement pattern used in `devworkspace-backup-integrated-openshift-registry.adoc` where it's appended at the very end. This keeps the registry credential setup instructions contiguous.

modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
```diff
@@ -0,0 +1,18 @@
...
+            name: che-dashboard
+----
```
gtrivedi88, line 20: Minor: The file should end with a final newline after `----` to comply with `.editorconfig` (`insert_final_newline = true`) and prevent AsciiDoc rendering issues when the snippet is included mid-file (as in the regular OCI registry page).

gtrivedi88 (changes requested):
Thanks for adding the `DWO_NAMESPACE` documentation — this is useful context for non-default operator installations.

A few CQA/style issues to address:

1. **Bold pseudo-heading** — `**Note:**` must be a proper `[NOTE]` admonition block (DITA compatibility, Vale `BlockTitle` rule)
2. **Bare placeholder** — `$OPERATOR_INSTALL_NAMESPACE` in the snippet has no explanation; the reader needs guidance on what value to substitute
3. **Snippet placement** — In the regular OCI registry file, the include breaks the flow between the callout and `authSecret` explanation; consider moving it to the end
4. **Missing final newline** — The snippet file should end with a newline

modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
dkwon17: Sounds good. I will also update the `**Note:**` in here as well: https://github.com/eclipse-che/che-docs/blob/85db9d503432ae2d6424698023234b1d3865641a/modules/administration-guide/pages/devworkspace-backup-integrated-openshift-registry.adoc?plain=1#L37

modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
dkwon17: Thank you, my latest commit follows Option B.

modules/administration-guide/partials/snip_defining-dwo-namespace-for-backups.adoc
dkwon17: fixed

modules/administration-guide/pages/devworkspace-backup-regular-oci-registry.adoc
dkwon17: Thank you, my latest commit moves it to the end of the file

gtrivedi88 (approved):

PR #3142, opened 2026-07-02
fix: flatten Upgrade nav to remove redundant nesting

gtrivedi88:
## Summary

Flatten the Upgrade navigation to remove redundant nesting.

**Before:** Upgrade → Upgrade Che → (procedures)
**After:** Upgrade → overview + procedures at the same level

The top-level nav already says "Upgrade", so nesting all procedures under another "Upgrade Che" page creates unnecessary depth.

## Test plan

- [ ] Nav renders flat: overview and procedures at the same level under Upgrade

Made with [Cursor](https://cursor.com)


gtrivedi88:
Superseded by PR #3148 which now includes the full JTBD audit (overview concept, persona cleanup, nav reorder, and more).

PR #3180, opened 2026-08-06
procedures: Add CLI Activity Tracker configuration documentation

tolusha:

Adds a new administration guide article documenting how to configure the CLI Activity Tracker at the cluster level via the CheCluster custom resource. The CLI Activity Tracker monitors running CLI processes in workspace containers and prevents workspace idling while those processes are active.

The new article covers:
- Enabling the CLI Activity Tracker via `spec.devEnvironments.cliActivityTracker`
- Configuring optional timing parameters (`secondsOfCheckPeriod`, `secondsOfActivityWindow`, `secondsOfGracePeriod`, `secondsOfMaxProcessAge`)
- Disabling the feature

Also updates the existing end-user guide article ("Prevent workspace idling for long-running commands") to cross-reference the new admin-level configuration.

Source PR: https://github.com/eclipse-che/che-operator/pull/2170
Related: https://github.com/eclipse-che/che-machine-exec/pull/376

- https://github.com/eclipse-che/che-operator/pull/2170
- https://issues.redhat.com/browse/CRW-11940

next

## Pull Request checklist

- Any procedure:
  - [ ] Successfully tested.
- Any page or link rename:

### DIFF
--- a/modules/end-user-guide/pages/proc_preventing-workspace-idling-for-long-running-commands.adoc
+++ b/modules/end-user-guide/pages/proc_preventing-workspace-idling-for-long-running-commands.adoc
@@ -10,7 +10,9 @@
 [role="_abstract"]
 Prevent a workspace from stopping while long-running CLI tools such as `helm`, `odo`, or `sleep` are active, so that unattended tasks can complete without interruption.
 
-The CLI Watcher periodically checks for running processes that match the commands you specify. When it detects an active watched command, it resets the workspace idle timer. This configuration works entirely in user space and does not require administrator privileges.
+The CLI Activity Tracker periodically checks for running processes that match the commands you specify. When it detects an active watched command, it resets the workspace idle timer. This configuration works entirely in user space and does not require administrator privileges.
+
+NOTE: An administrator can enable and configure the CLI Activity Tracker at the cluster level. See xref:administration-guide:configuring-cli-activity-tracker.adoc[].
 
 .Prerequisites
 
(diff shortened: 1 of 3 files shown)


modules/end-user-guide/pages/proc_preventing-workspace-idling-for-long-running-commands.adoc
```diff
@@ -10,7 +10,9 @@
 [role="_abstract"]
 Prevent a workspace from stopping while long-running CLI tools such as `helm`, `odo`, or `sleep` are active, so that unattended tasks can complete without interruption.
 
-The CLI Watcher periodically checks for running processes that match the commands you specify. When it detects an active watched command, it resets the workspace idle timer. This configuration works entirely in user space and does not require administrator privileges.
+The CLI Activity Tracker periodically checks for running processes that match the commands you specify. When it detects an active watched command, it resets the workspace idle timer. This configuration works entirely in user space and does not require administrator privileges.
+
+NOTE: An administrator can enable and configure the CLI Activity Tracker at the cluster level. See xref:administration-guide:configuring-cli-activity-tracker.adoc[].
```
vrubezhny, line 15: This change is not really enough. The following information with a `.noidle` config example, the description for the properties, etc. doesn't reflect the breaking change we've added: the use of `enabled: true` in `.noidle` is deprecated, so users won't be able to enable/disable the CLI Activity Tracker - the only admins can enable or disable it. 
Also the timing values that users can configure in `.noidle` config now are limited with the ones configured by the ones set by admins, f.i.  if admins sets `checkPeriodSeconds: 60` ([seconds] in spec or in custom user settings ConfigMap) and, at the same time,  user tries setting the same parameter to 120 seconds - the user's value won't be accepted and the admin's one will be effective. User can make these values to be more strict, but never relax them.

vrubezhny (changes requested):
User's `.noidle` configuration changes and a  breaking change on enablement are missing. See inline.

PS: All the said is true if `next` is for 3.31 (7.122/7.123) release.

